PR #15, opened 2021-06-16
chore: prepare get-safe-property-chain package for the mono-repo

pmareke:
Import the `get-safe-property-chain` package inside the new mono-repository


LuisMartinez15 (approved):

packages/get-safe-property-chain/package.json
```diff
@@ -0,0 +1,39 @@
+{
```
herrardo, line 1: They point number 4 in the bullet list is not done. 
image

davidmfempathy (approved):

PR #16, opened 2021-06-16
docs: Add LICENSE and update package.json files

pmareke:
Complete package information

### DIFF
--- a/packages/get-safe-property-chain/package.json
+++ b/packages/get-safe-property-chain/package.json
@@ -16,10 +16,12 @@
     "type": "git",
     "url": "git+https://cbn-javieri@bitbucket.org/colbenson/get-safe-property-chain.git"
   },
-  "keywords": [],
-  "author": "",
-  "license": "ISC",
-  "homepage": "https://bitbucket.org/colbenson/get-safe-property-chain#readme",
+  "keywords": [
+    "optional chaining"
+  ],
+  "author": "Empathy Systems Corporation S.L.",
+  "license": "Apache-2.0",
+  "homepage": "https://github.com/empathyco/x/tree/main/packages/x-get-safe-property-chain#readme",
   "publishConfig": {
     "access": "public"
   },
--- a/packages/get-safe-property-chain/package.json
+++ b/packages/get-safe-property-chain/package.json
@@ -14,14 +14,14 @@
   },
   "repository": {
     "type": "git",
-    "url": "git+https://cbn-javieri@bitbucket.org/colbenson/get-safe-property-chain.git"
+    "url": "git+https://github.com/empathyco/x.git"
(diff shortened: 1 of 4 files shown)


LuisMartinez15 (dismissed):

LuisMartinez15 (approved):

LuisMartinez15 (changes requested):
I totally forgot yesterday about this, but, I think we need to create a `CHANGELOG.md` file with the current `README.md` content. We have put something like this in the `README.MD` for now as a placeholder:

# x-get-safe-property-chain

A utility for safely retrieving nested values.

### Installation

You can install the package using `npm`

```
npm install @empathyco/x-get-safe-property-chain --save-dev
```

### Usage

```
import { getSafePropertyChain } from '@empathyco/x-get-safe-property-chain';

const obj = { nestedObject: { anotherNestedObject: { message: 'Hell yeah!' } } }; 
const message = getSafePropertyChain(obj, 'nestedObject.anotherNestedObject.message');
console.log(message); // 'Hell yeah!'
```

I think we should do this because now `lerna` is going to control the package versions and it will add the changes to the `CHANGELOG.md` (as we are doing in `x-components` with `standard-version`).

packages/get-safe-property-chain/package.json
```diff
@@ -16,10 +16,12 @@
     "type": "git",
     "url": "git+https://cbn-javieri@bitbucket.org/colbenson/get-safe-property-chain.git"
```
LuisMartinez15, line 17:
```suggestion
    "url": "git+https://github.com/empathyco/x.git"
```

LuisMartinez15 (changes requested):

LuisMartinez15 (approved):

packages/get-safe-property-chain/package.json
```diff
@@ -14,12 +14,14 @@
   },
   "repository": {
     "type": "git",
-    "url": "git+https://cbn-javieri@bitbucket.org/colbenson/get-safe-property-chain.git"
+    "url": "git+https://github.com/empathyco/x.git"
```
LuisMartinez15, line 17: This should be the url to the repo, without `git+ssh` or `git+https` as specify in the [documentation examples](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#repository).

LuisMartinez15 (changes requested):

packages/get-safe-property-chain/package.json
pmareke: I'll fix this `repository` property in the last PR.

herrardo (approved):

LuisMartinez15 (approved):

PR #24, opened 2021-06-28
chore: add `@empathyco/eslint-plugin-x` to all packages

LuisMartinez15:
EX-4273


javieri-empathy (changes requested):

pmareke:
As @javieri-empathy  mentioned we're running lint just in two packages

image

Please take a look.

pmareke:
And please fix the conflicts.

pmareke (commented):
Great job.

pmareke (changes requested):
Fix conflicts, please.

pmareke (approved):
LGTM

packages/get-safe-property-chain/.eslintignore
```diff
@@ -0,0 +1,2 @@
+**/*.js
+types
```
javieri-empathy, line 2: Why do you need to ignore the `types`? I tested removing this property for every package, run a fresh build, run the `lint` command, and everything is fine. As `type` files aren't included in the `tsconfig.eslint.json`, eslint shouldn't try to lint them

LuisMartinez15: Yep. However, if you open a file that is inside the `types` folder, you get the following error:

image

It is just to avoid it. If you think this is not necessary, I can remove `types` from the `.eslintignore` file.

javieri-empathy (approved):

PR #37, opened 2021-07-08
fix(components): make column picker & grid components listen to only …

javieri-empathy:
# Description

Fixes column picker & grid components to have an incorrect number of columns when re-rendered. This happens because the columns number state was handled using RxJS with 2 different `Subject`s. When a component re-rendered, it subscribed to those 2 events, and it never knew which one had the correct number of columns:

- `UserClickedColumnsPicker` had the last user selected columns
- `ColumnPickerSetColumnsNumber` had the last developer provided selected columns

Depending on the situation, you might want to listen either to the last number of selected columns by the final user:

- Search for something
- Select 2 columns grid
- Clear query
- Search for something new
- User expects to still have 2 columns grid

Or to the developer provided value

- Search for something
- Select 6 columns grid
- Resize until a breakpoint like mobile
- New columns number should be provided by the developer. If they changed the `:columns` list, or the `:value` property.

# Context

This was firstly noticed in the archetype.

# Test

You can check the last added test to every one of these components, and compare this behavior with the one deployed at https://x.test.empathy.co/


packages/x-components/package.json
```diff
@@ -22,7 +22,7 @@
     "test:unit-watch": "vue-cli-service test:unit --watch",
     "test:e2e": "vue-cli-service test:e2e",
     "test:e2e:ci": "vue-cli-service test:e2e --headless --browser chrome",
-    "test": "npm run lint && npm run test:unit",
+    "test": "npm run test:unit",
```
javieri-empathy, line 25: Linter is now run on a separate step

packages/x-components/src/components/column-picker/base-column-picker-list.vue
```diff
@@ -47,7 +56,10 @@
    * @public
    */
   @Component({
-    components: { BaseEventButton }
+    components: { BaseEventButton },
+    model: {
+      event: 'change'
```
javieri-empathy, line 61: This is to match the other similar components (like the sort selectors, or the dropdown)

packages/x-components/src/components/column-picker/base-column-picker-dropdown.vue
```diff
@@ -1,33 +1,31 @@
...
+    @change="emitEvent"
+    :value="selectedColumns"
+    :items="columns"
+    :animation="animation"
+  >
+    <template v-if="$scopedSlots.toggle" #toggle="{ item, isOpen }">
+      <!--
            @slot From `BaseDropdown` component: Used to render the contents of the dropdown toggle
            button. If not provided, it uses the `item` slot as fallback.
-               @binding {string|number|Identifiable} item - The item data to render.
```
tiborux, line 13: - [ ] Add again the spaces

packages/x-components/src/components/column-picker/base-column-picker-dropdown.vue
```diff
@@ -1,33 +1,31 @@
...
-      </template>
-      <template #item="{ item, isSelected, isHighlighted }">
-        <!--
+      <slot name="toggle" v-bind="{ item, isOpen }" />
+    </template>
+    <template #item="{ item, isSelected, isHighlighted }">
+      <!--
            @slot (required) From `BaseDropdown` component: Used to render each one of the items
            content, and as fallback for the toggle button content slot if it is not provided.
-               @binding {string|number|Identifiable} item - Item to render
```
tiborux, line 23: - [ ] Add again the spaces

tiborux (changes requested):

packages/x-components/src/components/__tests__/base-variable-column-grid.spec.ts
```diff
@@ -12,63 +12,64 @@ const itemsStub = [
   ...searchResponse.results
 ];
 
-function renderComponent({ items = itemsStub }: RenderOptions): RenderAPI {
+function renderComponent({ items = itemsStub }: RenderOptions = {}): RenderAPI {
   const [, localVue] = installNewXPlugin();
-  const wrapper = mount(BaseVariableColumnGrid, {
-    components: {
-      BaseVariableColumnGrid
```
davidmfempathy, line 19: Just for my information. This was wrong, wasn't it? I mean importing that component which is not used in the template.

javieri-empathy: Yep, the `components` option is used to locally register components, so then in the render function or in the template you can find them by their name. But this was directly mounting the `BaseVariableColumnGrid` component, there was no custom wrapper component like we do in other cases. So it is unnecessary.

packages/x-components/src/components/column-picker/base-column-picker-dropdown.vue
javieri-empathy: Let's ask in slack, because I remember someone telling me to remove those spaces in other PRs

tiborux (approved):

davidmfempathy (changes requested):

packages/x-components/src/components/base-variable-column-grid.vue
```diff
@@ -20,7 +20,7 @@
```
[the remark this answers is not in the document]
javieri-empathy: That's what it does, the events bus uses `ReplaySubject`, so even if you subscribe to an event and no events are  emitted, you will still receive the last emitted value. 
The column pickers already emit an event with this value before mounting them.

packages/x-components/src/components/column-picker/__tests__/base-column-picker-list.spec.ts
```diff
@@ -79,14 +93,14 @@ describe('testing Base Column Picker List', () => {
...
     const index = 1;
-    const { wrapper, clickEventButton } = renderBaseColumnPickerListComponent({ columns });
+    const { wrapper, clickNthItem } = renderBaseColumnPickerListComponent({ columns });
     const listenerColumnPicker = jest.fn();
     wrapper.vm.$x.on('UserClickedColumnPicker', true).subscribe(listenerColumnPicker);
-    await clickEventButton(index);
-    expect(wrapper.emitted('input')).toEqual([[columns[index]]]);
+    await clickNthItem(index);
+    expect(wrapper.emitted('change')).toEqual([[columns[index]]]);
     expect(listenerColumnPicker).toHaveBeenCalledTimes(1);
```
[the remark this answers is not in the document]
javieri-empathy: I moved the assertions to the above test, and added the missing `ColumnsNumberProvided` event

packages/x-components/src/components/column-picker/base-column-picker-list.vue
```diff
@@ -105,17 +147,31 @@ updated if it changed the value or if the parent changes it.
...
+    data() {
+      return { columns: [2, 4, 6] };
+    }
+  };
+</script>
 ```
 
 ## Events
 
 A list of events that the component will emit:
```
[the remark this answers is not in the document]
javieri-empathy: Yep, probably can be used for tracking, but nothing else so far

davidmfempathy (approved):

tiborux (approved):

PR #40, opened 2021-07-12
Feature/ex 4323 update release config

javieri-empathy:
# Description

Configures lerna to generate change-logs with the same format. Now it uses `conventionalcommits` preset instead of the default `angular` one, because  it does not support configurations.

# Context

# Test

You can try to do a LOCAL release. Please be careful and do not push these commits to the repo:

```
npm ci
npm run build
git fetch --tags
lerna version --conventional-commits --conventional-prerelease --no-push
```

There you can check the changes made to the packages. To restore the previous state of the branch run:

```
git tag | xargs git tag -d
git reset HEAD^1 --hard
git fetch --tags
```


packages/jest-utils/package.json
```diff
@@ -29,7 +29,7 @@
     "@types/jest": "^24.0.12 <= 24.0.13",
     "jest": "^24.7.1 <= 24.8.0",
     "ts-jest": "^24.0.2",
-    "typescript": "^3.4.5"
+    "typescript": "^3.8.3"
```
javieri-empathy, line 32: The build was failing because some dependency of jest had declaration files incompatible with the previous 3.4.5 TS version. This should be a harmless change, but you are free to check it more in depth if you want

davidmfempathy (approved):

tiborux (approved):

PR #41, opened 2021-07-12
feat(search): add `Promoted` component

javieri-empathy:
# Description

Adds the `Promoted` component. Just a link, an image, and a title.

# Test

There isn't much to test in this component. Just check that everything is working fine, and the component renders as it should.


packages/x-components/src/x-modules/search/components/promoted.vue
```diff
@@ -0,0 +1,74 @@
+<template>
+  <a :href="item.url" class="x-promoted" data-test="promoted">
+    <img :src="item.image" class="x-promoted__image" alt="" />
+    <h1 class="x-promoted__title">{{ item.title }}</h1>
```
javieri-empathy, line 4: This one shouldn't probably be a `h2`, but I added it to match the results one

tiborux (changes requested):
Add to the barrel the Promoted export

javieri-empathy:
> Add to the barrel the Promoted export

Good catch

packages/x-components/src/x-modules/search/components/promoted.vue
```diff
@@ -0,0 +1,74 @@
+<template>
+  <a :href="item.url" class="x-promoted" data-test="promoted">
+    <img :src="item.image" class="x-promoted__image" alt="" />
```
javieri-empathy, line 3: The empty alt is to skip the accesibility for this image. We don't have a field to add a descriptive text to the image, and we are already showing the title, which is the closer we can get to that descriptive text.

packages/x-components/src/x-modules/search/components/promoted.vue
```diff
@@ -0,0 +1,74 @@
+<template>
+  <a :href="item.url" class="x-promoted" data-test="promoted">
```
davidmfempathy, line 2: Should we emit an event `UserClickedAPromoted` as we do with the results? Then the developer user will be able to use it in a wire (I can't find an specific example to do it know).

packages/x-components/src/x-modules/search/components/promoted.vue
davidmfempathy, line 3: For the image, in the result we have a more complex component which manages also fallback images, placeholder images, etc.

Don't we need that here? I can accept as answer that Promoted and Banners are less relevant (well, not less relevant, but they are less frequent.

packages/x-components/src/x-modules/search/components/promoted.vue
```diff
@@ -0,0 +1,74 @@
...
+
+_Here you can see how the `Promoted` component is rendered._
+
+```vue
+<template>
+  <Promoted :item="promoted" />
+</template>
+
+<script>
+  import { Promoted } from '@empathyco/x-components/search';
```
davidmfempathy, line 53: Sometimes I see that we import it from x-components and sometimes from a module of x-components.
What should we do?

packages/x-components/src/x-modules/search/components/promoted.vue
```diff
@@ -0,0 +1,74 @@
...
+    name: 'PromotedDemo',
+    components: {
+      Promoted
+    },
+    data() {
+      return {
+        promoted: {
+          modelName: 'Promoted',
+          id: 'promoted-example',
+          link: 'https://my-website.com/summer-shirts',
```
davidmfempathy, line 65: This property is `url` instead of `link`.

packages/x-components/src/x-modules/search/components/__tests__/promoted.spec.ts
```diff
@@ -0,0 +1,75 @@
+import { Promoted as PromotedModel } from '@empathyco/x-types';
+import { mount, Wrapper } from '@vue/test-utils';
+import Vue from 'vue';
+import { createPromotedStub } from '../../../../__stubs__/promoted-stubs.factory';
+import { installNewXPlugin } from '../../../../__tests__/utils';
+import { getXComponentXModuleName, isXComponent } from '../../../../components/x-component.utils';
+import Promoted from '../promoted.vue';
+
+function renderPromoted({
```
davidmfempathy, line 9: We don't need to create a wrapper component to test it as it's a simple component without slots. But it's not mandatory to change it. 

The criteria or the recommendation to use it or not is in the glossary:

- Sometimes it's easier to create a wrapper component in which we can import and render our
  component or components, especially when we use slots, defining the template. It's preferable to
  test the component isolated without this wrapper. Notice that you can also include this within a
  factory function in order to override the template as another property.

packages/x-components/src/x-modules/search/components/__tests__/promoted.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+        title: 'Search UIs',
+        image: 'https://empathy.co/x-components.jpg',
+        tagging: {
+          click: { url: 'https://track-things.com', params: {} }
+        }
+      }
+    });
+
+    expect(wrapper.text()).toEqual('Search UIs');
+  });
```
davidmfempathy, line 62: We test if it renders the component in the `wrapper.findComponent(Promoted)` so we can change the it description to `renders the promoted title`; or change the test. We can test if it's rendered looking for it with the data-test (which is set in the component and it's not used anywhere.

davidmfempathy (changes requested):

packages/x-components/src/__stubs__/promoted-stubs.factory.ts
```diff
@@ -8,7 +8,7 @@ import { Promoted } from '@empathyco/x-types';
  * @internal
  */
 export function getPromotedStub(): Promoted[] {
```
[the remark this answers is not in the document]
tiborux: What i can see here is that we have two different stubs: 
image
One is to create promoted dynamically and the other one has a defined values, should we remove one or mix them?

packages/x-components/src/x-modules/search/components/promoted.vue
tiborux: I'm not sure but this will be useful for the tagging, i dont know how we are going to manage this behaviour.

packages/x-components/src/x-modules/search/components/promoted.vue
tiborux: Promoted and banners must have an image:
image

packages/x-components/src/x-modules/search/components/__tests__/promoted.spec.ts
davidmfempathy: won't do

packages/x-components/src/x-modules/search/components/promoted.vue
davidmfempathy: Wont't do

packages/x-components/src/x-modules/search/components/promoted.vue
davidmfempathy: Won't do

davidmfempathy (approved):

tiborux (approved):

PR #44, opened 2021-07-13
test(component): configure Cypress component testing

javieri-empathy:
# Description

Configures [Cypress component testing](https://docs.cypress.io/guides/component-testing/introduction#What-is-Component-Testing) for the x-components project. To do so, the following changes have been made:

- Separate Cypress e2e & unit tests in different folders
- Remove unnecessary `webpack.config.js`. Cypress has now support by default for TypeScript, and cucumber can be configured easily in other way.
- Adjust paths

# Context

Some things can't be tested with jest and JSDOM, we need to have a real browser environment to properly test them. One of this examples is the infinite scroll.

# Test

Nothing has broken, and you can run the new cypress component tests properly


packages/x-components/package.json
```diff
@@ -45,6 +45,8 @@
     "gen:typescript-docs": "api-documenter markdown --input-folder dist/report --output-folder docs/API-reference/api",
     "gen:component-docs": "vue-docgen",
     "test:unit": "vue-cli-service test:unit",
```
tiborux, line 47: We should remove the old one?

packages/x-components/tests/unit/test.spec.ts
```diff
@@ -0,0 +1,8 @@
+import { mount } from '@cypress/vue';
+
+describe('Sample test file', () => {
```
tiborux, line 3: Is this necessary? I mean its just an example right?

tiborux (changes requested):

davidmfempathy (approved):

tiborux (approved):

PR #46, opened 2021-07-15
ci: Fix publishing in x-components. Normalize build scripts.

javieri-empathy:
# Description

Tries to fix the publishing x components issue, and normalises the package json scripts. 

- Now installing the projects will call the `build` script in each one of the projects
- After building, every project will invoke the `npm pack` command
- Before publishing, a new fresh build should be triggered to update the `dist` folder in projects like x-components, which only publish the `dist` directory, not the project folder.
- Pushing will only be triggered if the publish command success. Previously it was triggered even if it failed.

# Context

This is needed to improve the CI flow

# Test

Make a fresh install, check that the projects are properly build, and understand the made changes.


packages/get-safe-property-chain/package.json
```diff
@@ -24,6 +24,9 @@
     "build": "tsc && tsc --project tsconfig.esm.json",
     "build:watch": "tsc --watch",
     "lint": "eslint src",
+    "postbuild": "npm pack",
+    "postinstall": "npm run build",
```
javieri-empathy, line 28: Lerna says that it invokes, `prepublish`, and `prepare` commands after a CI, but the problem with those commands is that `prepublish` is deprecated, and both of them are also invoked by npm before the `npm pack` command. So i chose to use `postinstall` and `prepublishOnly`.

LuisMartinez15 (approved):
The flow seems to be working as expected.

packages/deep-merge/package.json
```diff
@@ -29,12 +29,13 @@
...
   "scripts": {
     "build": "tsc && tsc --project tsconfig.esm.json",
-    "build:local": "webpack --mode production && npm pack",
     "clean": "rimraf dist && rimraf coverage",
     "format": "prettier --write \"src/**/*.ts\"",
     "lint": "eslint src",
+    "postbuild": "npm pack",
+    "postinstall": "npm run build",
     "prepublishOnly": "npm run build",
-    "prebuild": "npm run clean && npm run lint && echo Using TypeScript && tsc --version",
```
pmareke, line 37: This `npm run lint` that you remove when it's invoked now?

pmareke (changes requested):
I can't see the `.tgz` file for the eslint plugin:

image

pmareke (approved):
LGTM

PR #48, opened 2021-07-19
Feature/ex 4322 migrate search types develop bra 

tiborux:
Migrate develop branch to mono repo.

The problem that we had here, was that the package was deleted when you run npm install, to avoid that, we ran some commands:

`lerna clean`

`npm install`

I think that we can avoid to use the second one, `lerna bootstrap --hoist` , because npm install will do that command.


package.json
```diff
@@ -5,7 +5,7 @@
     "postinstall": "lerna bootstrap --hoist",
     "build": "lerna run build",
     "lint": "lerna run lint",
```
LuisMartinez15, line 7: As we discussed yesterday, we could add `--parallel` here too.

```suggestion
    "lint": "lerna run lint --parallel",
```

LuisMartinez15 (approved):

LuisMartinez15 (approved):

PR #50, opened 2021-07-21
Feature/ex 3433 mock the adapter for recommendations e2e tests

CachedaCodes:
## Motivation and Context ##

We needed to have a reliable method of doing e2e tests without relying on APIs outside of our control. A possible solution was suggested in [EX-3432](https://searchbroker.atlassian.net/browse/EX-3432) comments. This PR implements a version of that solution for the Recommendations (TopClicked).

* `mocked-adapter` changed so the `getTopRecommendations` can be intercepted by Cypress
* e2e test for the Recommendations changed to make use of Cypress' intercept
* Results stub changed to be more generic
* Tests that used results stub changed to make use of the more generic approach

In `cypress.json` you can change the `testFiles` property's first element to `"cucumber/**/*.features"` to be able to select the test to run.

## Task ##

[EX-3433](https://searchbroker.atlassian.net/browse/EX-3433)


packages/x-components/tests/e2e/cucumber/recommendations/recommendations.spec.ts
```diff
@@ -1,22 +1,44 @@
...
     recommendations: {
       config: {
         maxItemsToRequest
       }
     }
   };
-  cy.visit('/test/recommendations', {
+});
+
+And('suggestions response being mock {string}', (mockName: string) => {
```
CachedaCodes, line 16: Main changes from the proposed solution in the task's comment are here:

* The `intercept` has to be done before the request is done so it has to be set up before the `visit call`.
* Dynamic import is not that dynamic after all and you can't use a variable string as the first argument. My proposed solution is to have just one file for the stubs (instead of one per stub as suggested) and dynamically select the mock to use from the ones being exported in the stub file.

packages/x-components/tests/e2e/cucumber/recommendations/recommendations.spec.ts
```diff
@@ -1,22 +1,44 @@
...
+  cy.intercept('https://api.empathy.co/topRecommendations', req => {
+    import(`./stubs/recommendations.stub`)
+      .then(mock => {
+        return mock[mockName as keyof typeof mock];
+      })
+      .then(mockData => {
+        req.reply(mockData);
+      });
+  }).as('recommendationsApi');
```
LuisMartinez15, line 25: I think I was able to solve the issue you are describing above:

```suggestion
  cy.intercept('https://api.empathy.co/topRecommendations', async req => {
    const module = await import(`./stubs/recommendations.stub`);
    req.reply((module as any)[mockName]);
  }).as('recommendationsApi');
```

This is working fine for me.

This change implies changing  the `recommendations.feature` `mockName` values to the new one.

packages/x-components/tests/e2e/cucumber/recommendations/recommendations.spec.ts
```diff
@@ -1,22 +1,44 @@
...
     }
   });
 });
 
 // Scenario 1
+Then('number of rows requested is {int}', (maxItemsToRequest: number) => {
+  cy.wait('@recommendationsApi').then(interception => {
+    const interceptedRequestBody = JSON.parse(interception.request.body);
+    expect(interceptedRequestBody.rows).to.equal(maxItemsToRequest);
+  });
```
LuisMartinez15, line 39: You could refactor this to:

```suggestion
  cy.wait('@interceptedRecommendations').then(({ request }) => {
    const { rows } = JSON.parse(request.body) as TopRecommendationsRequest;
    expect(rows).to.equal(maxItemsToRequest);
  });
```

I think it is more readable using the real request type.

packages/x-components/tests/e2e/cucumber/recommendations/recommendations.feature
```diff
@@ -2,10 +2,12 @@ Feature: Recommendations component
...
+    Then  number of rows requested is <maxItemsToRequest>
     Then  number of displayed recommendations are equal or less than <maxItemsToRequest>
     And   each recommendation has an associated hyperlink containing image and text
 
     Examples:
-      | maxItemsToRequest |
-      | 7                 |
-      | 160               |
+      | maxItemsToRequest | mock    |
+      | 3                 | default |
```
LuisMartinez15, line 12: If you end up refactoring the tests using a `named export` instead of the `default` one, remember to update these values here.

packages/x-components/tests/e2e/cucumber/recommendations/stubs/recommendations.stub.ts
```diff
@@ -0,0 +1,9 @@
+import { createResultStub } from '../../../../../src/__stubs__/results-stubs.factory';
+
+export default {
+  results: [
+    createResultStub('Piscina 3 Anillos'),
+    createResultStub('Among Us Figura Acción'),
+    createResultStub('Barbie Sirenas Dreamtopia')
+  ]
+};
```
LuisMartinez15, line 9: If you end up refactoring the tests using a `named export` instead of the `default` one, this should be refactored to something like:

```suggestion
export const recommendationsStub = {
  results: [
    createResultStub('Piscina 3 Anillos'),
    createResultStub('Among Us Figura Acción'),
    createResultStub('Barbie Sirenas Dreamtopia')
  ]
};
```

LuisMartinez15 (changes requested):

packages/x-components/src/__stubs__/results-stubs.factory.ts
```diff
@@ -11,81 +10,46 @@ import { DeepPartial } from '../utils/types';
...
-      price: {
-        hasDiscount: true,
-        value: 30.99,
-        originalValue: 75.95
-      },
-      ...getResultCommonValues()
-    }
-  ] as Result[];
+    tagging: getResultTagging(kebabCaseName),
+    url: `/products/${name}`,
```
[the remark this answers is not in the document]
davidmfempathy: Agree

packages/x-components/tests/e2e/cucumber/recommendations/recommendations.spec.ts
davidmfempathy: The issue is that you cannot pass an argument to the import function and make it work as Guille said. What you've suggested do the same but await.

packages/x-components/tests/e2e/cucumber/recommendations/recommendations.spec.ts
```diff
@@ -1,22 +1,44 @@
...
         maxItemsToRequest
       }
     }
   };
-  cy.visit('/test/recommendations', {
+});
+
+And('suggestions response being mock {string}', (mockName: string) => {
+  cy.intercept('https://api.empathy.co/topRecommendations', req => {
+    import(`./stubs/recommendations.stub`)
```
davidmfempathy, line 18: Use single quotes

packages/x-components/tests/e2e/cucumber/recommendations/recommendations.spec.ts
davidmfempathy: I noticed that the dynamic import with dynamic argument doesn't work because it's not resolved because the spec are not pre processed correctly or in the correct order. https://docs.cypress.io/guides/tooling/plugins-guide.html#Preprocessors. We can leave it as it's but just a comment just in case we need it in the future.

packages/x-components/tests/e2e/cucumber/recommendations/stubs/recommendations.stub.ts
davidmfempathy: The whole file is intended to be used for the recommendations stubs. Each export should be that we expected to have in the filename.

davidmfempathy (changes requested):
1. The build is failing.
2. Is the unique stub going to be used as dummy too ?

packages/x-components/tests/e2e/cucumber/recommendations/recommendations.spec.ts
CachedaCodes: Leaving it as-is for now at least as I also prefer one stub file if possible.

packages/x-components/tests/e2e/cucumber/recommendations/recommendations.spec.ts
CachedaCodes: It actually fixes an error when running `test:e2e:ci` so I'm going with this approach.

packages/x-components/src/adapter/mocked-adapter.ts
```diff
@@ -111,3 +111,23 @@ function rejectIn(ms: number, rejectionValue: unknown): Promise<any> {
     }, ms);
   });
 }
+
+/**
+ * Creates a `fetch` call to a non existent API endpoint with the provided request body.
+ *
+ * @param request - The request's body to use in the `fetch`.
```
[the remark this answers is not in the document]
CachedaCodes: The type of `request` is `TopRecommendationsRequest` and the other `AdapterFeatures` also have `request` in the type name and the parameter is called `request`. I would maintain the name for consistency.

````
// search-adapter.types.ts
getNextQueries(request: NextQueriesRequest [...]
 getTopRecommendations(request: TopRecommendationsRequest [...]
 getSectionRecommendations(request: SectionRecommendationsRequest [...]
````

davidmfempathy (approved):
I approved it but the build still fails.

packages/x-components/tests/e2e/cucumber/recommendations/recommendations.spec.ts
LuisMartinez15: This issue is in reality caused by webpack, probably some config issues in vue-cli. I've to look deeper into it.

References:
https://github.com/webpack/webpack/issues/6680
https://webpack.js.org/api/module-methods/#dynamic-expressions-in-import

LuisMartinez15 (approved):

PR #52, opened 2021-07-22
ci(eslint): add `eslint-plugin-x`dependency

tiborux:
Add eslint-plugin-x


CachedaCodes (commented):
Should we add a `.eslintignore` with
````
**/*.js
types
````

And a `.eslintrc` with
````
{
  "extends": ["plugin:@empathyco/x/all"]
}
````

Like the other packages?

File: packages/jest-utils/.eslintrc [hunk 1]
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": ["plugin:@empathyco/x/all"],
+  "rules": {
+    "@typescript-eslint/restrict-template-expressions": ["off", { "allowAny": true }],
```
davidmfempathy, line 4: I've noticed that both problems come from: x/packages/jest-utils/src/jest-utils.ts.
Isn't it preferable to disable both rules at least just for that file ?

packages/jest-utils/tsconfig.eslint.json
```diff
@@ -0,0 +1,4 @@
+{
+  "extends": "./tsconfig.json",
```
davidmfempathy, line 2: Can we just have everything in the tsconfig.eslint.json instead of extending it from another file?

packages/jest-utils/src/jest-utils.types.ts
```diff
@@ -1,22 +1,23 @@
 export interface EmpathyExtendedExpect {
   arrayOf(classType: Newable): any;
-  arrayOfItemsMatching(schema: {}): any;
+  arrayOfItemsMatching(schema: Record<string, any>): any;
```
davidmfempathy, line 3: Just for my information, did the linter complain about this?

davidmfempathy (changes requested):

packages/jest-utils/src/jest-utils.types.ts
tiborux: the linter doesn't like to use an empty object, so i added a record

packages/jest-utils/tsconfig.eslint.json
tiborux: this is how its done in other packages

tiborux:
@CachedaCodes we dont have types and js files in this package

CachedaCodes (approved):

davidmfempathy (approved):

PR #54, opened 2021-07-22
Feature/ex 4327 rename design tokens

tajespasarela:
# Pull Request Template #

## Motivation and Context ##
The main goal of this PR is to adapt the tokens naming to work with Style Dictionary tool (https://amzn.github.io/style-dictionary).

The problems to solve before starting using this tool are:

1- The name of a token cant start with the name of other tokens:
          WRONG:  `--x-color-background-button`  `--x-color-background-button-primary`

This is because Style Dictionary merges all the JSON files in one and those tokes will have the same path inside the JSON, so the second one will override the first.

To solve this, what this PR does is to add the `-default` variant to the default tokens:
          RIGHT:  `--x-color-background-button-default`  `--x-color-background-button-primary`

2- The categories of the tokens must be these:
- `color`: For the color values (#FFAABB, lightblue, etc.).
- `size`: For any value with size units (px, em, %, etc).
- `number`: For any number value (font-weight, etc).
- `font`: For the font family.
- `time`: For animations times.
- `string`: For any other token value that doesn't fit in previous categories.

This is because the Style Dictonary uses that category to know how to transform/format the value of the token.

3- The tokens only must have one value. So the padding/margin/etc shorthands in the tokens are forbidden.
 WRONG:  `--x-size-margin-button`: 10px 10px 20px 30px;

  RIGHT:   `--x-size-margin-top-button`: 10px;
                 `--x-size-margin-right-button`: 10px;
                 `--x-size-margin-bottom-button`: 20px;
                 `--x-size-margin-left-button`: 30px;

This is because of the integration with Mobile (Android, iOS) where these shorthands don't work in the same way.

To make it easier to configure the tokens in some places we make the 4 tokens point to one single token. This way we can configure all the borders for example with only one token, but there will be 4 tokens generated to use in CSS and in Mobile.

## Type of change ##

- [ ] New feature (non-breaking change which adds functionality)
- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

All the Design System should be working as before, so check all the views that use the design system.


packages/x-components/src/design-system/components/sliding-panel/default.scss
```diff
@@ -3,8 +3,8 @@
   background-color: var(--x-color-background-sliding-panel);
```
tiborux, line 3: the z index is not right:

image

tiborux: Same for hover:

image

packages/x-components/src/design-system/components/list/default.scss
```diff
@@ -3,11 +3,11 @@
 .x-list {
   // layout
```
tiborux, line 4: Item sizing example is not working:

image

tiborux (changes requested):

packages/x-components/src/design-system/components/button/ghost.scss
```diff
@@ -1,19 +1,19 @@
...
+  --x-color-border-button-default: var(--x-color-border-button-ghost);
+  --x-color-text-button-default: var(--x-color-text-button-ghost);
 }
 
 .x-button--ghost.x-button--ghost-start.x-button,
 .x-button--ghost.x-button--ghost-start .x-button {
-  // spacing
-  padding-inline-start: 0;
+  // size
+  --x-size-padding-left-button-default: 0;
```
LuisMartinez15, line 12: I think this change and the other one below will "harm" the out-of-the-box `rtl` support.

LuisMartinez15 (changes requested):

packages/x-components/src/design-system/components/sliding-panel/default.scss
LuisMartinez15: Yep, `line 65` is what is breaking it:

```
&:not(.x-sliding-panel--no-gradient) {
    &::before,
    &::after {
      // layout
      position: absolute;
      z-index: 1;
      ...
```

packages/x-components/src/design-system/components/button/ghost.scss
tajespasarela: I changed the properties of `padding`, `border-width` and `margin`.
I left the names of the tokens to keep them more understandable, as this doesn't affect the CSS.

packages/x-components/src/design-system/components/sliding-panel/default.scss
tajespasarela: fixed with MORE z-index in the component

packages/x-components/src/design-system/components/list/default.scss
tajespasarela: fixed

tiborux (approved):

LuisMartinez15 (approved):

PR #59, opened 2021-07-27
Feature/ex 4435 update typescript

javieri-empathy:
# Description

Upgrades TS to 4.3.5 and TSLib to 2.3.0 in every package. API extractor and TS rollup plugin had to be upgraded too.

# Context

This is needed to sync the dependencies of the mono-repo, and to allow using new TS features like template strings.
In another task, linter should be upgraded too, it was too heavy to upgrade it in this task, as plugins include lots of new rules that should be reviewed, and enabled/customised/disabled. 
# Test

I have uploaded the needed changes for the [archetype in a branch](https://github.com/empathyco/x-archetype/tree/feature/EX-4480-update-mono-repo). You can install the local tarball files generated for each package in the mono-repo.

- Types are exported properly
- Docs are exported successfully
- Project keeps working
- You can use new TS features like template strings


packages/x-components/src/__tests__/utils.ts
```diff
@@ -125,34 +112,20 @@ export function getMockedAdapter(
   responseFeatures?: Partial<Omit<FeaturesResponseTypes, 'track'>>
 ): MockedSearchAdapter {
   return {
+    /* eslint-disable @typescript-eslint/no-non-null-asserted-optional-chain */
```
javieri-empathy, line 115: These types where lying, methods where not returning a `XXXResponse` object if the `responseFeatures.xxx` was undefined. But as this is just testing code, I think we can simplify it and just use non-null assertions rather than over-engineer types and mess up the code. But ask me once and I'll do it.

packages/x-components/src/plugins/x-plugin.ts
```diff
@@ -345,12 +345,8 @@ export class XPlugin implements PluginObject<XPluginOptions> {
...
       customizedStoreEmitters,
       (event, stateSelector: AnySimpleStateSelector | AnyStateSelector) => {
-        const {
-          selector,
-          immediate = false,
-          filter = () => true,
-          ...options
-        } = this.isSimpleSelector(stateSelector) ? { selector: stateSelector } : stateSelector;
+        const { selector, immediate, filter, ...options } =
+          this.normalizeStateSelector(stateSelector);
```
javieri-empathy, line 349: This was throwing an eslint error. It was complaining that we can't use unbound functions. To be honest I didn't understand why it was throwing that errors, `AnyStateSelector` and `AnySimpleSelector` are defined using the arrow function syntax, which should prevent these errors from happening. 
However I found that extracting this logic to a normalise function made TS stop complaining, so that is what I did.

packages/x-components/src/utils/__tests__/normalize.spec.ts
```diff
@@ -25,6 +25,8 @@ describe('testing string normalization', () => {
     it('trims and lowercases characters', () => {
       /* eslint-disable @typescript-eslint/unbound-method */
       const originalStringNormalize = String.prototype.normalize;
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-expect-error TS does not allow removing properties that are not optional.
```
javieri-empathy, line 29: Obviously, we shouldn't remove a property which is not optional. TS now detects this errors and complains about them. Again this is for a test that is testing exactly what happens if there is no `normalize` function available in strings. AFAIK this function is not available only in IE, which isn't a problem as we don't support it, and in Android webviews, which are a problem.

Can somebody with an Android device can try to check it using a browser like the Facebook one? I think this was one of the apps that included old webviews.

packages/x-components/src/utils/normalize.ts
```diff
@@ -8,9 +8,10 @@
  * @returns If the environment supports string normalization, the string without the combining
  * diacritical marks. Else the same string.
  */
-const removeCombiningDiacriticalMarks = String.prototype.normalize
-  ? (string: string) => string.normalize('NFD').replace(/[\u0300-\u036f]/g, '')
-  : (string: string) => string;
+const removeCombiningDiacriticalMarks =
+  typeof String.prototype.normalize === 'function'
```
javieri-empathy, line 12: TypeScript was complaining that the normalize function is always available, so the condition would always return true. This happens for modern browsers, but not for IE or android webviews. Changing the assertion made TS stop complaining, and now the assertion is also safer.

packages/x-components/src/wiring/utils/wire-racing-handling.ts
```diff
@@ -134,8 +134,11 @@ function updateRaceSubscription(
...
-function isDebounceRx(
-  timingOperatorRx: typeof debounceRx | typeof throttleRx
-): timingOperatorRx is typeof debounceRx {
+function isDebounceRx(timingOperatorRx: RXDebounce | RXThrottle): timingOperatorRx is RXDebounce {
   return timingOperatorRx === debounceRx;
 }
+
+/* Type hack to allow TS to infer correct types. If `__kind__` is removed, TS will infer `never` in
+`else` type guard clauses*/
+type RXDebounce = typeof debounceRx & { __kind__?: 'debounce' };
```
javieri-empathy, line 143: TS seems to have some serious problems when  using type guards for functions. I don't know how this was working before, I created this playground with the old 3.7.5 ts version and a simple example to illustrate this.

Fortunately, I came up with this hack that seems to do the trick. I don't like it, but It was the only solution that didn't imply changing the code, which was my initial intention as I don't understand why we have to pass the Rx function to use (either `throttle` or `debounce`) to the `handleTimedWire` function, and then this `handleTimedWire` function has two completely different paths for each of them.

packages/x-components/src/x-modules/search/components/search-items-injection.mixin.ts
```diff
@@ -19,12 +19,15 @@ export default class SearchItemsInjectionMixin extends Vue {
...
-   * @remarks It should be defined in the component that uses the mixin and it's intended to be
+   * @remarks It should be overridden in the component that uses the mixin and it's intended to be
    * filled with items from the state. Vue doesn't allow mixins as abstract classes.
-   *
+   * @returns An empty array as fallback in case it is not overridden.
    * @internal
    */
-  protected items!: SearchItem[];
+  @XProvide(SEARCH_ITEMS_KEY)
+  public get items(): SearchItem[] {
```
javieri-empathy, line 28: Now TS also complains if you are overriding a property with an accessor. The trick is to define the property here. The bonus part is that we can use the `XProvide` decorator here, and we don't have to use it in every component that uses this mixin.

packages/x-components/src/__tests__/utils.ts
LuisMartinez15: I wouldn't waste time with this, as we're going to refactor the adapter anytime soon(_tm_)

packages/react-wrapper/package.json
```diff
@@ -37,8 +37,8 @@
   "dependencies": {
     "react": "^16.9.0",
     "react-dom": "^16.9.0",
-    "tslib": "~1.11.1",
-    "vue": "~2.6.14"
+    "vue": "^2.6.0",
+    "tslib": "~2.3.0"
```
LuisMartinez15, line 41: Why did you reorder this? 😅

packages/x-components/src/utils/__tests__/normalize.spec.ts
LuisMartinez15: Good to know. Fortunately, I don't use android devices nor use facebook :trollface:

packages/x-components/src/x-modules/search/components/search-items-injection.mixin.ts
```diff
@@ -34,3 +37,7 @@ export default class SearchItemsInjectionMixin extends Vue {
   @XInject(SEARCH_ITEMS_KEY)
   public injectedSearchItems: SearchItem[] | undefined;
 }
+
+type Something = `Some ${string}`;
+/** @public */
+export const test: Something = 'Some test';
```
LuisMartinez15, line 43: This works like a charm :trollface:

```suggestion

```

LuisMartinez15 (changes requested):

packages/react-wrapper/package.json
javieri-empathy: Probably pressed the wrong keys, good catch

LuisMartinez15 (dismissed):

tajespasarela (commented):
Getting this warning on build, it seems we can disable that option in the rollup ts plugin:
image

https://github.com/egoist/bili/issues/305

CachedaCodes (approved):

tajespasarela (dismissed):

LuisMartinez15 (approved):

tajespasarela (approved):

PR #60, opened 2021-07-28
fix(promoted): fix name export of PromotedList component

tajespasarela:
EX-4484

# Pull Request Template #

Fix the Export name of the PromotedList component in the barrel file
Fix the vue-template-compiler version to use the latest minor version instead of the patch version


packages/x-components/src/x-modules/search/components/index.ts
```diff
@@ -3,7 +3,7 @@ export { default as BannersList } from './banners-list.vue';
 export { default as PartialQueryButton } from './partial-query-button.vue';
 export { default as PartialResultsList } from './partial-results-list.vue';
 export { default as Promoted } from './promoted.vue';
-export { default as PromotedLists } from './promoteds-list.vue';
+export { default as PromotedList } from './promoteds-list.vue';
```
tiborux, line 6:
```suggestion
export { default as PromotedsList } from './promoteds-list.vue';
```

tiborux (changes requested):

packages/x-components/src/x-modules/search/components/index.ts
LuisMartinez15: Well. Technically _promoteds_ doesn't exist. But we are using it in more places, so I'll leave up to you @tajespasarela

LuisMartinez15 (approved):

tiborux (approved):

PR #63, opened 2021-07-29
build: use 2 different versions of search-types

javieri-empathy:
# Description

Adds the searchs-types-next package which contains the new filter models.  This should install and not link the new package because it is the simplest solution:

- If we want to link the current search types, it should have the same dependency name: `@empathyco/x-types`. This means that we have to rename all current imports to another name like `@empathyco/x-types-old`. If this was only done in x-components it should be fine, but the adapter also complains because of the way dependencies are resolved (checking parent node_modules directories until one has the dependency). So I ended up with [a branch](https://github.com/empathyco/x/tree/feature/EX-4477-use-two-search-types) with way too many changes that we would have to revert in the future.
- Creating another module looked like a nice solution, but we had to configure another build, jest, and only for a couple of weeks.
- Given that we can easily release alphas, I think that the easiest way to achieve this is to just use the npm alpha versions, almost no code changes, and we shouldn't have to touch the search-types project. If we need to it is just bumping up a number.

# Context

This is needed to start working with the new facets module.

# Test

- Build keeps working
- You can use both new and old filters
- No more changes are needed


LuisMartinez15 (dismissed):

packages/x-components/build/api-extractor.json
```diff
@@ -2,7 +2,7 @@
   "$schema": "https://developer.microsoft.com/json-schemas/api-extractor/v7/api-extractor.schema.json",
   "projectFolder": "..",
   "mainEntryPointFilePath": "<projectFolder>/dist/types/index.d.ts",
-  "bundledPackages": ["@empathy/search-types", "@empathy/search-adapter"],
+  "bundledPackages": ["@empathyco/x-types", "@empathyco/x-adapter"],
```
LuisMartinez15, line 5: Good eye! 👀

LuisMartinez15 (approved):

tajespasarela (approved):

PR #66, opened 2021-07-29
test(e2e): mock adapter on next queries requests

tiborux:
In this task, we mocked the next queries but we are using the results in the view so I have to use a dummy to intercept the request and show some results.

Also, I clean a little bit the next queries view because we are showing a lot of components there but we are not testing them.

EX-4468


packages/x-components/tests/e2e/cucumber/next-queries/next-queries.spec.ts
```diff
@@ -18,14 +20,22 @@ Given(
...
-      }
-    });
   }
 );
 
+And('next queries response being mock {string}', (mockName: string) => {
+  cy.intercept('https://api.empathy.co/nextQueries', async req => {
+    const module = await import('./stubs/next-queries.stub');
+    req.reply(module[mockName as keyof typeof module]);
+  }).as('interceptedNextQueries');
```
javieri-empathy, line 30: Why are you storing the next queries request in a variable? I can't find any usage of it.

packages/x-components/tests/e2e/cucumber/next-queries/next-queries.spec.ts
```diff
@@ -18,14 +20,22 @@ Given(
...
+
+  cy.visit('/test/next-queries?useMockedAdapter=true', {
```
javieri-empathy, line 32: So every time you try to mock the next queries response it reloads the page?  Does that make sense? I think those should be separate steps.

Can't we intercept the nqs before loading the page if we have them in separate steps?

packages/x-components/tests/e2e/cucumber/next-queries/next-queries.spec.ts
```diff
@@ -18,14 +20,22 @@ Given(
...
-      qs: {
-        xModules: JSON.stringify(config)
-      }
-    });
   }
 );
 
+And('next queries response being mock {string}', (mockName: string) => {
+  cy.intercept('https://api.empathy.co/nextQueries', async req => {
+    const module = await import('./stubs/next-queries.stub');
```
javieri-empathy, line 28: The dynamic imports idea was made so we could have one stub per file. Some responses like the search one are going to have big stubs, and having multiple in a single file is going to be ugly. But if you are going to do it this way, it only adds unnecessary complexity. So either use dynamic imports with multiple files, or just import the whole module using a standard import.

packages/x-components/tests/e2e/cucumber/next-queries/next-queries.spec.ts
javieri-empathy: I think I understand now why you said that you needed to mock the search API. I would propose an alternative approach, which is extending the `EmpathyAdapter` until we have the full mocked adapter.

Something like this:

```
class E2ETestsAdapter extends EmpathyAdapter {
  getNextQueries(request: NextQueriesRequest): Promise<NextQueriesResponse> {
    return mockFetch(request, 'getNextQueries');
  }
  getTopRecommendations(request: TopRecommendationsRequest): Promise<TopRecommendationsResponse> {
    return mockFetch(request, 'getTopRecommendations');
  }
}

export const mockedAdapter = new EmpathyAdapterBuilder(
  undefined,
  undefined,
  E2ETestsAdapter
).build();
```

Then only the overridden methods are mocked, the remaining ones can continue with its standard flow. We can also remove the `useMockedAdapter` parameter from the URL and its related code.

packages/x-components/tests/e2e/cucumber/next-queries/next-queries.spec.ts
```diff
@@ -18,14 +20,22 @@ Given(
...
     };
-    cy.visit('/test/next-queries', {
-      qs: {
-        xModules: JSON.stringify(config)
-      }
-    });
   }
 );
 
+And('next queries response being mock {string}', (mockName: string) => {
```
javieri-empathy, line 26: Change the text with something like: `next queries API should respond with {string}`

packages/x-components/tests/e2e/cucumber/next-queries/next-queries.feature
```diff
@@ -2,47 +2,53 @@ Feature: Next queries component
...
     Then  related results are displayed
     And   next queries are displayed
     When  clear search button is pressed
     Then  next queries are still displayed
 
     Examples:
-      | hideSessionQueries | maxItemsToRequest | loadOnInit | query     |
-      | true               | 5                 | true       | playmobil |
-      | true               | 5                 | false      | playmobil |
+      | hideSessionQueries | maxItemsToRequest | loadOnInit | query     | mock            | mockedSearch            |
```
javieri-empathy, line 51: Rename the `mock` and `mockedSearch` parameters to match: `nextQueriesMock`, `searchMock`

javieri-empathy (changes requested):

packages/x-components/tests/e2e/cucumber/next-queries/next-queries.spec.ts
javieri-empathy: Given that NQs are now static, and how complex it seems to be to allow using dynamic imports inside a Cypress spec file, we can just resolve statically the nqs stub.

packages/x-components/tests/e2e/cucumber/next-queries/stubs/next-queries.stub.ts
```diff
@@ -0,0 +1,9 @@
+import { createNextQueryStub } from '../../../../../src/__stubs__/next-queries-stubs.factory';
+
+export const nextQueriesStub = {
```
javieri-empathy, line 3: Add the `NextQueriesResponse` type

javieri-empathy: And I don't know where to put this, but be careful with imports, for exmaple in `recommendations.spec` you are importing from a relative package, instead of node modules.

`import { TopRecommendationsRequest } from '../../../../../search-adapter/src/types/requests.types';`
should be
`import { TopRecommendationsRequest } from '@empathyco/x-adapter';`

packages/x-components/tests/e2e/cucumber/recommendations/recommendations.spec.ts
```diff
@@ -1,24 +1,21 @@
...
+Given('suggestions API should respond with mocked suggestions', () => {
+  cy.intercept('https://api.empathy.co/getTopRecommendations', req => {
+    req.reply(recommendationsStub);
+  }).as('interceptedRecommendations');
+});
 
-Given('following config: max items to store is {int}', (maxItemsToRequest: number) => {
-  config = {
+And('following config: max items to store is {int}', (maxItemsToRequest: number) => {
+  const config = {
```
javieri-empathy, line 11: Never define constant objects without a type. This is a pain point when we do refactors becase neither TS or us knows what type this object should match. In this case it should be `InstallXOptions['xModules']`

packages/x-components/tests/e2e/cucumber/next-queries/next-queries.spec.ts
```diff
@@ -1,10 +1,17 @@
...
+  cy.intercept('https://api.empathy.co/getNextQueries', req => {
+    req.reply(nextQueriesStub);
+  }).as('interceptedNextQueries');
+});
+
+And(
   'following config: hide session queries {boolean}, requested items {int}, loadOnInit {boolean}',
   (hideSessionQueries: boolean, maxItemsToRequest: number, loadOnInit: boolean) => {
-    const config: InstallXOptions['xModules'] = {
+    const config = {
```
javieri-empathy, line 13: Add the missing type `InstallXOptions['xModules']`

javieri-empathy (changes requested):

javieri-empathy (approved):

guillei10 (approved):

PR #67, opened 2021-07-30
Feature/ex 3655 add next facets store

javieri-empathy:
# Description

Adds the store for the facets. In this case this only means adding the getters, as this module has no actions:
image

# Context

This is needed as part of the new facets module

# Test

Check the getters works as expected as defined by the tests spec files. If you find any other use case missing tell me and I'll add it.


packages/x-components/src/x-modules/facets-next/store/__tests__/getters.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+import {
+  FacetsNextActions,
+  FacetsNextGetters,
+  FacetsNextMutations,
+  FacetsNextState
+} from '../types';
+import { resetFacetsStateWith } from './utils';
+
+/** Type safe single module store. */
+type SafeStore<
```
javieri-empathy, line 25: This was my firefighter morning idea 🤣. I realised we already have kinda a type safe store, so we can avoid having to cast, or to use the actions/getters/mutations keys as variables.

packages/x-components/src/x-modules/facets-next/store/__tests__/getters.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+        createNextEditableNumberRangeFilter('age', { min: null, max: 5 }),
+        createNextEditableNumberRangeFilter('size', { min: null, max: null }),
+        // Raw filters are always selected, but shouldn't be returned in this getter.
+        createRawFilter('size:xl')
+      ]);
+
+      //expect(store.getters.selectedFilters).toHaveLength(4);
+      expect(store.getters.selectedFilters).toEqual(
+        expect.arrayContaining([
+          store.state.filters['color:Blue'],
```
javieri-empathy, line 67: I'm a little bit concerned about having to know how ids are created in the stubs. Is this too much knowledge for this test? The other way of creating the filters in variables and then passing them to the store was too verbose I think.

packages/x-components/src/x-modules/facets-next/store/__tests__/getters.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+      const store = createFacetsStore([
+        createNextSimpleFilter('color', 'Red', false),
+        createNextSimpleFilter('color', 'Blue', true),
+        createNextHierarchicalFilter('category', 'Summer', false),
+        createNextHierarchicalFilter('category', 'Shorts', true),
+        createNextNumberRangeFilter('price', { min: 0, max: 25 }, false),
+        createNextNumberRangeFilter('price', { min: 25, max: 50 }, true),
+        createNextEditableNumberRangeFilter('age', { min: null, max: 5 }),
+        createNextEditableNumberRangeFilter('size', { min: null, max: null }),
+        // Raw filters are always selected, but shouldn't be returned in this getter.
```
LuisMartinez15, line 138:
```suggestion
        // Raw filters don't belong to a facet so they're not retrieved by this getter
```

LuisMartinez15 (changes requested):

packages/x-components/src/x-modules/facets-next/store/__tests__/getters.spec.ts
javieri-empathy: Not copy/paste at all

packages/x-components/src/x-modules/facets-next/store/__tests__/getters.spec.ts
LuisMartinez15: I agree. Maybe we can put it in the `createNextSimpleFilter` doc. Something like:

```
/**
 * Creates a {@link @empathyco/x-types-next#SimpleFilter | SimpleFilter}.
 *
 * @remarks The created simple filter id is going to be the following: <facetId>:<label>.
 *
 * @param facetId - The facet id this filter belongs to.
 * @param label - The text that this filter should display.
 * @param selected - True when the filter is checked, false otherwise.
 * @returns A stub for a {@link @empathyco/x-types-next#SimpleFilter | SimpleFilter}.
 */
export function createNextSimpleFilter(
...
```

LuisMartinez15 (dismissed):

packages/x-components/src/x-modules/facets-next/store/__tests__/getters.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+        createRawFilter('size:xl')
+      ]);
+
+      //expect(store.getters.selectedFilters).toHaveLength(4);
+      expect(store.getters.selectedFilters).toEqual(
+        expect.arrayContaining([
```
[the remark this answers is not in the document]
javieri-empathy: Because I don't care about the order, the order of selection does not matter in this test

packages/x-components/src/x-modules/facets-next/store/getters/selected-filters.getter.ts
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * @param state - Current {@link https://vuex.vuejs.org/guide/state.html | state} of the facets
+ * module.
+ *
+ * @returns A list containing the selected filters.
+ *
+ * @public
+ */
+export const selectedFilters: FacetsNextXStoreModule['getters']['selectedFilters'] = state =>
+  Object.values(state.filters).filter(filter => isFacetFilter(filter) && filter.selected);
```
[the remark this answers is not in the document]
javieri-empathy: I added it but in the getter doc

packages/x-components/src/x-modules/facets-next/store/__tests__/getters.spec.ts
```diff
@@ -0,0 +1,194 @@
...
+
+    it('returns a record containing the editable number range filters indexed by its id', () => {
+      const store = createFacetsStore([
+        createNextEditableNumberRangeFilter('age', { min: null, max: 5 })
+      ]);
+
+      expect(store.getters.filtersByFacet).toEqual({
+        age: [store.state.filters['age:*-5']]
+      });
+    });
```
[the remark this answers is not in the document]
javieri-empathy: I'll add it, just in case. I discussed this with @tajespasarela, and he didn't 100% convince me, I still see lots of duplicity between these tests, but I agree that one is testing mixing different filters, and the error will be harder to spot, while others could help you debug easily as you have less data in the store.

LuisMartinez15 (approved):

davidmfempathy (approved):

PR #75, opened 2021-08-03
feat(facets-next): add `HierarchicalFilterEntity`

javieri-empathy:
# Description

Adds the `HierarchicalFilterEntity` following the task diagram.

# Context

This is needed as part of the new facets module.

# Test

Check that the hierarchical filter entity works as expected, supporting multiselection by default.


pmareke (approved):
LGTM

packages/x-components/src/x-modules/facets-next/entities/hierarchical-filter.entity.ts
```diff
@@ -0,0 +1,73 @@
...
+  /**
+   * Saves the given filter to the store.
+   *
+   * @param filter - The filter to save to the store.
+   * @internal
+   */
+  protected saveFilter(filter: HierarchicalFilter): void {
+    this.store.commit('x/facetsNext/setFilter', filter);
+  }
+}
```
tajespasarela, line 73: Maybe create a method to cast only once?
```suggestion

  protected getFilterById(id: HierarchicalFilter['id']): HierarchicalFilter {
    return this.store.state.x.facetsNext.filters[id] as HierarchicalFilter;
  }
}
```

packages/x-components/src/x-modules/facets-next/entities/hierarchical-filter.entity.ts
```diff
@@ -0,0 +1,73 @@
...
+   * Deselects all the descendants of the given filter, saving them to the store.
+   *
+   * @remarks The descendants hierarchy is retrieved from the store.
+   * @param filter - The filter to deselect its descendants.
+   * @internal
+   */
+  protected deselectDescendants(filter: HierarchicalFilter): void {
+    if (filter.children) {
+      filter.children.forEach(childId => {
+        const child = this.store.state.x.facetsNext.filters[childId] as HierarchicalFilter;
```
tajespasarela, line 57:
```suggestion
        const child = this.getFilterById(childId);
```

packages/x-components/src/x-modules/facets-next/entities/hierarchical-filter.entity.ts
```diff
@@ -0,0 +1,73 @@
...
+  /**
+   * Selects all the ancestors of the given filter, saving them to the store.
+   *
+   * @remarks The ancestors hierarchy is retrieved from the store.
+   * @param filter - The filter to select its ancestors.
+   * @internal
+   */
+  protected selectAncestors(filter: HierarchicalFilter): void {
+    if (filter.parentId) {
+      const parent = this.store.state.x.facetsNext.filters[filter.parentId] as HierarchicalFilter;
```
tajespasarela, line 41:
```suggestion
      const parent = this.getFilterById(filter.parentId);
```

tajespasarela (changes requested):

tajespasarela (dismissed):

packages/x-components/src/__stubs__/filters-stubs.factory.ts
```diff
@@ -375,3 +375,49 @@ export function createNextEditableNumberRangeFilter(
...
+
+/**
+ * Creates a factory of
+ * {@link @empathyco/x-types-next#HierarchicalFilter | HierarchicalFilter}s.
+ *
+ * @param facetId - Facet id to which the filter belongs.
+ * @param parentId - Filter's parent id if exists.
+ * @returns A scoped function which is able to create `HierarchicalFilters`.
+ */
+export function createNextHierarchicalFilterFactory(
```
tajespasarela, line 398: Can we use this factory to define the `createNextHierarchicalFilter` function above?

packages/x-components/src/x-modules/facets-next/entities/hierarchical-filter.entity.ts
```diff
@@ -0,0 +1,83 @@
...
+   * Saves the given filter to the store.
+   *
+   * @param filter - The filter to save to the store.
+   * @internal
+   */
+  protected saveFilter(filter: HierarchicalFilter): void {
+    this.store.commit('x/facetsNext/setFilter', filter);
+  }
+
+  protected getFilterById(id: HierarchicalFilter['id']): HierarchicalFilter {
```
tajespasarela, line 80: documentation here?

packages/x-components/src/__stubs__/filters-stubs.factory.ts
```diff
@@ -375,3 +375,49 @@ export function createNextEditableNumberRangeFilter(
...
+      : [];
+    return [
+      {
+        id: filterId,
+        facetId: facetId,
+        parentId,
+        selected,
+        label,
+        totalResults: 10,
+        callbackInfo: {},
```
tajespasarela, line 415: remove the `callbackInfo` field

packages/x-components/src/__stubs__/filters-stubs.factory.ts
```diff
@@ -375,3 +375,49 @@ export function createNextEditableNumberRangeFilter(
...
+ *
+ * @param facetId - Facet id to which the filter belongs.
+ * @param parentId - Filter's parent id if exists.
+ * @returns A scoped function which is able to create `HierarchicalFilters`.
+ */
+export function createNextHierarchicalFilterFactory(
+  facetId: string,
+  parentId: string | null = null
+): CreateNextHierarchicalFilter {
+  return (label, selected, createChildren) => {
```
tajespasarela, line 402: If you had added the type, then you would realize about the `callbackInfo` and `value` fields
```suggestion
  return (label, selected, createChildren): NextHierarchicalFilter[] => {
```

packages/x-components/src/__stubs__/filters-stubs.factory.ts
```diff
@@ -375,3 +375,49 @@ export function createNextEditableNumberRangeFilter(
...
+      {
+        id: filterId,
+        facetId: facetId,
+        parentId,
+        selected,
+        label,
+        totalResults: 10,
+        callbackInfo: {},
+        children: children.map(filter => filter.id),
+        value: label.toLowerCase().replace(/\s+/g, '-'),
```
tajespasarela, line 417: remove the `value` field

packages/x-components/src/x-modules/facets-next/entities/__tests__/utils.ts
```diff
@@ -0,0 +1,53 @@
...
+}
+
+/**
+ * Creates a store with the {@link facetsNextXStoreModule} registered, and initialises its
+ * state with the provided filters.
+ *
+ * @param filters - The filters to set to the {@link facetsNextXStoreModule} module state.
+ * @returns A store with the {@link facetsNextXStoreModule} registered, and the provided facets.
+ */
+export function prepareFacetsStoreWithFilters(filters: Filter[]): Store<RootXStoreState> {
```
tajespasarela, line 36: what about using only one function `prepareFacetsStore` with optional `filters: Filter[]` parameter?

tajespasarela (changes requested):

packages/x-components/src/x-modules/facets-next/entities/hierarchical-filter.entity.ts
javieri-empathy: That is what happens when you accept suggestions in the PR. No more github suggestions, manual copy and paste ftw

packages/x-components/src/__stubs__/filters-stubs.factory.ts
javieri-empathy: I didn't know it was using relaxed types when it inferred return types. Good catch

packages/x-components/src/__stubs__/filters-stubs.factory.ts
javieri-empathy: We tried, but I wouldn't: this factory creates a hierarchy of filters, and returns them flattened, while the other method, `createNextHierarchicalFilter` only creates a single filter and returns it. They have different purposes.

tajespasarela (approved):

pmareke (approved):

PR #73, opened 2021-08-03
feat(facets-next): add `RawFilterEntity`

guillei10:
## Description ##

Adds the RawFilterEntity to the project. This entity contains the business logic to select/deselect RawFilter DTOs.

## Context ##

This is needed as part of the new facets module, to encapsulate filters logic together.

PR #76, opened 2021-08-03
feat(facets-next): add `EditableNumberRangeFilterEntity`

davidmfempathy:
# Description

Add EditableNumberRangeFilterEntity.

# Test
- Check functionality.
- Check if unit tests cover all the posible cases.

https://searchbroker.atlassian.net/browse/EX-4356


packages/x-components/src/x-modules/facets-next/entities/__tests__/editable-number-range-filter-entity.spec.ts
```diff
@@ -0,0 +1,71 @@
+import { createLocalVue } from '@vue/test-utils';
+import Vuex, { Store } from 'vuex';
+import { EditableNumberRangeFilter } from '@empathyco/x-types';
+import { createNextEditableNumberRangeFilter } from '../../../../__stubs__/filters-stubs.factory';
+import { RootXStoreState } from '../../../../store/store.types';
+import { facetsNextXStoreModule } from '../../store/module';
+import { EditableNumberRangeFilterEntity } from '../editable-number-range-filter.entity';
+
+describe('testing EditableNumberRangeFilterEntity', () => {
+  function prepareFacetsStore(): Store<RootXStoreState> {
```
tajespasarela, line 10: This method was moved to a utils file in this PR: https://github.com/empathyco/x/pull/73

Maybe it's a good idea

tajespasarela (changes requested):

packages/x-components/src/x-modules/facets-next/entities/__tests__/editable-number-range-filter-entity.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+          },
+          namespaced: true
+        }
+      }
+    });
+  }
+
+  it('allows selecting a filter when filter min is not null', () => {
+    const store = prepareFacetsStore();
+    const filter = createNextEditableNumberRangeFilter('price', { min: 3, max: null });
```
javieri-empathy, line 27: With this utility that I updated from the old `createEditableNumberRangeFilter`, the selected value is set immediately depending on the range value. Maybe we can add a third optional parameter to it, `selected = range.min !== null || range.max !== null`, and force `selected` to `false` initially in this test?

packages/x-components/src/x-modules/facets-next/entities/__tests__/editable-number-range-filter-entity.spec.ts
```diff
@@ -0,0 +1,71 @@
...
+    const store = prepareFacetsStore();
+    const filter = createNextEditableNumberRangeFilter('price', { min: 2, max: 10 });
+    const filterEntity = new EditableNumberRangeFilterEntity(store, filter);
+
+    filterEntity.select();
+    expect(store.state.x.facetsNext.filters[filter.id].selected).toBe(true);
+
+    filterEntity.deselect();
+    expect(store.state.x.facetsNext.filters[filter.id].selected).toBe(false);
+    expect(
```
javieri-empathy, line 53: Because of the verbosity of these tests, we created a `isFilterSelected` method, I'm going to move it to the `utils.ts` right now. Maybe you could do the same for the range? Just to avoid casting too many times, like @tajespasarela is suggesting for the entity code.

javieri-empathy (changes requested):
Check comments

packages/x-components/src/x-modules/facets-next/entities/editable-number-range-filter.entity.ts
```diff
@@ -0,0 +1,55 @@
...
+  EditableNumberRangeFilter,
+  Filter,
+  isEditableNumberRangeFilter
+} from '@empathyco/x-types-next';
+import { Store } from 'vuex';
+import { EquatableFilter } from './equatable-filter';
+import { FilterEntity } from './types';
+
+/**
+ * Allows selecting and deselecting a filter of {@link EditableNumberRangeFilter}.
```
guillei10, line 11: We should decide wether we document the implementation or not in the entities. Currently `SimpleFilterEntity` and `RawFilterEntity` ([in PR](https://github.com/empathyco/x/pull/73)) are not documented but this is.

I suggest to remove these docs despite each implementation is different. At the end  each entity is "selecting and deselecting" on its own.

guillei10 (changes requested):

packages/x-components/src/x-modules/facets-next/entities/__tests__/editable-number-range-filter-entity.spec.ts
davidmfempathy: Done.

packages/x-components/src/x-modules/facets-next/entities/__tests__/editable-number-range-filter-entity.spec.ts
davidmfempathy: I've also added checks for the initial filter.selected value which could be redundant but make the tests more explicit. Tell me if we don't need them at all.

packages/x-components/src/x-modules/facets-next/entities/editable-number-range-filter.entity.ts
davidmfempathy: As we've discussed during the stand up, the rest of entities must include this docs in order to explain how they do the select / deselect process.

packages/x-components/src/x-modules/facets-next/entities/__tests__/editable-number-range-filter-entity.spec.ts
tajespasarela: instead of only using the `isFilterSelected` function, maybe it is better to create a new `isRangeSelected` function where to check, not only if the `selected` field is `true` or `false`, but also if the `range` has or not values.

packages/x-components/src/x-modules/facets-next/entities/__tests__/editable-number-range-filter-entity.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+
+describe('testing EditableNumberRangeFilterEntity', () => {
+  it('allows selecting a non selected filter when its min value is not null', () => {
+    const store = prepareFacetsStore();
+    const filter = createNextEditableNumberRangeFilter('price', { min: 3, max: null }, false);
+    const filterEntity = new EditableNumberRangeFilterEntity(store, filter);
+
+    expect(filter.selected).toBe(false);
+    filterEntity.select();
+    expect(isFilterSelected(store, filter.id)).toBe(true);
```
tajespasarela, line 14: check also the range values are the same

packages/x-components/src/x-modules/facets-next/entities/__tests__/editable-number-range-filter-entity.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+  });
+
+  it('allows selecting a non selected filter when its max value is not null', () => {
+    const store = prepareFacetsStore();
+    const filter = createNextEditableNumberRangeFilter('price', { min: null, max: 5 }, false);
+    const filterEntity = new EditableNumberRangeFilterEntity(store, filter);
+
+    expect(filter.selected).toBe(false);
+    filterEntity.select();
+    expect(isFilterSelected(store, filter.id)).toBe(true);
```
tajespasarela, line 24: check also the range values are the same

tajespasarela (changes requested):

packages/x-components/src/__stubs__/filters-stubs.factory.ts
```diff
@@ -360,18 +360,20 @@ export function createNextNumberRangeFilter(
...
  * @param range - The range that this filter has.
+ * @param selected - The selected value which has priority over the range values.
  * @returns A stub for a
  * {@link @empathyco/x-types-next#EditableNumberRangeFilter | EditableNumberRangeFilter}.
  */
 export function createNextEditableNumberRangeFilter(
   facetId: string,
-  range: RangeValue = { min: null, max: null }
+  range: RangeValue = { min: null, max: null },
+  selected: boolean
```
tajespasarela, line 370: make `selected` parameter optional

tajespasarela (changes requested):

javieri-empathy (approved):

PR #77, opened 2021-08-03
Publish

javieri-empathy:
This is supposed to correct the version. An alpha version was tried to be released, and it failed because the x-translations package was not yet meant to be published. As a result some packages were published, and others were not, but because of the failure, no changes were pushed  here. This PR just updates the version and the tag of every package, so the next release does not fail because it has incorrect version numbers. It does not fix the issue of trying to publish private packages.

**Merge this PR with "merge commit"**

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,6 +1,7 @@
 {
   "packages": ["packages/*"],
   "version": "independent",
+  "ignoreChanges": ["**/package-lock.json"],
   "changelogPreset": {
     "name": "conventionalcommits",
     "issuePrefixes": ["EX-"],
--- a/packages/x-translations/CHANGELOG.md
+++ b/packages/x-translations/CHANGELOG.md
@@ -5,17 +5,9 @@ See [Conventional Commits](https://conventionalcommits.org) for commit guideline
 
 ## 1.1.0-alpha.0 (2021-08-03)
 
-
-### Features
-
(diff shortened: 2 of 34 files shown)


packages/x-translations/CHANGELOG.md
```diff
@@ -1,3 +1,20 @@
...
+
+All notable changes to this project will be documented in this file.
+See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
+
+## 1.1.0-alpha.0 (2021-08-03)
+
+
+### Features
+
+* **facets-next:** add `SimpleFilterEntity` ([9b551a4](https://github.com/empathyco/x/commit/9b551a41d6e67c64942cfe26051a06ce7dc91a65)), closes [EX-4352](https://searchbroker.atlassian.net/browse/EX-4352)
```
pmareke, line 11: Why do we have this change here? I think we should remove it.

pmareke (changes requested):
Great job, thanks!

pmareke (dismissed):

lerna.json
```diff
@@ -1,6 +1,7 @@
 {
   "packages": ["packages/*"],
   "version": "independent",
+  "ignoreChanges": ["**/package-lock.json"],
```
pmareke, line 4: Nice!

pmareke (approved):

davidmfempathy (approved):

PR #81, opened 2021-08-06
Ex 3651 add entity filter modifiers

javieri-empathy:
# Description

Adds the filters factory, and the first single select modifier. The filters factory allows to generate entities for a given type of filters, and to apply modifiers to them. One sample modifier has been created (`SingleSelectModifier`), which makes any kind of facet allow only one selected filter at a time.

To decouple entities from the store, filter entities have been modified, and now they don't receive the filter in the constructor, but in the `select` and `deselect` methods. This assumes that if an entity or a modifier has to make some operation over another relative filter (parent, child, sibling), this filter will have the same entity type, and the same modifiers applied.

To summarize, entities depend on the `modelName`, of the filter., and modifiers depend on the `facetId`, therefore it makes no sense to apply them to `RawFilters` for example. However as making this decorating logic type safe was too messy, we provided with a test to ensure that at least the app won't explode damaging users devices.

# Context

This is needed as part of the new facets module. The factory is the only place that should have different logic depending on the type of filter it is creating.

# Test

Several use cases have been provided. Check that they make sense, and if we are missing something.


packages/x-components/src/x-modules/facets-next/entities/__tests__/equatable-filter.spec.ts
```diff
@@ -1,20 +0,0 @@
-import { createNextSimpleFilter } from '../../../../__stubs__/filters-stubs.factory';
```
javieri-empathy, line 1: Equatable filter has been removed. The initial idea of this was to have a way to compare the same kind of filters. The editable number range filter was supposed to have a static Id, so we had to compare them using the range values, but we change it in this PR, and its id is now modified depending on the values.  So now we can just compare the filters id.

packages/x-components/src/x-modules/facets-next/entities/factory.ts
```diff
@@ -0,0 +1,87 @@
...
+   *
+   * @remarks The constructor is not private, so it's possible to create more instances,
+   * to simplify the testing.
+   */
+  public static instance = new FiltersFactory();
+
+  /**
+   * The registered entities by default to be used by the Factory.
+   */
+  protected entities: FilterEntityConstructor[] = [
```
javieri-empathy, line 26: To simplify configuration in the client projects, we decided to include them here by default. The trade of is that if there aren't filters compatible with one of them, you will still have them in the bundle. There won't be tree shaking for them. But as they are really small, and the common case should be to have 3/4 of them, we can accept it for now.

Another thing to note is that this array is order dependant. Right now entities have a 1/1 relationship with filter model names, but that won't be like this in the future. For example `EditableNumberRangeFilterEntity` should be able to accept `RawFilter` which its ids matches a regular expression.

packages/x-components/src/x-modules/facets-next/entities/single-select.modifier.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  /**
+   * Retrieves the rest of selected filters of the same facet than the one given.
+   *
+   * @param filter - The filter to find its relatives.
+   * @returns A list of selected filters that belong to the same facet than the filter passed.
+   * @internal
+   */
+  protected getOtherFilters(filter: FacetFilter): FacetFilter[] {
+    return isFacetFilter(filter)
```
javieri-empathy, line 27: This seems dumb because we are checking if a `FacetFilter` is a `FacetFilter`. The problem is that when you decorate some entity using the factory you loos this `FacetFilter` type in the select method. This decorator shouldn't be used with `RawFilters`, but we added this simple check as a crash prevention measure.

packages/x-components/src/x-modules/facets-next/entities/__tests__/factory.spec.ts
```diff
@@ -0,0 +1,233 @@
...
+    const rawFilter = createRawFilter('size:m');
+    const rawFilterEntity = factory.createFilterEntity(store, rawFilter);
+
+    // Selecting a raw filter that is not in the store should add it.
+    rawFilterEntity.select(rawFilter);
+    expect(isFilterSelected(store, rawFilter.id)).toBe(true);
+
+    // Deselecting a raw filter should remove it from the store.
+    rawFilterEntity.deselect(rawFilter);
+    expect(store.state.x.facetsNext.filters).not.toHaveProperty(rawFilter.id);
```
tajespasarela, line 35: to make the same in all tests and not depends on the store structure:
```suggestion
    expect(getStoreFilter(store, rawFilter.id)).toBeUndefined();
```

packages/x-components/src/x-modules/facets-next/entities/factory.ts
```diff
@@ -0,0 +1,87 @@
...
+      return entity;
+    }
+  }
+
+  /**
+   * Registers a new {@link FilterEntity} to be used by the factory.
+   *
+   * @param entity - The new {@link FilterEntity} to be registered in the factory.
+   */
+  registerFilterEntity(entity: FilterEntityConstructor): void {
```
tajespasarela, line 70: should we check if the entity is not already registered? just to prevent errors.

packages/x-components/src/x-modules/facets-next/entities/factory.ts
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  /**
+   * Registers a list of modifiers to be used with the {@link FilterEntity | FilterEntities} of a
+   * particular facet.
+   *
+   * @param facetId - The facet id whose Entities will be modified.
+   * @param modifiers - The list of modifiers to be registered.
+   */
+  registerFilterModifier(facetId: Facet['id'], modifiers: FilterEntityModifier[]): void {
```
tajespasarela, line 81: should we check if the modifier is not already registered? just to prevent errors.

tajespasarela (changes requested):

packages/x-components/src/x-modules/facets-next/entities/single-select.modifier.ts
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  /**
+   * Retrieves the rest of selected filters of the same facet than the one given.
+   *
+   * @param filter - The filter to find its relatives.
+   * @returns A list of selected filters that belong to the same facet than the filter passed.
+   * @internal
+   */
+  protected getOtherFilters(filter: FacetFilter): FacetFilter[] {
```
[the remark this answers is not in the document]
javieri-empathy: These are not the siblings, which I understand as same level filters AKA "brothers". These are parents, grandparents, "brothers", childs...

packages/x-components/src/x-modules/facets-next/entities/editable-number-range-filter.entity.ts
```diff
@@ -22,34 +21,59 @@ export class EditableNumberRangeFilterEntity extends EquatableFilter implements
...
-  select(): void {
+  select(filter: EditableNumberRangeFilter): void {
+    const newFilterId = this.getNewFilterId(filter);
+    this.removePreviousFilter(filter.facetId);
     this.store.commit('x/facetsNext/setFilter', {
-      ...this.filter,
-      selected: this.isSelected()
+      ...filter,
+      id: newFilterId,
+      selected: this.isSelected(filter)
```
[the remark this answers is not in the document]
javieri-empathy: What do you mean? We remove the previous filter, but we always add a new filter to the store, we don't care if it is selected or not to decide if we should add it. Or are you talking about the `RawFilter`?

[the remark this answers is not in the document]
tajespasarela: Never mind, I don't know what I was thinking about here...

tajespasarela (approved):

pmareke (approved):

PR #82, opened 2021-08-06
Feature/ex 3640 search adapter refactor

guillei10:
## Description ##

Refactor of the `search-adapter` to fit all the new x-types for filters. This means:

- Change the imports to new types
- Refactor some response mapper to be able to map the correct properties from the API.
- Refactor request request mappers to map the new filter types to strings that will be accepted by the API. 
- Change the way we construct the `HierarchicalFacet`

## Motivation ##

This task is part of the big refactor for Filters.

## Tests ##

Check all the changes between the new properties and the old ones.
Execute new updated tests -specially for the `HierarchicalFacets`-


packages/search-adapter/src/empathy/__tests__/response-mappers/empathy-facet-mapper.spec.ts
```diff
@@ -47,21 +47,31 @@ afterEach(() => {
 describe('Empathy Hierarchical Facet', () => {
   it('maps a hierarchical facet', () => {
     const mappedFacet = mapFacet(hierarchicalRawFacet, {} as HierarchicalFacet, emptyContext) as HierarchicalFacet;
-
```
pmareke, line 50: There is an error in the line 39

image

Regarding `tslib`, please take a look.

pmareke (commented):
There are an error during the instalation

image

Please would you mind to take a look?

packages/search-adapter/src/empathy/mappers/request/params/empathy-request-filters-solr-syntax.mapper.ts
```diff
@@ -72,7 +74,18 @@ export class EmpathyRequestFiltersSolrSyntaxMapper implements Mapper<Filter[], s
...
   /**
-   * Check if the filters passed are of type {@link @empathyco/x-types#EditableNumberRangeFilter | EditableNumberRangeFilter}.
+   * Check if the filters passed are of type {@link @empathyco/x-types-next#RawFilter | RawFilter}.
+   *
+   * @param filters - The array of filters to check.
+   *
+   * @internal
+   */
+  private isArrayOfRawFilters(filters: Filter[]): filters is RawFilter[] {
+    return isRawFilter(filters[0]);
```
pmareke, line 84: I'm not sure about this method name, what do you want to check? I mean, the description says one thing but the name another.

packages/search-adapter/src/empathy/mappers/response/facets/empathy-hierarchical-facet.mapper.ts
```diff
@@ -30,16 +30,25 @@ export class EmpathyHierarchicalFacetMapper implements ResponseMapper<EmpathyFac
...
-    });
+  protected mapDeepFilter(
+    rawFilter: EmpathyFilter,
+    initFilterProps: Readonly<Partial<HierarchicalFilter>>,
+    context: ResponseMapperContext,
+    filters: HierarchicalFilter[]
+  ): string {
+    const filter = this.mapFilter(rawFilter, { ...initFilterProps } as HierarchicalFilter, context);
+    filter.children =
+      rawFilter.values?.map(
```
pmareke, line 49: If you have to `.map()` and `.push()` you can use the` .reduce()` in order to simplify the implementation.

pmareke (changes requested):

pmareke:
Please fix the tests, I tried to run the build again by they are still failing.

packages/search-adapter/src/empathy/mappers/request/params/empathy-request-filters-solr-syntax.mapper.ts
davidmfempathy: I guess it's just type narrowing.

packages/search-adapter/src/empathy/mappers/request/params/empathy-request-filters-solr-syntax.mapper.ts
```diff
@@ -29,8 +31,8 @@ export class EmpathyRequestFiltersSolrSyntaxMapper implements Mapper<Filter[], s
   }
 
   private composeFilters(filters: Filter[]): string[] {
-    if (this.isArrayOfBooleanFilters(filters)) {
-      return filters.map(filter => filter.value);
+    if (this.isArrayOfBooleanFilters(filters) || this.isArrayOfRawFilters(filters)) {
+      return filters.map(filter => `${filter.id}`);
```
davidmfempathy, line 35: Just for my info, if we add more types in the future we can just treat the EditableNumberRangeFilterList as a specific case and the rest in an else statement ? Or are there a type of filters which don't need to be mapped here ?

packages/search-adapter/src/empathy/mappers/response/facets/empathy-hierarchical-facet.mapper.ts
davidmfempathy: The `.map` is for the `rawFilter.values` in order to transform them into valid DTOs added to the `filter.children`.
The `.push` is for adding the filter which children have been mapped to the `filters` passed as reference.

The reduce version would be more complex.

```
    filter.children = rawFilter.values?.reduce((accFilters, rawFilterChild) => {
      accFilters.push(this.mapDeepFilter(rawFilterChild, { ...initFilterProps, parentId: filter.id }, context, filters)));
      return accFilters;
    }, []);
 ```
 
Or did you think something different?

File: packages/search-adapter/src/empathy/__tests__/response-mappers/empathy-facet-mapper.spec.ts [hunk 5]
```diff
@@ -107,21 +117,25 @@ function expectFacetToMatchMock(mappedFacet: HierarchicalFacet | SimpleFacet | N
...
     expect(filter).toMatchObject(schema);
-    if (isHierarchicalFilter(filter) && filter.children) {
-      expectAllFiltersToBeValid(filter.children, HierarchicalFilterSchema);
-    }
   });
 }
 
-function isHierarchicalFilter(filter: Filter): filter is HierarchicalFilter {
-  return filter.modelName === 'HierarchicalFilter';
+function countHierarchicalChildren(filter: EmpathyFacet | EmpathyFilter): number {
```
javieri-empathy, line 129: This is not a good testing practice. Tests should remain simple, it is preferable to hardcode the number of filters rather that to have a recursive function that counts them. Who is asserting that this function is counting them properly?

The only good reason to automate this cont is to avoid changing it if the mock is changed, but I find quite helpful that tests fail if I change the mock they use.

packages/search-adapter/src/empathy/__tests__/response-mappers/empathy-facet-mapper.spec.ts
```diff
@@ -107,21 +117,25 @@ function expectFacetToMatchMock(mappedFacet: HierarchicalFacet | SimpleFacet | N
...
+  if (filter.values?.length) {
+    for (let value of filter.values) {
+      counter += countHierarchicalChildren(value);
+    }
+  }
+  return counter + (filter.values?.length || 0);
+}
+
+function findFilterInChildren({ parentId, id }: HierarchicalFilter, allFilters: HierarchicalFilter[]) {
+  return parentId !== null ? allFilters.some(filter => filter.children?.some(child => child === id)) : false;
```
javieri-empathy, line 140: Same happens for this, I don't get this function, you are using it to search that a filter that is in an array is contained in that array?  Below line doesn't make any sense.

```
findFilterInChildren(mappedFacet.filters[1], mappedFacet.filters)
```

javieri-empathy (changes requested):

packages/search-adapter/src/empathy/mappers/request/params/empathy-request-filters-solr-syntax.mapper.ts
```diff
@@ -29,8 +31,8 @@ export class EmpathyRequestFiltersSolrSyntaxMapper implements Mapper<Filter[], s
   }
 
   private composeFilters(filters: Filter[]): string[] {
-    if (this.isArrayOfBooleanFilters(filters)) {
-      return filters.map(filter => filter.value);
+    if (this.isArrayOfBooleanFilters(filters) || this.isArrayOfRawFilters(filters)) {
```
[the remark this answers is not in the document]
guillei10: I can see the previous implementation considered the fact that there could be a wrong `modelName`:
Screenshot 2021-08-10 at 08 14 22

Your version doesn't include this case. Did you forget about it, or we won't be needing this anymore?

packages/search-adapter/src/empathy/__tests__/response-mappers/empathy-facet-mapper.spec.ts
guillei10: Right, I'll get rid of it.

packages/search-adapter/src/empathy/__tests__/response-mappers/empathy-facet-mapper.spec.ts
guillei10: No, I am using this to find if a filter of the array is contained in any of the children arrays of any of the filters. I think this is a valid validation xD But I'll try to simplify this.

packages/search-adapter/src/empathy/mappers/response/filters/empathy-facet-filter.mapper.ts
```diff
@@ -9,16 +9,13 @@ import { EmpathyFilter } from '../../../models';
  * @public
  */
 @injectable()
-export class EmpathyFilterMapper implements ResponseMapper<EmpathyFilter, Filter> {
-
-  map(rawFilter: EmpathyFilter, filter: Filter, context: ResponseMapperContext): Filter {
+export class EmpathyFacetFilterMapper implements ResponseMapper<EmpathyFilter, FacetFilter> {
+  map(rawFilter: EmpathyFilter, filter: FacetFilter, context: ResponseMapperContext): FacetFilter {
     const value = rawFilter.filter;
     const id = `${ filter.facetId }:${ value }`;
```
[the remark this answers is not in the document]
guillei10: I don't understand this. We at least need to map the `rawFilter.filter` property with the `id` property right?

packages/search-adapter/src/empathy/__tests__/adapter/empathy-adapter-search.spec.ts
```diff
@@ -1,4 +1,5 @@
-import { BannerSchema, FacetSchema, PromotedSchema, RedirectionSchema, ResultSchema, TaggingSchema } from '@empathyco/x-types/schemas';
+import { BannerSchema, PromotedSchema, RedirectionSchema, ResultSchema, TaggingSchema } from '@empathyco/x-types/schemas';
```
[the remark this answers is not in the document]
guillei10: I don't understand this also. I can import all of these from the same package, that would be the `x-types-next`. But in the `package.json` of the adapter I can only find the `x-types` (not the next one). Should I change the `package.json` to get the new types and import all them from there?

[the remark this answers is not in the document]
javieri-empathy: I'll try to explain myself better:

> You should not import from a package that is not defined as a dependency in the `package.json`

This is just a general rule. You made all imports from `@empathyco/x-types-next`, but there is no `@empathyco/x-types-next` entry at `packages/search-adapter/package.json` `dependencies` or `devDependencies` property.  This only works because it is searching in the parent `node_modules`, and lerna is hoisting the dependencies, but you should never do this.

> In this case you want the adapter to pick the search types of the mono-repo, there is no need to use two versions of search types here.

Instead of importing `@empathyco/x-types-next`, you just have to use the repository `x-types` package. Just like it was before. To do so just modify the `packages/search-adapter/package.json` `dependencies` entry with the following entry: `"@empathyco/x-types": "^10.0.0-alpha.7"`. Because this version matches the one defined in the `packages/search-types/package.json`, lerna will link it when bootstraping, and you can do all imports from `@empathyco/x-types` instead of `@empathyco/x-types-next`.

> Then in x-components you will have to lock the adapter version to the last released one, because it is not yet ready for this new adapter.

Because you are not going to update x-components to support the new adapter, you will have to lock its version from npm. To do so, modify `packages/x-components/package.json` `dependencies` with this entry: `"@empathyco/x-adapter": "npm:@empathyco/x-adapter@^6.2.0-alpha.8"`

packages/search-adapter/src/empathy/mappers/request/params/empathy-request-filters-solr-syntax.mapper.ts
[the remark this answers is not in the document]
javieri-empathy: Yep, I think we can do this simplification. Previously there was no hierarchy of filters, now there is, and the only distinction here should be if the filter is editable or not.

pmareke (dismissed):

javieri-empathy (changes requested):
I just realised that the `EmpathyNumberRangeFilterMapper` is overridding the `id` property of the filters. Now these filters should use the `rawFilter.filter` like the rest of boolean filters do. Remove the `id` from `EmpathyNumberRangeFilterMapper`

pmareke (approved):

javieri-empathy (approved):

PR #85, opened 2021-08-10
test(e2e): mock all the e2e requests of Next Queries

CachedaCodes:
[EX-4532](https://searchbroker.atlassian.net/browse/EX-4532)

## Motivation and Context ##

After the changes to use a mocked adapter in the Next Queries e2e test we needed to change the e2e test that had Next Queries requests to also use the mocked adapter. This time using dummies instead of stubs.

Run the e2e tests and everything should be working as previous but the dummy next queries should be showing in the changed tests

### DIFF
--- a/packages/x-components/tests/e2e/cucumber/next-queries/next-queries.spec.ts
+++ b/packages/x-components/tests/e2e/cucumber/next-queries/next-queries.spec.ts
@@ -1,5 +1,7 @@
 import { And, Given, Then, When } from 'cypress-cucumber-preprocessor/steps';
 import { InstallXOptions } from '../../../../src/x-installer/x-installer/types';
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
 import { nextQueriesStub } from './stubs/next-queries.stub';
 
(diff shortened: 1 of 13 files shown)


packages/x-components/tests/e2e/cucumber/next-queries/next-queries.spec.ts
```diff
@@ -1,5 +1,7 @@
 import { And, Given, Then, When } from 'cypress-cucumber-preprocessor/steps';
 import { InstallXOptions } from '../../../../src/x-installer/x-installer/types';
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
```
pmareke, line 4: Remove these comments please, the `@ts-ignore` is forbidden.

In order to remove the error, you need to change de `tsconfig` file inside the `tests` folder and add the rootDir to the `compilerOptions`.

pmareke (changes requested):

pmareke (changes requested):
Please add the `Background` step

pmareke (approved):

mavmaf (approved):

PR #93, opened 2021-08-13
test: mock popular searches adapter in e2e tests

mavmaf:
EX-4466

## Motivation and Context ##

We needed to have a reliable method of doing e2e tests without relying on APIs outside of our control. This PR implements a version of that solution for the Popular Searches.

- mocked-adapter changed so the getPopularSearches can be intercepted by Cypress
- e2e test for the Popular Searches changed to make use of Cypress' intercept
- Scenario 2 of Popular Searches tests was simplified, together with its .vue, in order to skip checking other components behavior (NQs, RTs, QSs and Results) since it's already being tested in the Search tests (although its refactoring to avoid using the real API is still pending).

*EX-4466*

In `cypress.json` you can change the testFiles property's first element to `"cucumber/**/*.features"` to be able to select the test to run.


packages/x-components/src/__stubs__/suggestions-stubs.factory.ts
```diff
@@ -1,43 +1,37 @@
...
  *
- * @param modelName - Model name for each suggestion.
- * @returns Array of suggestion stub.
+ * @param amount - Number of stubbed suggestions to create.
+ *
+ * @returns Array of suggestions stub.
+ *
+ * @internal
+ */
+export function getSuggestionStub(amount = 3): Suggestion[] {
```
pmareke, line 12: This change from `getSuggestionsStub` to `getSuggestionStub` breaks some tests, please take a look.

packages/x-components/tests/e2e/cucumber/popular-searches/popular-searches.spec.ts
```diff
@@ -46,7 +59,7 @@ When('popular search number {int} is clicked', (popularSearchItem: number) => {
 });
 
 And(
-  'the clicked popular search is removed from Popular Searches if {boolean} is true',
+  'the clicked popular search is removed from Popular Searches if hideSessionQueries = {boolean} is true',
```
pmareke, line 62: Please fix this error

image

pmareke (changes requested):
Fix the build and all the errors please.

davidmfempathy (approved):

packages/x-components/src/__stubs__/suggestions-stubs.factory.ts
```diff
@@ -41,3 +41,24 @@ export function getSuggestionsStub(modelName: ModelNameType): Suggestion[] {
...
+ * @returns A suggestion.
+ */
+export function createSuggestionStub(query: string, suggestion?: Partial<Suggestion>): Suggestion {
+  return {
+    facets: [],
+    key: 'key',
+    query,
+    totalResults: 10,
+    results: [],
+    modelName: 'PopularSearch',
```
pmareke, line 61: If we're creating suggestions why are we using always the `PopularSearch` modelName?

pmareke (changes requested):

pmareke (approved):

packages/x-components/tests/e2e/cucumber/empathize.feature
```diff
@@ -2,6 +2,7 @@ Feature: Full Empathize
 
   Background:
     Given a next queries API
+    Given a suggestions API
```
pmareke, line 5: I'm not sure why are we adding this `Given`

packages/x-components/tests/e2e/cucumber/history-queries.feature
```diff
@@ -2,6 +2,7 @@ Feature: History queries component
 
   Background:
     Given a next queries API
+    Given a suggestions API
```
pmareke, line 5: I'm not sure why are we adding this Given

packages/x-components/tests/e2e/cucumber/search-box.feature
```diff
@@ -2,6 +2,7 @@ Feature: Search-box component
 
   Background:
     Given a next queries API
+    Given a suggestions API
```
pmareke, line 5: I'm not sure why are we adding this Given

pmareke (changes requested):

pmareke (approved):

davidmfempathy (approved):

PR #95, opened 2021-08-17
Add new extra params component

pmareke:
EX-4575


javieri-empathy:
Be careful when merging, it is not EX-4545, it is EX-4575

packages/x-components/src/components/__tests__/base-extra-params.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+import Vue, { VueConstructor } from 'vue';
+import Component from 'vue-class-component';
+import { ComponentOptions } from 'vue/types/umd';
+import { Dictionary } from '../../utils';
+import { ExtraParamXEvent } from '../../wiring';
+import { XProvide } from '../decorators/injection.decorators';
+import BaseExtraParams from '../base-extra-params.vue';
+
+describe('testing extra params component', () => {
+  const emitSpy = jest.fn();
```
javieri-empathy, line 11: Using globals is unnecessary. It is better and easier to create a new jest function mock for each test, and return it as part of the test API. I would just name it `emit` in that case. Then you can avoid the `beforeEach` loop.

packages/x-components/src/components/__tests__/base-extra-params.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+
+  @Component({
+    template: `<div><slot/></div>`
+  })
+  class Provider extends Vue {
```
javieri-empathy, line 16: I would nest this component under a new describe section.

packages/x-components/src/components/__tests__/base-extra-params.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+      },
+      events: ['UserChangedRecommendationsRequestParam']
+    });
+
+    expect(emitSpy).toHaveBeenCalledWith('UserChangedRecommendationsRequestParam', {
+      warehouse: 'warehouse'
+    });
+    expect(emitSpy).toHaveBeenCalledTimes(1);
+  });
+});
```
javieri-empathy, line 143: I'm missing some tests changing the value and the provided value, and asserting the emitted events.

packages/x-components/src/components/base-extra-params.vue
```diff
@@ -0,0 +1,165 @@
...
+     *
+     * @param newExtraParams - New extra params.
+     * @param oldExtraParams - Previous extra params.
+     *
+     * @internal
+     */
+    @Watch('extraParams', { immediate: true, deep: true })
+    protected syncExtraParams(
+      newExtraParams: Record<string, any>,
+      oldExtraParams: Record<string, any>
```
javieri-empathy, line 88: As I told you yesterday, this is not a good idea. When watching objects deeply, the watcher might be triggered because the object instance has changed, or because a single property has changed. In this last case, you will [receive in both new and old parameters](https://vuejs.org/v2/api/#vm-watch) the same object with the same values.

packages/x-components/src/components/base-extra-params.vue
```diff
@@ -0,0 +1,165 @@
...
+   *
+   * Each extra param is emitted by each event.
+   *
+   * By default if there is an extra param passed as prop, it will override
+   * the snippet config in case it exists.
+   *
+   * @public
+   */
+  @Component
+  export default class BaseExtraParams extends Vue {
```
javieri-empathy, line 21: I'm going to put here what we talked yesterday:

* Create a new x-module (sorry for that 😅). We can do this in another PR.
* 3 events: `ExtraRequestParamsProvided`, and `UserChangedRequestParams` (this one you won't use it yet) emitted from the components, and `ExtraRequestParamsChanged`, emitted from a watcher in the store.
	* Modules that need the extra params are subscribed to `ExtraRequestParamsChanged` using the wiring.
	* `ExtraRequestParamProvided` and `UserChangedRequestParams` is only used in the extra params module.
* 2 components:
	* `<ExtraRequestParams :values="myExtraRequestParams" />`
	  * Where `myExtraRequestParams` is a dictionary
		* Uses `@XEmit` with `immediate` and `deep`
		* Does not have to perform diffs because `value` prop is trust to be different.
	* `<SnippetConfigExtraRequestParams />`
		* Renders an `ExtraRequestParams` component.
		* Keeps manually an object with the desired snippet config properties to avoid unnecessary re-renders:
```
    public extraRequestParams: Dictionary<unknown> = { };

    @Watch('snippetConfig', { deep: true, immediate: true })
    syncExtraRequestParams({
      lang,
      searchLang,
      instance,
      scope,
     // Every other param
      ...extraRequestParams
    }: SnippetConfig): void {
      forEach(extraRequestParams, (name, value) => {
        this.$set(this.extraRequestParams, name, value);
      });
    }
```

packages/x-components/src/components/base-extra-params.vue
```diff
@@ -0,0 +1,165 @@
...
+     *
+     * @remarks In the merge strategy the props have more priority than the snippet.
+     *
+     * @returns An object with the extra params.
+     *
+     * @internal
+     */
+    protected get extraParams(): Dictionary {
+      const {
+        _instance,
```
javieri-empathy, line 60: With these `_` prefixing the variables you are not removing the snippet config known values, `extraParams` still contains the `lang`, `env`... properties.

javieri-empathy (changes requested):

packages/x-components/src/components/base-extra-params.vue
javieri-empathy: Note that this is the initial idea, but when doing it we might realise that things make more sense in other way.

packages/x-components/src/x-modules/extra-params/components/__tests__/extra-params.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+    };
+  }
+
+  it('is an XComponent which has an XModule', () => {
+    const { wrapper } = renderExtraParams({ warehouse: 1234 });
+    expect(isXComponent(wrapper.vm)).toEqual(true);
+    expect(getXComponentXModuleName(wrapper.vm)).toEqual('extraParams');
+  });
+
+  it('emits the ExtraRequestParamsProvided event when the values changed', async () => {
```
davidmfempathy, line 35: changed > change?

packages/x-components/src/x-modules/extra-params/components/__tests__/extra-params.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+    const { wrapper } = renderExtraParams({ warehouse: 1234 });
+    const extraRequestParamsProvidedCallback = jest.fn();
+
+    wrapper.vm.$x
+      .on('ExtraRequestParamsProvided', true)
+      .subscribe(extraRequestParamsProvidedCallback);
+
+    expect(extraRequestParamsProvidedCallback).toHaveBeenCalledWith<
+      [WirePayload<Dictionary<unknown>>]
```
davidmfempathy, line 44: Remove this as we did in the rest or add it to the rest. I'd remove them as it's done in the rest of the app.

packages/x-components/src/x-modules/extra-params/components/extra-params.vue
```diff
@@ -0,0 +1,62 @@
+<script lang="ts">
+  import Vue from 'vue';
+  import { Component, Prop } from 'vue-property-decorator';
+  import { xComponentMixin, XEmit } from '../../../components';
+  import { Dictionary } from '../../../utils';
+  import { extraParamsXModule } from '../x-module';
+
+  /**
+   * It emits a {@link ExtraParamsXEvents.ExtraRequestParamsProvided}
```
davidmfempathy, line 9: Multiline, You might include some words of the next line in this.

packages/x-components/src/x-modules/extra-params/components/extra-params.vue
```diff
@@ -0,0 +1,62 @@
...
+
+[1](./../../api/x-components.extraparamsxevents.extrarequestparamsprovided.md)
+
+## See it in action
+
+_See how the event is triggered when the component is rendered._
+
+```vue
+<template>
+  <ExtraParams :values="{ warehouse: 123456789 }" />
```
javieri-empathy, line 48: Don't hardcode objects in the templates, use data variables

packages/x-components/src/x-modules/extra-params/components/extra-params.vue
```diff
@@ -0,0 +1,62 @@
...
+</template>
+
+<script>
+  import { ExtraParams } from '@empathyco/x-components/extra-params.vue';
```
javieri-empathy, line 52: Import from the module `@empathyco/x-components/extra-params`

packages/x-components/src/views/FullApp.vue
```diff
@@ -347,6 +347,7 @@
   import { Result, Sort } from '@empathyco/x-types';
   import { getBannersStub } from '../__stubs__/banners-stubs.factory';
   import { getPromotedsStub } from '../__stubs__/promoteds-stubs.factory';
+  import { XProvide } from '../components';
```
javieri-empathy, line 350: Rollback this file

packages/x-components/src/x-modules/extra-params/components/extra-params.vue
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   * It emits a {@link ExtraParamsXEvents.ExtraRequestParamsProvided}
+   * with the values received as a prop.
+   *
+   * @public
+   */
+  @Component({
+    mixins: [xComponentMixin(extraParamsXModule)]
+  })
+  export default class ExtraParams extends Vue {
```
javieri-empathy, line 17: Export the component from the module barrel

javieri-empathy (changes requested):

davidmfempathy (changes requested):

javieri-empathy (approved):

davidmfempathy (approved):

PR #96, opened 2021-08-17
feat: modify search function in XAPI to open X

tiborux:
This PR modifies the current XAPI to use the search function to open X when it's called without query.

To make this event works with the current X components, you have to use de BaseEventsModal instead of the BaseIdModal, this modal will be opened when an event that it's been listened to by it, its called.

Some of the events were renamed to be more accurate.

To test this, you can go to the BaseEventsModal view and change the events to open the modal:

`<BaseEventsModal :eventsToOpenModal=['UserClickedOpenEventsModal, 'UserOpenXProgramatically']>`

EX-3555

### DIFF
--- a/packages/x-components/src/components/decorators/__tests__/bus.decorators.spec.ts
+++ b/packages/x-components/src/components/decorators/__tests__/bus.decorators.spec.ts
@@ -42,22 +42,22 @@ describe('testing @XOn decorator', () => {
       dataListener(this, payload);
     }
 
-    @XOn(['UserClickedOpenX', 'UserClickedCloseX'])
+    @XOn(['UserClickedOpenEventsModal', 'UserClickedCloseEventsModal'])
     testingXOnMultiple(): void {
       multipleListener(this);
     }
 
-    @XOn('UserClickedCloseX', { moduleName: 'searchBox', origin: 'default' })
+    @XOn('UserClickedCloseEventsModal', { moduleName: 'searchBox', origin: 'default' })
     testingXOnMultipleOptionsFiltered(): void {
       filteredWithMultipleOptionsListener(this);
     }
 
-    @XOn('UserClickedOpenX', { moduleName: 'searchBox' })
+    @XOn('UserClickedOpenEventsModal', { moduleName: 'searchBox' })
     testingXOnOptions(): void {
       optionsListener(this);
     }
 
-    @XOn('UserClickedCloseX', { moduleName: 'empathize' })
+    @XOn('UserClickedCloseEventsModal', { moduleName: 'empathize' })
     testingXOnOptionsFiltered(): void {
       filteredOptionsListener(this);
     }
@@ -106,8 +106,8 @@ describe('testing @XOn decorator', () => {
   });
 
   it('subscribes to a defined array of events', () => {
-    component.vm.$x.emit('UserClickedOpenX');
-    component.vm.$x.emit('UserClickedCloseX');
+    component.vm.$x.emit('UserClickedOpenEventsModal');
+    component.vm.$x.emit('UserClickedCloseEventsModal');
 
     expect(multipleListener).toHaveBeenCalledTimes(2);
   });
@@ -136,8 +136,8 @@ describe('testing @XOn decorator', () => {
     component.vm.$x.emit('UserAcceptedAQuery', 'que pasara que misterios habra');
     component.vm.$x.emit('UserIsTypingAQuery', 'estare escribiendo?');
     component.vm.$x.emit('UserTalked', 'no he dicho nada');
-    component.vm.$x.emit('UserClickedOpenX');
-    component.vm.$x.emit('UserClickedCloseX');
+    component.vm.$x.emit('UserClickedOpenEventsModal');
+    component.vm.$x.emit('UserClickedCloseEventsModal');
 
     expect(singleListener).not.toHaveBeenCalled();
     expect(multipleListener).not.toHaveBeenCalled();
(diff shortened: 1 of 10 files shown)


packages/x-components/src/components/decorators/__tests__/bus.decorators.spec.ts
```diff
@@ -42,22 +42,22 @@ describe('testing @XOn decorator', () => {
       dataListener(this, payload);
```
davidmfempathy, line 42: I'm not sure if the rename of `UserClickedOpenX`, and  `UserClickedCloseX` with those generic events is correct.
I think they're different events but tell me if I'm wrong.

davidmfempathy: In this test makes sense to have the specific events.

packages/x-components/src/components/modals/__tests__/base-events-close-button.spec.ts
```diff
@@ -37,10 +37,10 @@ function renderBaseEventsModalClose({
 }
```
davidmfempathy, line 37: In this test makes sense to have the generic events (events for the modal).

packages/x-components/src/components/modals/base-events-modal-close.vue
```diff
@@ -18,16 +18,17 @@
   import BaseEventButton from '../base-event-button.vue';
 
   /**
-   * Component containing an event button that emits {@link XEventsTypes.UserClickedCloseX} when
-   * clicked. It has a default slot to customize its contents.
+   * Component containing an event button that emits
```
davidmfempathy, line 21: just `contains`?

packages/x-components/src/components/modals/base-events-modal-open.vue
```diff
@@ -18,16 +18,16 @@
   import BaseEventButton from '../base-event-button.vue';
 
   /**
-   * Component containing an event button that emits {@link XEventsTypes.UserClickedOpenX} when
-   * clicked. It has a default slot to customize its contents.
+   * Component containing an event button that emits {@link XEventsTypes.UserClickedOpenEventsModal}
```
davidmfempathy, line 21: just `contains`?

packages/x-components/src/wiring/events.types.ts
```diff
@@ -143,6 +143,11 @@ export interface XEventsTypes
    * * Payload: The scroll id which has scrolled to top.
    */
   UserClickedScrollToTop: string;
+  /**
+   * The user has open X with the API.
```
davidmfempathy, line 147: In order to have the same time as the rest of descriptions it should be: `The user opened X programmatically`. Notice that I also suggest to replace with the API with programmatically as the developer user can configure it to be emitted from any part of the app.

packages/x-components/src/x-installer/api/base-api.ts
```diff
@@ -56,9 +56,12 @@ export class BaseXAPI implements XAPI {
    *
    * @public
    */
-  search(query: string): void {
+  search(query?: string): void {
     // TODO - It should do more things like emit the query was changed out of the normal user flow
```
davidmfempathy, line 60: Should we remove this line?

davidmfempathy (changes requested):

packages/x-components/src/components/decorators/__tests__/bus.decorators.spec.ts
```diff
@@ -106,8 +106,8 @@ describe('testing @XOn decorator', () => {
   });
 
   it('subscribes to a defined array of events', () => {
-    component.vm.$x.emit('UserClickedOpenX');
-    component.vm.$x.emit('UserClickedCloseX');
+    component.vm.$x.emit('UserClickedOpenEventsModal');
```
pmareke, line 109: I'm not sure why do we need to change this events, I mean you're changing the events of the close and open, but here you're testing the decorator.

I'll leave this file as it was before.

packages/x-components/src/components/modals/base-events-modal-close.vue
pmareke: I like `Component containing...`

packages/x-components/src/components/modals/base-events-modal.vue
```diff
@@ -194,7 +196,7 @@ see a full example on how this would work with custom events.
 
 A list of events that the component will emit:
 
-- `UserClickedCloseX`: the event is emitted after clicking outside the content rendered in the
-  default slot.
+- `UserClickedCloseEventsModal`: the event is emitted after clicking outside the content rendered in
```
pmareke, line 199: Remember to add the new event `UserClickedOutOfEventsModal `.

packages/x-components/src/x-installer/api/base-api.ts
pmareke: I'll remove it too.

pmareke (changes requested):

packages/x-components/src/components/decorators/__tests__/bus.decorators.spec.ts
tiborux: The thing here is that I was taking a look with Ivan and we show that these events are related to opening a modal, not X, I mean, if we use the modals for other purposes, we are launching the UserClickedOpenX event when you open an Events Modal, and it doesn't make sense. 

So we think that we have to rename the events to have a standard name that refers to close the modal itself instead of closing X components.

packages/x-components/src/components/decorators/__tests__/bus.decorators.spec.ts
tiborux: Because the other events do not longer exist.

packages/x-components/src/components/modals/base-events-modal-close.vue
tiborux: With the david's suggestion we can change the text and format it in just one line

pmareke (approved):

davidmfempathy (approved):

PR #105, opened 2021-08-20
Feature/ex 4592 move old facets components

javieri-empathy:
# Description

This PR moves 3 facets components from the old to the new module. It tries to touch as minimum code as possible to make them work.

> feat(facets-next): add `AllFilter`, `BaseFilter`, `EditableNumberRangeFilter` components to the `facets-next` x-module.

### DIFF
--- a/packages/x-components/src/x-modules/facets-next/components/filters/__tests__/editable-number-range-filter.spec.ts
+++ b/packages/x-components/src/x-modules/facets-next/components/filters/__tests__/editable-number-range-filter.spec.ts
@@ -9,7 +9,6 @@ import {
   getXComponentXModuleName,
   isXComponent
 } from '../../../../../components/x-component.utils';
-import { XPlugin } from '../../../../../plugins/x-plugin';
 import { RootXStoreState } from '../../../../../store/store.types';
 import { DeepPartial } from '../../../../../utils/types';
 import { facetsNextXModule } from '../../../x-module';
@@ -41,10 +40,7 @@ function renderEditableNumberRangeFilter({
   const localVue = createLocalVue();
   localVue.use(Vuex);
   const store = new Store<DeepPartial<RootXStoreState>>({});
-  installNewXPlugin({ store }, localVue);
-
-  XPlugin.resetInstance();
-  XPlugin.registerXModule(facetsNextXModule);
+  installNewXPlugin({ store, initialXModules: [facetsNextXModule] }, localVue);
 
   resetXFacetsStateWith(store, {});
   const wrapper = mount<EditableNumberRangeFilterComponent>(
@@ -111,8 +107,8 @@ describe('testing BaseNumberRangeFilter component', () => {
     expect(
       (filterWrapper.find(getDataTestSelector('range-max')).element as HTMLInputElement).value
     ).toBe('5');
-    expect(applyButtonWrapper.text()).toBe('✅');
-    expect(clearButtonWrapper.text()).toBe('🗑');
+    expect(applyButtonWrapper.text()).toBe('✓');
+    expect(clearButtonWrapper.text()).toBe('𐄂');
   });
 
   // eslint-disable-next-line max-len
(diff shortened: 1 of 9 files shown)


packages/x-components/src/x-modules/facets/components/filters/__tests__/editable-number-range-filter.spec.ts
```diff
@@ -15,11 +15,13 @@ import { XPlugin } from '../../../../../plugins/x-plugin';
 import { facetsXModule } from '../../../x-module';
 import { resetXFacetsStateWith } from '../../__tests__/utils';
 
-Object.defineProperty(HTMLInputElement.prototype, 'valueAsNumber', {
-  get() {
-    return parseFloat(this.value);
-  }
-});
+if (!HTMLInputElement.prototype.valueAsNumber) {
```
tajespasarela, line 18: WTF!?!?

packages/x-components/src/x-modules/facets-next/components/filters/editable-number-range-filter.vue
```diff
@@ -0,0 +1,463 @@
...
+        v-if="!isInstant"
+        @click="emitUserModifiedFilter"
+        class="x-button x-editable-number-range-filter__apply"
+        :disabled="hasError"
+        data-test="range-apply"
+      >
+        <!--
+            @slot Slot used to customize the apply button content.
+        -->
+        <slot name="apply-content">✅</slot>
```
tajespasarela, line 55: please change the emoji by an UNICODE symbol: https://unicode-table.com/es/2713/

packages/x-components/src/x-modules/facets-next/components/filters/editable-number-range-filter.vue
```diff
@@ -0,0 +1,463 @@
...
+      <button
+        v-if="renderClearButton"
+        @click="clearValues"
+        class="x-button x-editable-number-range-filter__clear"
+        data-test="range-clear"
+      >
+        <!--
+            @slot Slot used to customize the clear button content.
+        -->
+        <slot name="clear-content">🗑</slot>
```
tajespasarela, line 67: please change the emoji by an UNICODE symbol: https://unicode-table.com/es/10102/

packages/x-components/src/x-modules/facets-next/components/filters/__tests__/editable-number-range-filter.spec.ts
```diff
@@ -0,0 +1,343 @@
...
+  hasClearButton = true
+}: EditableNumberRangeFilterOptions = {}): EditableNumberRangeFilterAPI {
+  Vue.observable(filter);
+  const localVue = createLocalVue();
+  localVue.use(Vuex);
+  const store = new Store<DeepPartial<RootXStoreState>>({});
+  installNewXPlugin({ store }, localVue);
+
+  XPlugin.resetInstance();
+  XPlugin.registerXModule(facetsNextXModule);
```
tajespasarela, line 47: maybe move the clear getters cache to the `installNewXPlugin` util?

tajespasarela (changes requested):

packages/x-components/src/x-modules/facets/components/filters/__tests__/editable-number-range-filter.spec.ts
tajespasarela: Why was it working before?

packages/x-components/src/x-modules/facets-next/components/filters/__tests__/editable-number-range-filter.spec.ts
javieri-empathy: You still need to reset the `instance`. What I did was moving the `XPlugin.resetInstance` call to the `installNewXPlugin` utility. Seems tests keep passing, So we can start removing those calls from the `spec` files.

packages/x-components/src/x-modules/facets/components/filters/__tests__/editable-number-range-filter.spec.ts
javieri-empathy: It was working because the polyfill was only defined once, and had `configurable: false`, but as no test was trying to override it, there was no problem at all. 

The part that I still do not understand is that the JSDOM prototypes are shared between tests somehow, and it only happens in the CI. In our laptops, both @davidmfempathy and me tried running the tests with different versions of node, and they always worked, but in GitHub, it looked like they were sharing the `valueAsNumber` property polyfill, so when the second test was launched this property was already defined, and because of the `configurable: false` default value, it failed defining the polyfill.

davidmfempathy (approved):

tajespasarela (approved):

PR #110, opened 2021-08-24
Reset search state when extra params change

pmareke:
EX-4554

### DIFF
--- a/packages/x-components/src/x-modules/search/wiring.ts
+++ b/packages/x-components/src/x-modules/search/wiring.ts
@@ -85,7 +85,7 @@ export const setSort = wireCommit('setSort');
 export const setPage = wireCommit('setPage');
 
 /**
- * Sets the search state `page`.
+ * Sets the search state `params`.
  *
  * @public
  */
--- a/packages/x-components/src/x-modules/search/wiring.ts
+++ b/packages/x-components/src/x-modules/search/wiring.ts
@@ -85,6 +85,13 @@ export const setSort = wireCommit('setSort');
 export const setPage = wireCommit('setPage');
 
 /**
+ * Sets the search state `page`.
+ *
+ * @public
+ */
+export const setExtraParams = wireCommit('setExtraParams');
+
+/**
  * Increases the current search state `page` by one.
  *
  * @public
@@ -155,5 +162,10 @@ export const searchWiring = createWiring({
   },
   UserReachedResultsListEnd: {
     increasePageAppendingResults
+  },
+  ExtraParamsChanged: {
+    setExtraParams,
+    resetPage,
+    resetSort
   }
 });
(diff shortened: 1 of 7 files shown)


pmareke:
```
 <!-- Extra params components -->
    <RenderlessExtraParam #default="{ value, updateValue }" name="warehouse" defaultValue="1234">
      <BaseDropdown @change="updateValue" :value="value" :items="items">
        <template #item="{ item, isHighlighted, isSelected }">
          <CheckIcon v-if="isSelected" />
          <ChevronLeftIcon v-if="isHighlighted" />
          Warehouse: {{ item }}
        </template>
      </BaseDropdown>
    </RenderlessExtraParam>
    <RenderlessExtraParam #default="{ value, updateValue }" name="catalog" defaultValue="en">
      <BaseDropdown @change="updateValue" :value="value" :items="['en', 'es']">
        <template #item="{ item, isHighlighted, isSelected }">
          <CheckIcon v-if="isSelected" />
          <ChevronLeftIcon v-if="isHighlighted" />
          Catalog: {{ item }}
        </template>
      </BaseDropdown>
    </RenderlessExtraParam>
```

packages/x-components/src/x-modules/search/wiring.ts
```diff
@@ -162,5 +176,11 @@ export const searchWiring = createWiring({
   },
   UserReachedResultsListEnd: {
     increasePageAppendingResults
+  },
+  ExtraParamsChanged: {
```
davidmfempathy, line 180: Do we have to reset the rest of entities within the module? results, partialResults, relatedTags, banners, promoteds, selectedFilters, spellcheckedQuery ?

davidmfempathy (changes requested):

davidmfempathy (dismissed):

packages/x-components/src/x-modules/search/wiring.ts
pmareke: I think as the request is send again this entities will be overridden right?

packages/x-components/src/x-modules/search/wiring.ts
```diff
@@ -84,6 +84,13 @@ export const setSort = wireCommit('setSort');
  */
 export const setPage = wireCommit('setPage');
 
+/**
+ * Sets the search state `page`.
```
CachedaCodes, line 88: Correct this doc.

```suggestion
 * Sets the search state `params`.
```

- [ ] Correct doc

CachedaCodes (changes requested):

packages/x-components/src/x-modules/search/wiring.ts
davidmfempathy: As we've talked through Slack, you're right.

guillei10 (dismissed):

davidmfempathy (approved):

guillei10 (approved):

PR #112, opened 2021-08-24
feat(facets-next): add all remaining components to Facets Next X-Module

tajespasarela:
Add `ClearFilters`, `ExcludeFiltersWithNoResults`, `SortedFilters`, `FiltersSearch`,
`SlicedFilters`, `SelectedFilters` and  `SelectedFiltersList` components.

EX-4615

# Pull Request Template #

Add all pending components to `NextFacets` X-Module. (Except `HierarchicalFIlter`, which has its own PR).
Also used these new components in the `LayoutNext` view to check them working.

## Motivation and Context ##

The motivation is to move all pending things from the old `Facets` X-Module to the new `FacetsNext`X-Module, before removing the old and replace it with the new.

## How to test ##
Use the `LayoutNext` view to check it working.

**REMARK:** All the facets are multiselect by default. The `MultiSelectFilters` component disappears. Now it is a modifier in the Entities which modifies the facet to be "Single Select":

```
      FilterEntityFactory.instance.registerFilterModifier('brand_facet', [SingleSelectModifier]);
```

As in Juguettos the facets are not configured as multiselected in back, and we don't do the "OR" transform in the adapter any more, some filters will return no results if you select more than one.

We will keep this way at the moment until we mock all the tests and move to the Platform Backend in the X-Components project.

### DIFF
--- a/packages/x-components/src/views/LayoutNext.vue
+++ b/packages/x-components/src/views/LayoutNext.vue
@@ -10,17 +10,21 @@
             "
           >
             <div class="x-input-group x-input-group--card x-list__item--expand">
-              <SearchInput placeholder="Search" aria-label="Search for products" />
+              <SearchInput aria-label="Search for products" placeholder="Search" />
               <ClearSearchInput aria-label="Clear query">Clear</ClearSearchInput>
-              <SearchButton class="x-input-group__action" aria-label="Search">
+              <SearchButton aria-label="Search" class="x-input-group__action">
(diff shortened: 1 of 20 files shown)


pmareke (approved):
The view is working without errors, the tests are passing and the new components have the same content as the old ones adapting them to the new module.

davidmfempathy (approved):

guillei10 (approved):

PR #115, opened 2021-08-25
feat(next-queries): nq extra params & state reset when they change

CachedaCodes:
[EX-4624](https://searchbroker.atlassian.net/browse/EX-4624)

Adds extra params to the Next Queries request


packages/x-components/src/x-modules/next-queries/store/__tests__/getters.spec.ts
```diff
@@ -52,6 +60,9 @@ describe('testing next queries module getters', () => {
         config: {
           maxItemsToRequest: 5,
           hideSessionQueries: true
+        },
```
pmareke, line 63: This params is not necessary, please remove it.

pmareke (changes requested):

pmareke (approved):

tiborux (approved):

PR #116, opened 2021-08-25
test: mock search-box e2e tests

mavmaf:
## Motivation and Context ##
We needed to have a reliable method of doing e2e tests without relying on APIs outside of our control. This PR implements a version of that solution for the search.box tests.

- mocked-adapter changed so the search can be intercepted by Cypress
- e2e test for the search-box changed to make use of Cypress' intercept

EX-4474

In cypress.json you can change the testFiles property's first element to "cucumber/**/*.features" to be able to select the test to run.


pmareke (approved):
Great job!

PR #117, opened 2021-08-25
feat(facets-next): move HierarchicalFilter component to new `facets-next` x-module.

tajespasarela:
# Pull Request Template #

feat(facets-next): move HierarchicalFilter component to new `facets-next` x-module.

EX-4622

## Motivation and Context ##

Move the `HierarchicalFilter` to the `FacetsNext` module

Since the new `FacetsNext` modules uses the new `x-types` to make this HierarchicalFilter work, we should use the new version of the `x-adapter` that uses that new `x-types`.
We had to add both versions of the Adapter (the old and the new) at the same time. This will be a lot of work and a mess with the dependencies.

Therefore I will do it in the next task, which removes the old Facets module and is replaced by the new one. Then we can remove all the duplicities in the dependencies.


packages/x-components/src/x-modules/facets-next/components/filters/__tests__/hierarchical-filter.spec.ts
```diff
@@ -0,0 +1,489 @@
...
+
+function renderHierarchicalFilter({
+  template = `<HierarchicalFilter :filter="filter" />`,
+  filter = createNextHierarchicalFilter('category', 'clothing'),
+  filters = [filter]
+}: HierarchicalFilterOptions = {}): HierarchicalFilterAPI {
+  const [, localVue] = installNewXPlugin({ initialXModules: [facetsNextXModule] });
+  const emit = jest.spyOn(XPlugin.bus, 'emit');
+  const store = XPlugin.store;
+  resetXFacetsStateWith(store, { filters: arrayToObject(filters, 'id') });
```
davidmfempathy, line 23: This is failing in the tests.

davidmfempathy (changes requested):

packages/x-components/src/x-modules/facets-next/components/filters/__tests__/hierarchical-filter.spec.ts
pmareke: Yes, the `FacetNextState` needs more properties like `query`, `groups` or `facets`

```
export interface FacetsNextState {
  filters: Record<Filter['id'], Filter>;
  groups: Record<Facet['id'], GroupId>;
  query: string;
  facets: Record<Facet['id'], Omit<Facet, 'filters'>>;
}
```

pmareke (changes requested):
Nice job!

PR #127, opened 2021-08-30
feat(mixin): add origin injection in the mixin and the bus

davidmfempathy:
# Description

It adds the origin injection in the mixin and also in the generated bus API.

NOTE: Please, remember to add the task reference EX-4636. I forgot it in the commit message.

### DIFF
--- a/packages/x-components/src/plugins/__tests__/x-plugin-mixin.spec.ts
+++ b/packages/x-components/src/plugins/__tests__/x-plugin-mixin.spec.ts
@@ -35,7 +35,7 @@ describe('testing $x component API global mixin', () => {
       expect(listener).toHaveBeenCalledTimes(1);
       expect(listener).toHaveBeenCalledWith({
         eventPayload: 'So awesome, much quality, such skill',
-        metadata: { moduleName: null, origin: 'default' }
+        metadata: { moduleName: null }
       });
     });
 
--- a/packages/x-components/src/plugins/__tests__/x-plugin-mixin.spec.ts
+++ b/packages/x-components/src/plugins/__tests__/x-plugin-mixin.spec.ts
@@ -1,5 +1,6 @@
 import { mount, shallowMount, Wrapper } from '@vue/test-utils';
-import { ComponentOptions, default as Vue } from 'vue';
+import { ComponentOptions, CreateElement, VNode, default as Vue } from 'vue';
+import { Component } from 'vue-property-decorator';
 import { installNewXPlugin } from '../../__tests__/utils';
 import { xComponentMixin } from '../../components/x-component.mixin';
 import { searchBoxXModule } from '../../x-modules/search-box/x-module';
@@ -22,6 +23,66 @@ describe('testing $x component API global mixin', () => {
   afterEach(() => {
     componentInstance.destroy();
     jest.clearAllMocks();
+  });
+
+  describe('testing origin', () => {
+    it('includes "default" origin in the $x.emit metadata', () => {
+      const listener = jest.fn();
+
+      componentInstance.vm.$x.on('UserIsTypingAQuery', true).subscribe(listener);
+      componentInstance.vm.$x.emit('UserIsTypingAQuery', 'So awesome, much quality, such skill');
+
+      expect(listener).toHaveBeenCalledTimes(1);
+      expect(listener).toHaveBeenCalledWith({
+        eventPayload: 'So awesome, much quality, such skill',
+        metadata: { moduleName: null, origin: 'default' }
+      });
+    });
+
+    it('overrides the origin in the $x.emit metadata', () => {
+      const listener = jest.fn();
+
+      @Component({
+        template: `<div><slot/></div>`,
+        provide: {
+          origin: 'origin-test'
+        }
+      })
+      class Provider extends Vue {}
+
+      @Component({
+        mixins: [xComponentMixin(searchBoxXModule)]
+      })
+      class Injecter extends Vue {
+        render(createElement: CreateElement): VNode {
+          return createElement();
+        }
+      }
+
+      const wrapper = mount(
+        {
+          template: `<Provider><Injecter /></Provider>`,
+          components: {
+            Provider,
+            Injecter
+          }
+        } as ComponentOptions<any> & ThisType<Vue>,
+        {
+          localVue
+        }
+      );
+
+      wrapper
+        .findComponent(Injecter)
+        .vm.$x.emit('UserIsTypingAQuery', 'So awesome, much quality, such skill');
+      componentInstance.vm.$x.on('UserIsTypingAQuery', true).subscribe(listener);
+
+      expect(listener).toHaveBeenCalledTimes(1);
+      expect(listener).toHaveBeenCalledWith({
+        eventPayload: 'So awesome, much quality, such skill',
+        metadata: { moduleName: 'searchBox', origin: 'origin-test' }
+      });
+    });
   });
 
   it('allows emitting and subscribing to events via $x object', () => {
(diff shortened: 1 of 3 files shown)


packages/x-components/src/plugins/x-plugin.mixin.ts
```diff
@@ -16,18 +17,27 @@ declare module 'vue/types/vue' {
...
+    origin: {
+      default: undefined
+    }
+  },
   created(): void {
     this.xComponent = getRootXComponent(this);
 
     const aliasAPI = getAliasAPI(this.$store);
-    const busAPI = getBusAPI(bus, this.xComponent);
+    const busAPI = getBusAPI(bus, this.xComponent, this.origin);
```
tajespasarela, line 41: remove blank line

packages/x-components/src/plugins/x-plugin.mixin.ts
```diff
@@ -16,18 +17,27 @@ declare module 'vue/types/vue' {
...
-): ComponentOptions<Vue> & ThisType<Vue & { xComponent: XComponent | undefined }> => ({
+): ComponentOptions<Vue> &
+  ThisType<Vue & { xComponent: XComponent | undefined; origin: QueryOrigin }> => ({
+  inject: {
+    origin: {
+      default: undefined
+    }
+  },
   created(): void {
     this.xComponent = getRootXComponent(this);
```
tajespasarela, line 38: remove blank line

packages/x-components/src/plugins/__tests__/x-plugin-mixin.spec.ts
```diff
@@ -24,6 +25,66 @@ describe('testing $x component API global mixin', () => {
     jest.clearAllMocks();
   });
 
+  describe('testing origin', () => {
+    it('includes "default" origin in the $x.emit metadata', () => {
```
tajespasarela, line 29: here is not emitting any origin, so change the description

packages/x-components/src/plugins/__tests__/x-plugin-mixin.spec.ts
```diff
@@ -24,6 +25,66 @@ describe('testing $x component API global mixin', () => {
...
+      @Component({
+        template: `<div><slot/></div>`,
+        provide: {
+          origin: 'origin-test'
+        }
+      })
+      class Provider extends Vue {}
+
+      @Component({
+        mixins: [xComponentMixin(searchBoxXModule)]
```
tajespasarela, line 54: It is not necessary to create an XComponent to test this.
Maybe rename the component to `Injector`
And maybe with a simple template, the component remains less verbose:

```
      @Component({ template: '<div>{{ $origin }}</div>' })
      class Injector extends Vue {}
```

packages/x-components/src/plugins/__tests__/x-plugin-mixin.spec.ts
```diff
@@ -24,6 +25,66 @@ describe('testing $x component API global mixin', () => {
...
+        }
+      );
+
+      wrapper
+        .findComponent(Injecter)
+        .vm.$x.emit('UserIsTypingAQuery', 'So awesome, much quality, such skill');
+      componentInstance.vm.$x.on('UserIsTypingAQuery', true).subscribe(listener);
+
+      expect(listener).toHaveBeenCalledTimes(1);
+      expect(listener).toHaveBeenCalledWith({
```
tajespasarela, line 81: instead of checking the whole event param maybe just check the origin:

```
      const { metadata } = listener.mock.calls[0][0];
      expect(metadata.origin).toBe('origin-test');
```

This way if metadata changes (that will always change), this test doesn't break

tajespasarela (changes requested):

packages/x-components/src/plugins/x-plugin.mixin.ts
```diff
@@ -37,20 +47,27 @@ export const createXComponentAPIMixin = (
...
+ * @param origin - The Vue component injected origin.
+ *
  * @returns An object containing the {@link XComponentBusAPI}.
  * @internal
  */
-export function getBusAPI(bus: XBus, xComponent: XComponent | undefined): XComponentBusAPI {
+export function getBusAPI(
+  bus: XBus,
+  rootComponent: XComponent | undefined,
+  origin: QueryOrigin
```
javieri-empathy, line 59: Instead of adding a new parameter, because both `rootComponent` and `origin` are properties available in the `this` context, I would just pass the `this`.  This way we have less parameters in the function, and it is already retrieved dynamically, so when we make the change to allow changing it we don't have to remove the code.

As the `this` is now more complex, I would extract it to an interface

```ts
interface PrivateExtendedVueComponent extends Vue {
  xComponent: XComponent | undefined;
  origin: QueryOrigin;
}
```

```ts
): ComponentOptions<Vue> & ThisType<PrivateExtendedVueComponent> => ({
  inject: {
    origin: {
      default: undefined
    }
  },
  created(): void {
    this.xComponent = getRootXComponent(this);

    const aliasAPI = getAliasAPI(this.$store);
    const busAPI = getBusAPI(bus, this);

    this.$x = Object.assign(aliasAPI, busAPI);
  }
});
```

```ts
    emit: <Event extends XEvent>(
      event: Event,
      payload?: XEventPayload<Event>,
      metadata: Omit<WireMetadata, 'moduleName' | 'origin'> = {}
    ) => {
      const rootComponent = component.xComponent;
      const moduleName = rootComponent ? getXComponentXModuleName(rootComponent) : null;
      bus.emit(event, payload as any, { moduleName, origin: component.origin, ...metadata });
      rootComponent?.$emit(event, payload);
    },
    on: bus.on.bind(bus)
```

Another question is, should we try to block passing the origin to the `$x.emit` function like we do with the module name? Or do we want to leave this option open?

javieri-empathy (changes requested):

packages/x-components/src/plugins/x-plugin.mixin.ts
davidmfempathy: I've discussed about it with Beltrán and Guille while we were working in the other related task and we should block it.

packages/x-components/src/plugins/x-plugin.mixin.ts
```diff
@@ -37,20 +50,22 @@ export const createXComponentAPIMixin = (
...
     emit: <Event extends XEvent>(
       event: Event,
       payload?: XEventPayload<Event>,
-      metadata: Omit<WireMetadata, 'moduleName'> = {}
+      metadata: Omit<WireMetadata, 'moduleName' | 'origin'> = {}
     ) => {
-      const moduleName = xComponent ? getXComponentXModuleName(xComponent) : null;
-      bus.emit(event, payload as any, { ...metadata, moduleName });
-      xComponent?.$emit(event, payload);
+      const rootComponent = component.xComponent;
```
tajespasarela, line 65: `rootComponent` is confusing, keep `xComponent` but I would do it directly in the next line:

`const moduleName = getXComponentXModuleName(component.xComponent) ?? null;`

tajespasarela (changes requested):

packages/x-components/src/plugins/x-plugin.mixin.ts
tajespasarela: what do you mean with block?
`metadata: Omit<WireMetadata, 'moduleName' | 'origin'> = {}` this in the emit type?

davidmfempathy: Yes, it's already done

packages/x-components/src/plugins/x-plugin.mixin.ts
davidmfempathy:
```
export function getXComponentXModuleName(component: Vue): XModuleName | null {
  return (component as XComponent)[XComponentModule] ?? null;
}
```

We can't pass a possible `null` value to the getXComponentXModuleName. I can modify it but we'd to the same as this line.

javieri-empathy (approved):

javieri-empathy (approved):

tajespasarela (approved):

PR #129, opened 2021-09-01
feat(url): add `url` x-module config.

javieri-empathy:
# Description

Adds the URL module config, which for now is just a dictionary for the URL mappings. Features like avoiding some values in the URL (for example when page is 1), or removing all the URL parameters when certain condition happens (like we did in EmpathyX when the query is empty) have NOT been done in this PR. The getter has only been touched to check that the configuration works properly.

# Context

This is needed as part of the new URL module, which will help to keep state in the URL in the lifecycle of the application.

# Test

Pay special attention to types, wiring, and the new URLConfigProvided event.


packages/x-components/src/x-modules/url/config.types.ts
```diff
@@ -3,4 +3,11 @@
...
  * @public
  */
-export interface UrlConfig {}
+export interface UrlConfig {
+  query?: string;
+  page?: string;
+  filters?: string;
+  sort?: string;
+  relatedTags?: string;
+  [extraParam: string]: string | undefined;
```
javieri-empathy, line 12: I'm not sure whether the behaviors I stated in the PR description (excluding certain values) will be hardcoded , or should also be part of some configuration. If they should be configurable, I would nest right now these parameters under `UrlConfig['mappings']`

davidmfempathy: What would those mappings be? The resultant values of mapper functions that would be part of the configuration?

davidmfempathy (changes requested):

guillei10 (approved):
lgtm

davidmfempathy (approved):

guillei10 (approved):

PR #130, opened 2021-09-01
feat(search): add origin to the request

tiborux:
WAIT FOR https://github.com/empathyco/x/pull/127

# Pull Request Template #

This pr adds the origin in the searchRequest.

## Motivation and Context ##

This pr is needed to emit the origin in every request. To achieve this, we had to change the wireDispatch and commit to adding the metadata on them because right now every events send a new parameter that its the metaData and we have to recover it from these events using the wiring.

## Type of change ##

- [ X] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

Just search something and check how the origin changes, as the origin is default always, you can set a breakpoint in the mutation to check that it's called every time that you search for a query.


packages/x-components/src/store/store.types.ts
```diff
@@ -30,6 +31,28 @@ export interface RootXStoreState {
  */
 export type RootStoreStateAndGetters = Pick<Store<RootXStoreState>, 'state' | 'getters'>;
 
+/**
+ * State and Getters Store type for {@link RootXStoreState} and payload and metadata types.
```
davidmfempathy, line 35: The link is wrong as it has to be `RootStoreStateAndGetters` but I'd put I'd write its purpose instead of what it has in the interface.

packages/x-components/src/store/store.types.ts
```diff
@@ -30,6 +31,28 @@ export interface RootXStoreState {
...
+ *
+ * @public
+ */
+export type WiringData<Payload> = RootStoreStateAndGetters & {
+  payload: Payload;
+  metadata: WireMetadata;
+};
+
+/**
+ * Type safe which allows the access to the State, the Getters,the payload and metadata of
```
davidmfempathy, line 45: space after the ,

packages/x-components/src/x-modules/search/store/types.ts
```diff
@@ -51,6 +52,8 @@ export interface SearchState extends StatusState {
   totalResults: number;
   /** The extra params property of the state. */
   params: Dictionary<unknown>;
+  /** The origin property of the request. */
```
davidmfempathy, line 55: can we remove it ?

packages/x-components/src/x-modules/search/store/types.ts
```diff
@@ -166,6 +169,12 @@ export interface SearchMutations extends StatusMutations {
    * @param params - The new extra params.
    */
   setParams(params: Dictionary<unknown>): void;
+  /**
+   * Sets the origin of the module.
+   *
+   * @param origin - The new origin.
+   */
+  setOrigin(origin: QueryOrigin): void;
```
davidmfempathy, line 177: can we remove it ?

packages/x-components/src/store/store.types.ts
```diff
@@ -1,4 +1,5 @@
 import { Store } from 'vuex';
```
davidmfempathy, line 1: build failing

davidmfempathy (changes requested):

javieri-empathy (changes requested):

davidmfempathy (dismissed):

javieri-empathy (approved):

davidmfempathy (approved):

PR #134, opened 2021-09-03
Feature/ex 4567 as a developer i want to have ur

tiborux:
# Pull Request Template #

Adds the parameters in the URL state to the URL of the web browser.

## Motivation and Context ##

This is needed to allow users to save the current state of their search.

## Test ##

Register in the full app the module:

`XPlugin.registerXModule(urlXModule);`

Goes to the wiring in the URL module and add an event there to update the url:

```
const wireDispatch = namespacedWireDispatch('url');
export const updateUrl = wireDIspatch('updateUrl');

EmpathizeOpened: {
  updateUrl
}
```

Now every time that the empathize is opened the URL is updated, you can modify the state of the url in the module to check how the values are setted in the url:

```
  state: () => ({
    config: {
      urlParamNames: {
        query: 'ebq',
        relatedTags: 'tag'
      }
    },
    query: '',
    page: 2,
    filters: [],
    sort: '',
    relatedTags: ['red', 'blue'],
    extraParams: {
      category: 'shirts',
      portal: 'platinum'
    }
  }),
```

### DIFF
--- a/packages/x-components/src/x-modules/url/store/actions/update-url.action.ts
+++ b/packages/x-components/src/x-modules/url/store/actions/update-url.action.ts
@@ -16,6 +16,7 @@ export const updateUrl: UrlXStoreModule['actions']['updateUrl'] = ({
   }
 }) => {
   const url = new URL(window.location.href);
+
   forEach(urlMappedParamNames, (_, value) => {
     url.searchParams.delete(value);
   });
--- a/packages/x-components/src/x-modules/url/store/actions/update-url.action.ts
+++ b/packages/x-components/src/x-modules/url/store/actions/update-url.action.ts
@@ -1,4 +1,4 @@
-import { UrlParamKey, UrlParamValue, UrlXStoreModule } from '../types';
+import { MappedParam, MappedParams, UrlXStoreModule } from '../types';
 
 /**
  * Default implementation for the {@link UrlActions.updateUrl}.
@@ -8,18 +8,55 @@ import { UrlParamKey, UrlParamValue, UrlXStoreModule } from '../types';
  *
  * @public
  */
+
 // eslint-disable-next-line max-len
-export const replaceUrl: UrlXStoreModule['actions']['updateUrl'] = ({
+export const updateUrl: UrlXStoreModule['actions']['updateUrl'] = ({
   getters: {
-    urlParams: { scroll, ...urlParams }
+    urlParams: { page, ...params }
   }
 }) => {
   const currentUrl = window.location.href;
   const url = new URL(currentUrl);
 
-  Object.keys(urlParams).forEach((paramKey: UrlParamKey) => {
-    url.searchParams.set(paramKey, encodeURIComponent(urlParams[paramKey]));
-  });
+  clearUrlParams(url, Object.keys({ page, ...params }));
+  urlSetPage(url, page);
+  urlSetParams(url, params);
 
-  window.history.replaceState({ ...window.history.state, ...urlParams }, document.title, url.href);
+  window.history.replaceState({ ...window.history.state }, document.title, url.href);
 };
+
+function clearUrlParams(url: URL, paramsKeys: string[]): void {
+  paramsKeys.forEach(paramKey => {
+    url.searchParams.delete(paramKey);
+  });
+}
+
+function urlSetPage(url: URL, { key, value }: MappedParam): void {
+  if (value !== 1) {
+    url.searchParams.set(key, value.toString());
+  }
+}
+
+function urlSetString(url: URL, { key, value }: MappedParam): void {
+  if (value) {
+    url.searchParams.set(key, value.toString());
+  }
+}
+
+function urlSetArray(url: URL, { key, value: valArr }: MappedParam): void {
+  if (Array.isArray(valArr) && valArr.length) {
+    valArr.forEach(val => {
+      url.searchParams.append(key, val as string);
+    });
+  }
+}
+
+function urlSetParams(url: URL, mappedParams: MappedParams): void {
+  for (const mappedParam of Object.values(mappedParams)) {
+    if (Array.isArray(mappedParam.value)) {
+      urlSetArray(url, mappedParam);
+    } else {
+      urlSetString(url, mappedParam);
+    }
+  }
+}
--- a/packages/x-components/src/x-modules/url/store/actions/update-url.action.ts
+++ b/packages/x-components/src/x-modules/url/store/actions/update-url.action.ts
@@ -1,0 +1,15 @@
+import { UrlXStoreModule } from '../types';
+
+/**
+ * Default implementation for the {@link UrlActions.updateUrl}.
+ *
+ * @param context - The {@link https://vuex.vuejs.org/guide/actions.html | context} of the actions,
+ * provided by Vuex.
+ *
+ * @public
+ */
+// eslint-disable-next-line max-len
+export const updateUrl: UrlXStoreModule['actions']['updateUrl'] = ({ getters }) => {
+  debugger;
+  window.history.replaceState({ ...getters.urlParams }, 'guille');
+};
--- a/packages/x-components/src/x-modules/url/store/actions/update-url.action.ts
+++ b/packages/x-components/src/x-modules/url/store/actions/update-url.action.ts
@@ -1,4 +1,5 @@
-import { MappedParam, MappedParams, UrlXStoreModule } from '../types';
+import { forEach } from '../../../../utils/object';
+import { UrlXStoreModule } from '../types';
 
 /**
  * Default implementation for the {@link UrlActions.updateUrl}.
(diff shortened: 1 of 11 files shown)


packages/x-components/src/x-modules/url/store/actions/update-url.action.ts
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ * Default implementation for the {@link UrlActions.updateUrl}.
+ *
+ * @param context - The {@link https://vuex.vuejs.org/guide/actions.html | context} of the actions,
+ * provided by Vuex.
+ *
+ * @public
+ */
+export const updateUrl: UrlXStoreModule['actions']['updateUrl'] = ({
```
javieri-empathy, line 11: I'm missing the tests for this action

packages/x-components/src/x-modules/url/store/actions/update-url.action.ts
```diff
@@ -0,0 +1,102 @@
...
+};
+
+/**
+ * Clears the parameters in the URL.
+ *
+ * @param url - The current URL.
+ * @param mappedParams - The params in {@link UrlXStoreModule.state} mapped with the overridden keys
+ * if they are.
+ *
+ * @public
```
javieri-empathy, line 33: These module scoped functions shouldn't be public, they should be either private or internal

packages/x-components/src/x-modules/url/store/actions/update-url.action.ts
```diff
@@ -0,0 +1,102 @@
...
+ *
+ * @param url - The current url.
+ * @param param - The param to set in the url.
+ *
+ * @public
+ * */
+function urlSetArray(url: URL, { key, value: values }: MappedParam): void {
+  if (Array.isArray(values) && values.length) {
+    values.forEach(value => {
+      url.searchParams.append(key, value as string);
```
javieri-empathy, line 85: Use `.toString()` like below

packages/x-components/src/x-modules/url/store/actions/update-url.action.ts
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * Sets the array's values in the URL.
+ *
+ * @param url - The current url.
+ * @param param - The param to set in the url.
+ *
+ * @public
+ * */
+function urlSetArray(url: URL, { key, value: values }: MappedParam): void {
+  if (Array.isArray(values) && values.length) {
```
javieri-empathy, line 83: You are already testing if you have an array in the `urlSetParams` function. Maybe you can either split the parameters in this function, or make the `MappedParam` type generic, so you can define the value type as an array here.

Also, you don't need the `values.length` check.

packages/x-components/src/x-modules/url/store/actions/update-url.action.ts
```diff
@@ -0,0 +1,102 @@
...
+ * Sets the rest of params in the URL depending on if they are an array or a string.
+ *
+ * @param url - The current URL.
+ * @param mappedParams - The params in {@link UrlXStoreModule.state} mapped with the overridden keys
+ * if they are.
+ *
+ * @public
+ * */
+function urlSetParams(url: URL, mappedParams: MappedParams): void {
+  for (const mappedParam of Object.values(mappedParams)) {
```
javieri-empathy, line 65: I suggest using our `forEach` method

packages/x-components/src/x-modules/url/store/actions/update-url.action.ts
```diff
@@ -0,0 +1,102 @@
...
+  }
+}) => {
+  const currentUrl = window.location.href;
+  const url = new URL(currentUrl);
+
+  clearUrlParams(url, params);
+  urlSetPage(url, page);
+  urlSetParams(url, params);
+
+  window.history.replaceState({ ...window.history.state }, document.title, url.href);
```
javieri-empathy, line 23: Are the rest of the actions (loading the state from the URL, pushing the state instead of replacing it) going to be added in another PR?

packages/x-components/src/x-modules/url/store/actions/update-url.action.ts
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ * Sets the page in the URL if the value is greater than 1.
+ *
+ * @param url - The current URL.
+ * @param page - The page param to set in the URL.
+ *
+ * @public
+ * */
+function urlSetPage(url: URL, { key, value }: MappedParam): void {
```
javieri-empathy, line 49: This `urlSetXX` methods sound a bit strange, I would rename them to `setUrlXXX`

packages/x-components/src/x-modules/url/store/getters/url-params.getter.ts
```diff
@@ -1,26 +1,30 @@
...
-    { ...rawUrlParams, ...extraParams },
-    (urlParams, key, value) => {
-      const newKey = config.urlParamNames[key] ?? key;
-      urlParams[newKey] = value as UrlParamValue;
-      return urlParams;
+    { ...extraParams, ...params },
+    (mappedParams, key, value) => {
+      const mappedKey = urlParamNames[key] ?? key;
+      mappedParams[key] = {
+        key: mappedKey,
```
javieri-empathy, line 23: I'm not 100% sure about this new abstraction... I think I now have the context to understand it, but I think it kinda divides the state handling into two different parts... On one hand we have the getter generating the URL params, but the action also decides what should go to a URL param and what is not.

For me the getter should generate the URL full dictionary of URL params, and the action should only take care of translating this object to the URL, without having different paths for different parameters. Single responsabilities for both of them. Then you can use an emitter to always keep the URL synced with this object generated by the getter. What do you think?

packages/x-components/src/x-modules/url/store/types.ts
```diff
@@ -53,7 +72,14 @@ export interface UrlMutations {
  *
  * @public
  */
-export interface UrlActions {}
+export interface UrlActions {
+  /**
+   * It updates the URL with values from the store.
```
javieri-empathy, line 77: Remove the `It`. Explain that it replaces the current URL without pushing a new entry to the browser history.

javieri-empathy (changes requested):

packages/x-components/src/x-modules/url/store/getters/url-params.getter.ts
tajespasarela: I agree. This getter should return the final parameters to set un URL, removing for example the page when it is "1".

But I would also move the name mapping responsibility to the Action. This getter should return only an object with the parameters to set in URL, with the internal naming.

Then the action gets the configuration of mapping and updates the URL with the right naming for each parameter.

tajespasarela (dismissed):

packages/x-components/src/x-modules/url/store/getters/url-params.getter.ts
```diff
@@ -17,10 +18,41 @@ export const urlParams: UrlXStoreModule['getters']['urlParams'] = ({
   return reduce(
     { ...rawUrlParams, ...extraParams },
     (urlParams, key, value) => {
-      const newKey = config.urlParamNames[key] ?? key;
-      urlParams[newKey] = value as UrlParamValue;
+      if (isValidParam(value as string | Array<string>) && !isFirstPage(key, value as number)) {
```
javieri-empathy, line 21: - This if can grow really big, and I don't think it will serve us for every case.  For example, right now you can't save `0` or `false` values to the URL, which might be a valid use case.
- Another code smell is that you are doing 2 casts that should be unnecessary.
- And finally, conceptually is not a `reduce`, this is a `filter`. 

With all of that, there goes my proposal:

```
export const urlParams: UrlXStoreModule['getters']['urlParams'] = ({
  config,
  extraParams,
  ...rawUrlParams
}) =>  objectFilter({ ...rawUrlParams, ...extraParams }, isValidParam);

const invalidUrlValues: Record<string, UrlParamValue> = {
  page: 1
};

function isValidParam(key: string, value: UrlParamValue): boolean {
  return Array.isArray(value)
    ? value.length > 0
    : value != null && value !== '' && value !== invalidUrlValues[key];
}
```

packages/x-components/src/x-modules/url/store/actions/update-url.action.ts
```diff
@@ -0,0 +1,34 @@
...
+  getters: { urlParams },
+  state: {
+    config: { urlParamNames }
+  }
+}) => {
+  const url = new URL(window.location.href);
+
+  forEach(urlParams, (paramKey, paramValue) => {
+    const newKey = urlParamNames[paramKey] ?? paramKey;
+    url.searchParams.delete(newKey);
```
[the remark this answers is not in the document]
javieri-empathy: Maybe we can have a new getter that normalises the url param names, so instead of doing `const newKey = urlParamNames[paramKey] ?? paramKey;` you just do `const newKey = getters.urlParamNames[paramKey]`, and use that getter to clean the parameters?

javieri-empathy (changes requested):

guillei10 (dismissed):

pmareke (changes requested):
Good job!

packages/x-components/src/x-modules/url/store/actions/update-url.action.ts
```diff
@@ -0,0 +1,41 @@
...
+ */
+export const updateUrl: UrlXStoreModule['actions']['updateUrl'] = ({
+  getters: { urlParams, urlMappedParamNames },
+  state: {
+    config: { urlParamNames }
+  }
+}) => {
+  const url = new URL(window.location.href);
+
+  forEach(urlMappedParamNames, (_, value) => {
```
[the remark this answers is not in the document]
tiborux: nop, it has to be value because we are using the value where its mapped the value from the config or the default one.

guillei10 (dismissed):

pmareke (approved):

pmareke (changes requested):
Replace the query of the no suggestions feature file

```
    Examples:
      | queryWithoutSuggestions | queryWithSuggestions |
      | asdfg                   | lego                 |
```

pmareke (approved):

guillei10 (approved):

PR #139, opened 2021-09-07
Feature/configure precommit hooks

JaviIG:
# Pull Request Template

Adds husky, linting, fixing and formatting the code before commiting it.

[Husky](https://github.com/typicode/husky) is a tool that allows to run git hooks, for example, it allows to execute code before commiting, which is exactly what we need. Before commiting, we make husky use another tool, called [lint-staged](https://github.com/okonet/lint-staged), which allows us to configure a series of commands to execute against the staged files. For example, we run both eslint and prettier against `.ts` files, but for `.json` files, we only run prettier.

Both eslint and prettier have been configured from the root of the mono-repo, instead of having per-project configurations. This will ensure that every package (except the search-adapter) always uses our recommended style. Now mono-repo projects only have to specify overrides for these shared recommended rules. 

However, because of performance reasons, each project now has now at least one `tsconfig.eslint.json` file, which is configured to not emit any code (to avoid using it with other tools accidentaly), and which specifies which files should be linted. If we don't do that, eslint typescript plugin tries to create somehow different instances of TypeScript global programs, which makes the lint process last a long time, we are speaking here in terms of minutes, instead of just a few seconds with the current approach.

There are lots of files touched, but you should see the same pattern in every package:

1. Removing eslint, prettier, commitizen configs and commands.
2. Adding eslint sub-configs, mainly for TypeScript.

## Motivation and Context

This is needed to keep ensuring a minimum code quality and homogeneity in the monorepo.

## Type of change

- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

`main` branch

## How Has This Been Tested?

Several attempts trying to upload code non following the eslint rules.

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

You can fetch the fork branch into your standard x repo by using this command: `git fetch git@github.com:JaviIG/x.git feature/configure-precommit-hooks:feature/configure-precommit-hooks`


.husky/pre-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
```
javieri-empathy, line 1: When husky is installed, it adds some files to the repo that should be uploaded, like this one or the above one.

PR #141, opened 2021-09-07
feat(animations): add new TranslateFromRight animation

tajespasarela:
EX-4696

This PR adds new TranslateFromRight animation, and also adds enter animation to BaseModal component overlay item.

### DIFF
--- a/packages/x-components/src/components/modals/base-modal.vue
+++ b/packages/x-components/src/components/modals/base-modal.vue
@@ -1,7 +1,7 @@
 <template>
-  <component :is="animation">
+  <component :is="animation" @beforeEnter="showOverlay = false" @afterEnter="showOverlay = true">
     <div v-if="open" class="x-modal" data-test="modal">
-      <div ref="modal" class="x-modal__content" data-test="modal-content" role="dialog">
+      <div ref="modal" class="x-modal__content x-list" data-test="modal-content" role="dialog">
         <!-- @slot (Required) Modal container content -->
         <slot />
       </div>
@@ -9,6 +9,7 @@
         ref="overlay"
         @click="emitOverlayClicked"
         class="x-modal__overlay"
+        :class="{ 'x-modal__overlay--visible': showOverlay }"
         data-test="modal-overlay"
       />
     </div>
@@ -44,6 +45,8 @@
     protected previousBodyOverflow = '';
     /** The previous value of the HTML element overflow style. */
     protected previousHTMLOverflow = '';
+    /** To animate the overlay opacity after and before the animation. */
+    protected showOverlay = true;
 
     public $refs!: {
       modal: HTMLDivElement;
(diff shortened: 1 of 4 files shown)


packages/x-components/src/components/modals/base-modal.vue
```diff
@@ -1,14 +1,15 @@
...
-  <component :is="animation">
+  <component :is="animation" @beforeEnter="showOverlay = false" @afterEnter="showOverlay = true">
     <div v-if="open" class="x-modal" data-test="modal">
-      <div ref="modal" class="x-modal__content" data-test="modal-content" role="dialog">
+      <div ref="modal" class="x-modal__content x-list" data-test="modal-content" role="dialog">
         <!-- @slot (Required) Modal container content -->
         <slot />
       </div>
       <div
         ref="overlay"
```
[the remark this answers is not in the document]
javieri-empathy: The doc is wrong, it talks about emitting `click:body`, but the event is `click:overlay`.

packages/x-components/src/components/modals/base-modal.vue
```diff
@@ -1,14 +1,15 @@
 <template>
-  <component :is="animation">
+  <component :is="animation" @beforeEnter="showOverlay = false" @afterEnter="showOverlay = true">
```
javieri-empathy, line 2: Like you said, this needs a rethink in the future. We want to animate the content div, not the `x-modal` one, probably leaving this wrapper div is a trade of that I would happily do

javieri-empathy (changes requested):

packages/x-components/src/components/modals/base-modal.vue
```diff
@@ -173,7 +174,14 @@
     &__overlay {
       width: 100%;
       height: 100%;
-      position: absolute;
+      position: fixed;
+      background-color: rgba(0, 0, 0, 0.7);
+      opacity: 0;
+
+      &--visible {
```
javieri-empathy, line 181: Shouldn't the class be `is-visible`?

packages/x-components/src/components/modals/base-modal.vue
tajespasarela: That tradeoff breaks my mobile layout.

We will come back to this in other moment

packages/x-components/src/components/modals/base-modal.vue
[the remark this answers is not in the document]
tajespasarela: done

I don't know if this was for some corner case or it was just a refactor garbage

packages/x-components/src/components/modals/base-modal.vue
tajespasarela: you're right. Done

javieri-empathy (approved):

PR #143, opened 2021-09-08
[rgutialo] Adding edocs-synchronization github action

rgutialo:
Adding a github action to complete the workflow between X and eDocs repositories


javieri-empathy (changes requested):

.github/workflows/edocs-synchronization.yml
```diff
@@ -0,0 +1,24 @@
...
+jobs:
+  connect-with-edocs:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Calling to edocs action...
+        run: |
+          curl --request POST \
+          -u empathyco:${GITHUB_TOKEN} \
+          --header 'content-type: application/json' \
+          --url https://api.github.com/repos/empathyco/docs-empathy-documentation/actions/workflows/14218449/dispatches \
```
pmareke, line 21: Instead of hardcoded the URL, maybe we can use an [ENV variable](https://docs.github.com/en/actions/learn-github-actions/environment-variables), what do you think?

javieri-empathy (approved):

PR #148, opened 2021-09-14
[EX-4647] Mock adapter for partial results e2e tests

mavmaf:
# Pull Request Template #

This PR mocks the adapter for the partial results, changes the pertinent tests to make it use it and deletes from the view the components that were not longer needed. 

## Motivation and Context ##

Due to the variability of the real API that was being used for the e2e tests, the data was changing quite often causing the tests to fail. Mocking the adapter for the different components solves this issue, as we now control the data.

EX-4647

As usual, please change the following line in cypress.json and ensure everything works as expected:
"testFiles": ["cucumber/**/*.feature"],


CachedaCodes (commented):
All good. Merge with main to solve conflicts and I'll approve

packages/x-components/tests/e2e/cucumber/no-suggestions/no-suggestions.feature
```diff
@@ -1,16 +1,20 @@
 Feature: No Suggestions component
```
pmareke, line 1: Please remove this file.

packages/x-components/tests/e2e/cucumber/no-suggestions/no-suggestions.spec.ts
```diff
@@ -1,7 +1,15 @@
 import { And, Given } from 'cypress-cucumber-preprocessor/steps';
```
pmareke, line 1: Please remove this file.

packages/x-components/tests/e2e/cucumber/partial-results/partial-results.feature
```diff
@@ -1,7 +1,11 @@
 Feature: Partial results component
 
+  Background:
```
pmareke, line 3: I think in this case we can remove the `Background` as we're doing more requests than we should and move this `Given` step to the Scenarios 2 and 3 just before the `And  no special config for partial-results view`.

packages/x-components/tests/e2e/cucumber/partial-results/partial-results.spec.ts
```diff
@@ -1,20 +1,46 @@
...
 Given('no special config for partial-results view', () => {
-  cy.visit('test/partial-results');
+  cy.visit('test/partial-results?useMockedAdapter=true');
 });
 
 Then('at least {int} related results are displayed', (minResultsWithoutPartials: number) => {
   cy.getByDataTest('regular-result').should('have.length.at.least', minResultsWithoutPartials);
 });
 
 And('no partial results are displayed', () => {
```
pmareke, line 42: `partial results are not displayed` ?

pmareke (changes requested):

packages/x-components/tests/e2e/cucumber/partial-results/partial-results.spec.ts
mavmaf: I would leave it as it is per consistency with the rest of steps across the project (`no identifier results are displayed`, `no query suggestion are displayed`, `no history queries are displayed`...)

pmareke (approved):

CachedaCodes (approved):

PR #150, opened 2021-09-14
refactor: use `NamedModel` generic parameter to better type models.

javieri-empathy:
EX-4734

# Description

This little refactor just makes use of the `NamedModel` generic parameter in all the models that extends it.  In facet/filters the hierarchy is more complex, and this can't be done through extending, so I left it overriding values. In any case if the filter/facet has a wrong model name TS will still complain, so it is no issue.


pmareke (changes requested):
The build is failing because of the `Suggestions`:

```
src/empathy/mappers/response/suggestions/empathy-suggestion.mapper.ts(16,7): error TS2322: Type '"PopularSearch" | "QuerySuggestion"' is not assignable to type '"Suggestion" | undefined'.
  Type '"PopularSearch"' is not assignable to type '"Suggestion" | undefined'.
```

Please take a look.

packages/x-components/src/__stubs__/suggestions-stubs.factory.ts
```diff
@@ -58,7 +58,7 @@ export function createSuggestionStub(query: string, suggestion?: Partial<Suggest
     query,
     totalResults: 10,
     results: [],
-    modelName: 'Suggestion',
+    modelName: 'QuerySuggestion',
```
pmareke, line 61: This file is getting [errors](https://github.com/empathyco/x/pull/150/checks?check_run_id=3599561851#step:6:3131) regarding the ModelName.

Also we didn't update the `getSuggestionsStub` method as in the other modules:

```javascript
/**
 * Creates a list of {@link Suggestions} of the given length.
 *
 * @param length - The length of the list to create.
 * @returns A list of simple {@link Suggestions}s.
 */
export function getSuggestionsStub(length = 3): Suggestion[] {
  return Array.from({ length }, (_, index) => createSuggestionStub(index));
}
```

pmareke: If we do this change, we need to change in a lot of files (sorry for that), what do you think?

pmareke (changes requested):
Fix errors and think about the `getSuggestionsStub` refactor

pmareke (changes requested):
Also I've just realised that in the popular searches stub we are calling the suggestions stub so the model name is always `QuerySuggestion` and `PopularSearch` is never used.

tiborux (changes requested):
Fix conflicts

packages/x-components/src/__stubs__/index.ts
```diff
@@ -1,12 +1,13 @@
-export * from './banners-stubs.factory';
```
pmareke, line 1: Please for my TOC.

```javascript
export * from './banners-stubs.factory';
export * from './facets-stubs.factory';
export * from './filters-stubs.factory';
export * from './history-queries-stubs.factory';
export * from './next-queries-stubs.factory';
export * from './partials-results-stubs.factory';
export * from './popular-searches-stubs.factory';
export * from './promoteds-stubs.factory';
export * from './query-suggestions-stubs.factory';
export * from './redirections-stubs.factory';
export * from './related-tags-stubs.factory';
export * from './results-stubs.factory';
export * from './search-response-stubs.factory';
```

packages/x-components/src/__stubs__/popular-searches-stubs.factory.ts
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Creates a list of generic popular searches.
+ *
+ * @param amount - The number of popular searches to create.
+ *
+ * @returns An array of popular searches.
+ */
+export function getPopularSearchesStub(amount = 3): Suggestion[] {
+  return Array.from({ length: amount }, (_, index) =>
```
pmareke, line 11: `Array.from<number, Suggestion>`

packages/x-components/src/__stubs__/query-suggestions-stubs.factory.ts
```diff
@@ -1,46 +1,35 @@
...
-    },
-    {
-      facets: [],
-      query: `bë&éf shórt ribs ${query}`,
-      key: `beef short ribs ${query}`,
-      modelName: modelName
-    }
-  ];
+export function getQuerySuggestionsStub(query: string, amount = 3): Suggestion[] {
+  return Array.from({ length: amount }, (_, index) =>
```
pmareke, line 12: `Array.from<number, Suggestion>`

packages/x-components/src/x-modules/popular-searches/components/__tests__/popular-search.spec.ts
```diff
@@ -36,12 +35,12 @@ describe('testing popular search item component', () => {
     const wrapperComponent = {
       template: `
         <PopularSearch :suggestion="suggestion">
-          <template #default="{ suggestion }">
-            <svg data-test="icon" height="10" width="10">
-              <circle cx="5" cy="5" r="4" stroke="black" />
-            </svg>
-            <span data-test="query" :aria-label="suggestion.query">{{ suggestion.query }}</span>
-          </template>
+        <template #default="{ suggestion }">
```
pmareke, line 38: Please rollback the indentation here.

packages/x-components/src/x-modules/popular-searches/components/__tests__/popular-searches.spec.ts
```diff
@@ -43,11 +43,11 @@ describe('testing popular searches component', () => {
     const wrapperComponent = {
       template: `
         <PopularSearches>
-          <template #suggestion-content="suggestionContentScope">
-            <img src="./popular-search-icon.svg" class="x-popular-search__icon" data-test="icon"/>
-            <span class="x-popular-search__query" :data-index="suggestionContentScope.index"
-                  data-test="query">{{ suggestionContentScope.suggestion.query }}</span>
-          </template>
+        <template #suggestion-content="suggestionContentScope">
```
pmareke, line 46: Please rollback the indentation here.

packages/x-components/src/x-modules/popular-searches/components/__tests__/popular-searches.spec.ts
```diff
@@ -73,18 +73,18 @@ describe('testing popular searches component', () => {
...
-                <img src="./popular-search-icon.svg"
-                     class="x-popular-search__icon"
-                     data-test="icon"/>
-                <span class="x-popular-search__query"
-                      data-test="query">{{ suggestion.query }}</span>
-              </template>
-            </PopularSearch>
-            <button data-test="custom-button">Custom Behaviour</button>
-          </template>
+        <template #suggestion="{suggestion}">
```
pmareke, line 76: Please rollback the indentation here.

packages/x-components/src/__stubs__/popular-searches-stubs.factory.ts
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * Creates a popular search stub with the provided options.
+ *
+ * @param query - The query of the suggestion.
+ * @param suggestion - An optional object with fields to override the suggestion.
+ *
+ * @returns A popular search.
+ */
+export function createPopularSearch(query: string, suggestion?: Partial<Suggestion>): Suggestion {
```
pmareke, line 24: Can we re-use the `createQuerySuggestion` method adding the modelName as parameter and remove this one?

pmareke (changes requested):
Well done!

packages/x-components/src/__stubs__/popular-searches-stubs.factory.ts
javieri-empathy: Not needed. The fake array you are creating doesn't need any type, which in your case is the `number` type. And as you are returning if you don't fulfill the `Suggestion[]` type TS will complain. You earn nothing but writing more code here.

packages/x-components/src/x-modules/popular-searches/components/__tests__/popular-search.spec.ts
javieri-empathy: I don't know why Webstorm keeps doing this to me

packages/x-components/src/__stubs__/popular-searches-stubs.factory.ts
javieri-empathy: No because of the missunderstandings. If you want to create a query suggestion you can use the `createQuerySuggestionStub`, if you want to create a  popular search you can use the `createPopularSearchStub`. Passing the model name from the beginning was a bad idea because it allowed you to create invalid models.

packages/x-components/src/__stubs__/popular-searches-stubs.factory.ts
pmareke: I was trying to follow the same approach in all the stubs, we have most of then with the `<type>`

pmareke (approved):

tiborux (approved):

PR #151, opened 2021-09-14
feat(no-suggestions): Remove `no-suggestions` module

javieri-empathy:
BREAKING-CHANGE: Remove `no-suggestions` module

EX-4736

# Description

Removes the no-suggestions module as it was never used and not needed. Nothing else.


pmareke (changes requested):
I found some `no-suggestions` stuff in `.github/contributing/commits.md` and `packages/x-components/contributing/commits.md`, please remove it also.

tajespasarela (dismissed):

tajespasarela (approved):

PR #152, opened 2021-09-15
Feature/ex 4712 as a developer i want to update the store from the URL

guillei10:
# Description #

The goal of this PR is to implement the action that will allow the Url module to receive parameters from the url and update the state. 

In order to do this, the action has to check one by one all the params from the state and look if some of them have been updated or added in the url. Then it also validates which type is the parameter in the state to be able to parse the url value from the url to the state with the correct type.

## Motivation and Context ##

This is part of the url module implementation. There are also some changes to adapt from previous PRs about this module.

## Test  ##

As previous PRs, you can add a wire that links an Event with this action and fire that event in the interface. Then you can debug the state of the module to see if it updates accordingly.


packages/x-components/src/x-modules/url/store/actions/update-store-from-url.action.ts
```diff
@@ -0,0 +1,66 @@
...
+ */
+function getParamByType(
+  urlSearchParams: URLSearchParams,
+  stateValue: UrlParamValue,
+  urlParam: UrlParamKey
+): UrlParamValue {
+  switch (typeof stateValue) {
+    case 'number':
+      return Number(urlSearchParams.get(urlParam));
+    case 'boolean':
```
pmareke, line 58: At this point you're sure that the value is present, so you can use:

```javascript
case 'boolean':
  return urlSearchParams.get(urlParam)!.toLowerCase() === 'true';
case 'string':
  return urlSearchParams.get(urlParam)!;
```

pmareke (changes requested):

packages/x-components/src/x-modules/url/store/types.ts
```diff
@@ -58,12 +69,17 @@ export interface UrlMutations {
...
   updateUrl(): void;
+  /**
+   * Updates the store with values from the URL.
+   *
+   * @public
+   */
+  updateStoreFromUrl(): void;
 }
 
 /**
```
tiborux, line 85: Can you document the mutations? The documentation is missing.

packages/x-components/src/x-modules/url/store/actions/update-store-from-url.action.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  commit('setParams', mappedParams);
+  commit('setExtraParams', mappedExtraParams);
+};
+
+/**
+ *
+ * Gets the parameter from the url depending on the state parameter type.
+ *
+ * @param urlSearchParams - All the search params
```
tiborux, line 42: All these sentences needs a dot at the end.

tiborux (changes requested):

pmareke (changes requested):
The build is failing because of the type of the `setParams` mutation:

https://github.com/empathyco/x/pull/152/checks?check_run_id=3609406247#step:4:235

It should be a `Dictionary<Params>`

pmareke (approved):

pmareke (approved):
Great job!

pmareke (approved):

tiborux (approved):

PR #157, opened 2021-09-16
feat(design-system): add utilities

tajespasarela:
background-color, border-color, border-radius, border-width, box-shadow, font-color, margin,
padding.

EX-4746


packages/x-components/src/design-system/utilities/border-width.scss
```diff
@@ -0,0 +1,34 @@
+@import '../mixins';
+
+*[class*='x-border-width--'] {
+  border: 0;
```
javieri-empathy, line 4: What is this for? I think it is breaking combining border width and color utilities. Using border-width here seems to fix the issue, but I don't know if it is breaking something else

javieri-empathy (changes requested):

packages/x-components/src/design-system/utilities/box-shadow.scss
```diff
@@ -1,7 +1,9 @@
-@import '../../mixins';
+@import '../mixins';
 
+.x-shadow--none {
+  box-shadow: var(--x-string-box-shadow-none);
```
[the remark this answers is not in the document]
tajespasarela: you're right, it makes no sense. changed

packages/x-components/src/design-system/utilities/border-width.scss
tajespasarela: This is to reset the default browser styles, with a width of 3px;
Tailwind resets to all elements with `*` selector. 
By default an element doesn't have border, even with `x-border-color`. You have to add also a width.

If I remove this I get this:
image

instead of this:
image

javieri-empathy: Yeah, I got the reset width part.  My point is that in those design system view examples you are setting the border color to lead, but it still black, and that's because the `border:0` is resetting the color. The attribute selector and class selector have the same specifity, but the color one is loaded earlier, so it is overridden by the reset:

image

tajespasarela: fucking hell!
Ok it is changing `border` by `border-width`

javieri-empathy (approved):

PR #158, opened 2021-09-16
Update url action regarding pusheable parameters

pmareke:
EX-4713


packages/x-components/src/x-modules/url/store/__tests__/actions.spec.ts
```diff
@@ -54,6 +54,67 @@ describe('testing Url module actions', () => {
...
+      });
+
+      await store.dispatch(actionKeys.updateUrl);
+
+      // eslint-disable-next-line @typescript-eslint/unbound-method
+      expect(window.history.pushState).toHaveBeenCalled();
+    });
+  });
+
+  describe(`${actionKeys.updateStoreFromUrl}`, () => {
```
tiborux, line 84: Remove this describe, its duplicated.

packages/x-components/src/x-modules/url/store/__tests__/actions.spec.ts
```diff
@@ -54,6 +54,67 @@ describe('testing Url module actions', () => {
...
+          relatedTags: [],
+          page: 1,
+          scroll: 2
+        }
+      });
+
+      await store.dispatch(actionKeys.updateUrl);
+
+      // eslint-disable-next-line @typescript-eslint/unbound-method
+      expect(window.history.pushState).toHaveBeenCalled();
```
tiborux, line 80: As you said, maybe we can add another test to check replaceSate or we can add it in the first It:

```
    it('should add the correct params to the url', async () => {
       window.history.replaceState = jest.fn();
       .
      .
      expect(window.history.replaceState).toHaveBeenCalled();
```

tiborux (changes requested):

guillei10 (changes requested):

guillei10 (approved):

tiborux (approved):

PR #162, opened 2021-09-17
[EX-4759] Update Cypress version

mavmaf:
# Pull request template
In order to update Cypress version, `npm install --save-dev cypress@8.4.0` was run and it was also needed to manually change the version in `packages/react-wrapper/package.json` and `packages/x-components/package.json`.

## Motivation and context
Several improvements have been made since v7.7.0. Some of them are the following:

- Projects will now open much faster on MacOS Big Sur with improved browser detection performance
- Fixed a regression in 7.2.0 that would cause cy.visit() to take longer to fire its load event in some circumstances
- Fixed a regression in 8.3.0 where the correct exit code would not be issued during cypress run-ct 
- Upgraded Chrome browser version used during cypress run 
- Multiple fixes for cy.intercept

- `Main`

## How has this been tested?

Run tests using Cypress


pmareke (approved):
Great job, everything works fine as before so for me it's ok!

pmareke (changes requested):
In order to use the latest version of Cypress (8.4.1), please run `npm update`, discard the changes in the package.json and just leave the ones in the package-lock.json.

package.json
```diff
@@ -13,7 +13,8 @@
     "publish-release": "node scripts/publish-release"
   },
   "devDependencies": {
-    "conventional-changelog-conventionalcommits": "~4.6.0",
+    "conventional-changelog-conventionalcommits": "^4.6.1",
```
pmareke, line 16: Revert the changes of this file please, we only need the ones in the lock files.!

packages/react-wrapper/package.json
```diff
@@ -46,7 +46,7 @@
     "@types/react": "~16.9.0",
     "@types/react-dom": "~16.9.0",
     "commitizen": "~4.1.2",
-    "cypress": "~7.7.0",
+    "cypress": "~8.4.1",
```
pmareke, line 49: Revert the changes of this file please, we only need the ones in the lock files.!

packages/x-components/package.json
```diff
@@ -104,7 +104,7 @@
     "@vue/test-utils": "~1.0.3",
     "autoprefixer": "~9.7.4",
     "commitizen": "~4.0.3",
-    "cypress": "~7.7.0",
+    "cypress": "~8.4.1",
```
pmareke, line 107: Revert the changes of this file please, we only need the ones in the lock files.!

pmareke (changes requested):
Leave the changes only in the lock files.

CachedaCodes (dismissed):

pmareke (approved):
Top!

CachedaCodes (approved):

PR #167, opened 2021-09-20
feat(components): Add `TranslateFromLeft` animation.

javieri-empathy:
EX-4772

# Description

Just adds a missing simple translate animation


packages/x-components/src/components/result/base-result-image.vue
```diff
@@ -180,45 +191,47 @@
   .x-result-picture {
     min-width: 1px;
     min-height: 1px;
+    position: relative;
+
     &__image {
       max-width: 100%;
       max-height: 100%;
-      object-fit: fill;
+      object-fit: cover;
```
tajespasarela, line 199: change this by `contain`

packages/x-components/src/components/result/base-result-image.vue
```diff
@@ -1,33 +1,36 @@
...
+        v-if="imageSrc"
+        v-show="hasImageLoaded"
+        @error="flagImageAsFailed"
+        @load="flagImageLoaded"
+        :alt="result.name"
+        :src="imageSrc"
+        class="x-picture__image x-result-picture__image"
+        data-test="result-picture__image"
+      />
+      <NoElement v-else-if="hasAllImagesFailed" class="x-picture__image x-picture__image--fallback">
```
tajespasarela, line 23: Why is this a `NoComponent`? can't be a `<template>`?

tajespasarela (changes requested):

packages/x-components/src/components/result/base-result-image.vue
javieri-empathy: Are you asking me 🤣? 
image

I remember discussing about using `NoElement` to add classes to slots, and I guess that's the case here, adding the design system classes to the slot content. But I'm not sure if there are other reasons.

tajespasarela: Don't play with my poor memory!

I guess it was for the classes as you say.

tajespasarela (approved):

PR #169, opened 2021-09-20
feat(url): set the related tags from the url in the store and viceversa

pmareke:
EX-3513


packages/x-components/src/x-modules/search/store/actions/set-query-from-url.action.ts
```diff
@@ -0,0 +1,11 @@
+import { SearchXStoreModule } from '../types';
+
+export const setQueryFromUrl: SearchXStoreModule['actions']['setQueryFromUrl'] = (
```
guillei10, line 3: Wait is this the same exact code than the other action in the search-box module? Is there a possibility to move this function somewhere else and just use it once in the two modules instead of having to copies?

packages/x-components/src/x-modules/search-box/store/actions/set-query-from-url.action.ts
```diff
@@ -0,0 +1,11 @@
+import { SearchBoxXStoreModule } from '../types';
+
+export const setQueryFromUrl: SearchBoxXStoreModule['actions']['setQueryFromUrl'] = (
```
guillei10, line 3: I know this action is pretty simple but maybe we should add a documentation here as well.

guillei10 (changes requested):

packages/x-components/src/x-modules/related-tags/wiring.ts
```diff
@@ -75,13 +75,27 @@ export const toggleRelatedTagWire = wireDispatch('toggleRelatedTag');
  */
 export const clearSelectedRelatedTags = wireCommit('setSelectedRelatedTags', []);
 
+/**
+ * Sets the selected related tags.
+ *
+ * @public
+ */
+export const setSelectedRelatedTagsWire = wireCommit('setSelectedRelatedTags');
```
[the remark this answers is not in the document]
pmareke: Sure!

packages/x-components/src/x-modules/search-box/store/actions/set-query-from-url.action.ts
pmareke: Sure!

packages/x-components/src/x-modules/search/store/actions/set-query-from-url.action.ts
pmareke: After a couple of ideas, I didn't like none of them. 

- If you create a helper method you need to pass around the commit and the query, and the types are a mess.
- If you create a generic helper method like `export const setQueryFromUrl: T['actions']['setQueryFromUrl'] = <T>(...)` it doesn't work either.

We can take a look together but I think that these actions should keep separate as in the future they could have different implementations

tiborux (changes requested):

tiborux (approved):

guillei10 (approved):

PR #170, opened 2021-09-21
Cleanup and format code

javieri-empathy:
# Description

Adds a new design component, the `badge`. This can be used to display minimal amounts of information on top of other components. For example with the selected filters count.

image

### DIFF
--- a/packages/x-components/src/design-system/components/badge/default.scss
+++ b/packages/x-components/src/design-system/components/badge/default.scss
@@ -3,7 +3,6 @@
   border-radius: var(--x-size-border-radius-badge-default);
   border-style: solid;
   border-width: var(--x-size-border-width-badge-default);
-  border-color: var(--x-color-border-badge-default);
 
   // font
   font-size: var(--x-size-font-badge-default);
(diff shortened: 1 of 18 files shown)


packages/x-components/src/design-system/components/badge/default.scss
```diff
@@ -0,0 +1,28 @@
+.x-badge {
+  // border
+  border-radius: 50%;
```
pmareke, line 3: This border radius is going to be always round? Should be allow the option to change it?

packages/x-components/src/design-system/components/badge/default.scss
```diff
@@ -0,0 +1,28 @@
...
+
+  // font
+  font-size: var(--x-size-font-badge-default);
+  font-weight: var(--x-number-font-weight-badge-default);
+  text-align: center;
```
pmareke, line 8: Move to the layout section please.

```css
// layout
position: absolute;
text-align: center;
vertical-align: middle;
```

pmareke (changes requested):
Grande capo!

tajespasarela (approved):

PR #175, opened 2021-09-23
refactor(facets): fix facets performance issues

tajespasarela:
EX-4800

### DIFF
--- a/packages/x-components/src/x-modules/facets/store/module.ts
+++ b/packages/x-components/src/x-modules/facets/store/module.ts
@@ -24,10 +24,16 @@ export const facetsXStoreModule: FacetsXStoreModule = {
   },
   mutations: {
     setFilter(state, filter) {
-      Vue.set(state.filters, filter.id, filter);
+      Vue.set(state.filters, filter.id, Object.freeze(filter));
+    },
+    setFilters(state, filters) {
+      filters.forEach(filter => Vue.set(state.filters, filter.id, Object.freeze(filter)));
     },
     removeFilter(state, { id }) {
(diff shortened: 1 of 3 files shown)


packages/x-components/src/x-modules/facets/store/module.ts
```diff
@@ -24,11 +24,17 @@ export const facetsXStoreModule: FacetsXStoreModule = {
   },
   mutations: {
     setFilter(state, filter) {
-      Vue.set(state.filters, filter.id, filter);
+      Vue.set(state.filters, filter.id, Object.freeze(filter));
+    },
+    setFilters(state, filters) {
+      filters.forEach(filter => Vue.set(state.filters, filter.id, Object.freeze(filter)));
```
pmareke, line 30: This is the same code as the `setFilter` method

```javascript
filters.forEach(filter => this.setFilter(state, filter));
```

packages/x-components/src/x-modules/facets/store/module.ts
```diff
@@ -24,11 +24,17 @@ export const facetsXStoreModule: FacetsXStoreModule = {
...
     },
     removeFilter(state, { id }) {
       Vue.delete(state.filters, id);
     },
+    removeFilters(state, filters) {
+      filters.forEach(({ id }) => Vue.delete(state.filters, id));
```
pmareke, line 36: Same here

```javascript
 filters.forEach(({ id }) => this.removeFilter(state, filter));
```

javieri-empathy (approved):

pmareke (changes requested):
Great job!

PR #177, opened 2021-09-24
Add filters in the url

pmareke:
EX-3511


packages/x-components/src/x-modules/search/store/actions/set-params-from-url.action.ts
```diff
@@ -24,4 +26,8 @@ export const setParamsFromUrl: SearchXStoreModule['actions']['setParamsFromUrl']
   if (newPage) {
     commit('setPage', newPage);
   }
+
+  if (newFilters) {
+    commit('setSelectedFilters', createRawFilters(newFilters));
```
javieri-empathy, line 31: Is this necessary? Doesn't the facet's module communicate the selected filters in time for the first request?

packages/x-components/src/x-modules/facets/store/actions/set-filters-from-url.action.ts
```diff
@@ -0,0 +1,29 @@
...
+/**
+ * Default implementation for the {@link FacetsActions.setUrlParamsFromTheUrl}.
+ *
+ * @param context - The {@link https://vuex.vuejs.org/guide/actions.html | context} of the actions,
+ * provided by Vuex.
+ * @param urlParams - The url params from the url.
+ *
+ * @public
+ */
+export const setUrlParamsFromTheUrl: FacetsXStoreModule['actions']['setUrlParamsFromTheUrl'] = (
```
javieri-empathy, line 14: Why is this file called `set-filters-from-url.action`, but the exported method `setUrlParamsFromTheUrl`?. Both names should match the action name

javieri-empathy (changes requested):

packages/x-components/src/x-modules/facets/store/actions/set-filters-from-url.action.ts
```diff
@@ -0,0 +1,29 @@
+import { createRawFilters, Dictionary } from '../../../../utils';
+import { UrlParamValue } from '../../../url';
```
javieri-empathy, line 2: Avoid importing from barrels. It is very easy to create cyclical dependencies that can result in errors at runtime. In this case you are even importing from a barrel that should be exporting this action, which is worse.

packages/x-components/src/x-modules/facets/store/actions/set-filters-from-url.action.ts
```diff
@@ -0,0 +1,29 @@
...
+ */
+export const setUrlParamsFromTheUrl: FacetsXStoreModule['actions']['setUrlParamsFromTheUrl'] = (
+  { commit },
+  { query, filters }: Dictionary<UrlParamValue>
+) => {
+  const newQuery = query as string;
+  const newFilters = filters as string[];
+
+  if (newQuery) {
+    commit('setQuery', newQuery);
```
tajespasarela, line 22: Why is necessary the `query` in the `Facets` module?

I only see it in an emitter of a `FacetsQueryChanged` Event which is never used.

I would remove all of it: the query, the emitter and the Event

packages/x-components/src/x-modules/facets/store/types.ts
```diff
@@ -99,7 +101,13 @@ export interface FacetsMutations {
  *
  * @public
  */
-export interface FacetsActions {}
+export interface FacetsActions {
+  /**
+   * Receives a list of params from the url, builds the entities and
+   * set them in the store.
+   */
+  setUrlParamsFromTheUrl(urlParams: Dictionary<UrlParamValue>): void;
```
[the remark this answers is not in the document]
tajespasarela: I would keep the `set` verb, to keep consistency in all mutations/actions we have:
`setFiltersFromUrl`

As I said up, I would remove the query and then this action only set the filters.

packages/x-components/src/x-modules/search/store/actions/set-params-from-url.action.ts
tajespasarela: And if not, we have to fix it in another way. 

We can not have the facets module and the search module doing the same.

packages/x-components/src/x-modules/url/store/module.ts
```diff
@@ -44,6 +44,12 @@ export const urlXStoreModule: UrlXStoreModule = {
     setRelatedTags(state, relatedTags) {
       state.params.relatedTags = relatedTags;
     },
+    setFilters(state, newFilters) {
+      state.params.filters = newFilters;
+    },
+    resetFilters(state) {
```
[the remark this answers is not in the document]
tajespasarela: right, remove this mutation

tajespasarela (changes requested):

packages/x-components/src/components/scroll/base-scroll.vue
```diff
@@ -58,7 +58,7 @@
      *
      * @internal
      */
-    @XOn(['UserAcceptedAQuery', 'SortChanged'])
+    @XOn(['UserAcceptedAQuery', 'SortChanged', 'UserPickedARelatedTag'])
```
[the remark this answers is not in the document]
pmareke: I think we should change algo the `UserPickedARelatedTag ` right?

packages/x-components/src/x-modules/facets/service/__tests__/facets-service.spec.ts
```diff
@@ -296,7 +297,12 @@ describe('testing facets service', () => {
           getStoreEditableNumberRangeFilter(priceUpTo10Filter)
         ])
       ).toBe(false);
-      expect(getSelectedFilters()).toEqual([]);
+      expect(getSelectedFilters()).toEqual([
```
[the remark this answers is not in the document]
pmareke: The problem is that the test is broken and the `[]` is not the expected response, that's why I did the change.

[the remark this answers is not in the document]
pmareke: I rollback the entire spec file as @javieri-empathy suggested just changing `updateFacets` is ok.

packages/x-components/src/x-modules/url/wiring.ts
```diff
@@ -37,6 +44,22 @@ export const setRelatedTagsWire = mapWire(
...
+ * Sets the {@link @empathyco/x-types#Filter | filters }.
+ *
+ * @public
+ */
+export const setFiltersWire = mapWire(wireCommit('setFilters'), (filters: Filter[]) =>
+  filters.map(f => f.id)
+);
+
+/**
+ * Resets the {@link RelatedTag | related tags }.
```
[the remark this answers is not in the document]
pmareke: Removed it.

packages/x-components/src/x-modules/facets/store/actions/set-filters-from-url.action.ts
pmareke: Fixed.

tiborux (changes requested):

tiborux:
Remove the FacetsQueryChanged from the event.types.ts in the facets module

packages/x-components/src/x-modules/facets/wiring.ts
```diff
@@ -74,11 +74,11 @@ const clearAllFiltersWire = wireFacetsServiceWithoutPayload('clearFilters');
 const selectFilterWire = wireFacetsService('select');
 
 /**
- * Sets the facets state `query`.
+ * Saves the params from the url.
  *
  * @public
  */
-const setFacetsQuery = facetsWireCommit('setQuery');
+export const setFiltersFromUrl = facetsWireDispatch('setFiltersFromUrl');
```
tiborux, line 81: is this exported for any reason?

tiborux (changes requested):

packages/x-components/src/x-modules/facets/wiring.ts
pmareke: My fault.

tiborux (approved):

guillei10 (approved):

PR #178, opened 2021-09-29
Feature/ex 4718 minimise initial duplicated requests

javieri-empathy:
EX-4718

# Description

This PR tries to minimise the duplicated requests by using a `debounce` with no time over the emitters. Vue already tries to do this with watchers, but it tries to do so using microtasks (promises). We use a greedier strategy using macrotasks (a debounce which uses setTimeout). This should lower the priority of all `XXXChanged` events, so Vue reactivity together with our wiring can have more time to group changes.

On the other hand, it also makes all `fetchXXX` actions receive the request via payload. This is to ensure that they use the request object they was triggered with always. Previously, two subsequent requests caused by the asynchronicity we are dealing with, were triggered because of two different state changes, but the request they made was the same because the getter that triggered the first request was already invalidated before the request was made.


packages/x-components/src/plugins/__tests__/x-plugin.spec.ts
```diff
@@ -403,7 +411,8 @@ describe('testing X Plugin', () => {
     it('store-emitters emit a changed event when the observed store state changes', async () => {
       component.vm.$x.emit('UserIsTypingAQuery', 'New York strip steak');
 
-      await localVue.nextTick();
+      await localVue.nextTick(); // Needed so Vue has updated the reactive dependencies.
```
pmareke, line 414: Can we use here the new method `waitNextTick`?

packages/x-components/src/plugins/__tests__/x-plugin.spec.ts
```diff
@@ -440,9 +472,21 @@ describe('testing X Plugin', () => {
...
       component.vm.$x.emit('UserIsTypingAQuery', '');
 
-      await localVue.nextTick();
+      await waitNextTick();
 
       expect(searchBoxQueryChangedSubscriber).toHaveBeenCalledTimes(0);
     });
   });
 });
+
```
pmareke, line 481: Can we extract this method to a `utils` folder an avoid the code in these two tests duplicate?

```javascript
import { default as Vue, VueConstructor } from 'vue';
/**
 * Waits for Vue's reactivity to update getters and watchers, and flushes the pending emitters.
 *
 * @remarks It needs `jest.useFakeTimers()` to have been called to wait for the emitters.
 * @returns A promise that resolves after the reactivity has been updated and the pending emitters
 * have been run.
 */
export async function waitNextTick(localVue: VueConstructor<Vue>): Promise<void> {
  await localVue.nextTick();
  jest.runAllTimers();
}
```

packages/x-components/src/store/__tests__/utils.ts
```diff
@@ -12,4 +13,5 @@ export type SafeStore<
   Getters extends Dictionary,
   Mutations extends MutationsDictionary<Mutations>,
   Actions extends ActionsDictionary<Actions>
-> = Omit<XActionContext<State, Getters, Mutations, Actions>, 'rootGetters' | 'rootState'>;
+> = Omit<Store<State>, 'dispatch' | 'commit' | 'state' | 'getters'> &
```
pmareke, line 16: Why are we using string literals instead of types as before?

packages/x-components/src/x-modules/facets/components/facets/__tests__/facets-provider.spec.ts
```diff
@@ -155,10 +162,12 @@ function renderFacetsProviderComponent({
     async selectFilters(filters: Filter[]): Promise<void> {
       filters.forEach(filter => DefaultFacetsService.instance.select(filter));
       await localVue.nextTick();
```
pmareke, line 164: If we extract the method `waitForNextTick` we could use here too.

packages/x-components/src/x-modules/facets/components/facets/__tests__/facets-provider.spec.ts
```diff
@@ -155,10 +162,12 @@ function renderFacetsProviderComponent({
...
+      jest.runAllTimers();
     },
     async deselectFilters(filters: Filter[]): Promise<void> {
       filters.forEach(filter => DefaultFacetsService.instance.deselect(filter));
       await localVue.nextTick();
```
pmareke, line 169: And here.

packages/x-components/src/x-modules/identifier-results/store/__tests__/actions.spec.ts
```diff
@@ -1,47 +1,65 @@
...
+  IdentifierResultsMutations,
+  IdentifierResultsState
+} from '../types';
 import { resetIdentifierResultsStateWith } from './utils';
 
 describe('testing identifier results module actions', () => {
   const mockedResults = getResultsStub();
   const adapter = getMockedAdapter({ searchById: { results: mockedResults } });
 
-  const actionKeys = map(identifierResultsXStoreModule.actions, action => action);
```
pmareke, line 14: I like the use of this map in order to know which actions the module has and test all of them, why did you remove them?

packages/x-components/src/x-modules/identifier-results/store/actions/fetch-identifier-results.action.ts
```diff
@@ -4,17 +4,16 @@ import { IdentifierResultsXStoreModule } from '../types';
 /**
  * Default implementation for the {@link IdentifierResultsActions.fetchIdentifierResults}.
  *
- * @param context - The {@link https://vuex.vuejs.org/guide/actions.html | context} of the actions,
+ * @param _ - The {@link https://vuex.vuejs.org/guide/actions.html | context} of the actions,
```
pmareke, line 7: Maybe `_context`? `_` is quite weird to read I think.

pmareke (changes requested):
I would rollback the remove of the map actions in all the tests and try to extract the waitNextTick method in order to use it everywhere.

tajespasarela (dismissed):

packages/x-components/src/x-modules/identifier-results/store/__tests__/actions.spec.ts
tajespasarela: no necessary any more thanks to TS

packages/x-components/src/store/__tests__/utils.ts
tajespasarela: I don't got you. Those "strings" are actually members of Types

tajespasarela (approved):

pmareke (approved):

PR #179, opened 2021-09-30
Feature/ex 4827 refactor url module

guillei10:
# Description

The goal of this PR is to refactor some parts of the URL module:

- Change the `updateStoreFromUrl` action into an `ActionClass`
- Change `relatedTags` property to singular to be displayed correctly in the URL.
- Added a `isLoadedFromUrl` flag that controls the flux and avoids undesired loops or duplicated events.

## Motivation

A review and improvement of the URL Module.

## Test

Play with the parameters of the Url (query, related tags, page) and verify everything is set correctly.


tajespasarela:
When the URL is loaded with query, related tag & page, the page disappears

https://user-images.githubusercontent.com/5759712/135474573-130a943f-73fd-4a9e-8789-6194e02aa7dc.mov

:

packages/x-components/src/x-modules/url/store/actions/update-url.action.ts
```diff
@@ -1,61 +1,107 @@
...
- * @internal
+ * @public
  */
-export const updateUrl: UrlXStoreModule['actions']['updateUrl'] = ({
-  getters: { urlParams, urlMappedParamNames }
-}) => {
-  const url = new URL(window.location.href);
-  const oldParams = url.searchParams;
+export class UpdateUrlAction implements ActionsClass<UrlXStoreModule> {
+  /** Parameters that will require to push the browser state instead of replace. */
```
pmareke, line 12: Why this is a class variable? It's only used inside the `pushableParamsChanged`.

pmareke (changes requested):

packages/x-components/src/x-modules/url/events.types.ts
```diff
@@ -26,15 +26,14 @@ export interface UrlXEvents {
    * Document has finished loading.
    * * Payload: none.
```
tiborux, line 27: Update the payload value

packages/x-components/src/x-modules/url/store/actions/update-url.action.ts
```diff
@@ -1,61 +1,107 @@
...
+    });
   }
-};
 
+  /**
+   * Checks if the pushable params have changed.
+   *
+   * @param urlParams - The new Url params.
+   * @param oldValues - The old Url params.
+   * @internal
```
tiborux, line 86: move the internal tag after the return.

packages/x-components/src/x-modules/url/wiring.ts
```diff
@@ -12,6 +13,7 @@ import { createWiring } from '../../wiring/wiring.utils';
  * @internal
  */
 const wireCommit = namespacedWireCommit('url');
+const wireDispatch = namespacedWireDispatch('url');
```
tiborux, line 16: Document this

tiborux (changes requested):

packages/x-components/src/x-modules/url/store/actions/update-url.action.ts
guillei10: I followed the same approach we did with the `urlParams` getter. Moving this to a variable to make it more easy to find when we make changes if needed.

packages/x-components/src/x-modules/url/components/url-handler.vue
```diff
@@ -63,17 +57,24 @@
     }
 
     /**
-     * Emits the event with the state values when the document is loaded and open X if there is
-     * a query in the Url.
+     * Checks if the parameter query is in the url.
+     *
+     * @returns Boolean.
```
pmareke, line 62: @returns Boolean if the parameter query is in the url.

packages/x-components/src/x-modules/url/store/actions/update-url.action.ts
pmareke: The idea of having a variable is really good, what I meant was to declare inside the method `pushableParamsChanged` as it's the only place where it0s used.

pmareke (changes requested):

packages/x-components/src/x-modules/url/store/actions/update-url.action.ts
guillei10: This way you don't have to look for the variable into a function.

tiborux (dismissed):

pmareke (approved):

tiborux (approved):

PR #181, opened 2021-10-04
[EX-4750] use layout in mocked tests

mavmaf:
# Pull request template
The aim of this PR is to use the same view, `Layout.vue`, in all the mocked tests which are the following:
```
- identifier-results
- next-queries
- partial-results
- popular-searches
- query-suggestions
- recommendations
- related-tags
- search-box
```

Moreover, it was necessary to perform two minor changes in `history-queries `and `keyboard-navigation` tests in order not to fail while they are still using the old views.

## Motivation and context
The motivation for this change is to reduce the amount of views for the e2e tests

- `Main`

## How has this been tested?

All tests have been run independently as usual.

### DIFF
--- a/packages/x-components/src/views/Layout.vue
+++ b/packages/x-components/src/views/Layout.vue
@@ -273,9 +273,9 @@
             <template #default="{ partialResult }">
               <span data-test="partial-query">{{ partialResult.query }}</span>
               <BaseGrid :animation="resultsAnimation" :columns="4" :items="partialResult.results">
-                <template #Result="{ item: partialResult }">
+                <template #Result="{ item }">
                   <article class="result" style="max-width: 300px">
-                    <BaseResultImage :result="partialResult" class="x-picture--colored">
+                    <BaseResultImage :result="item" class="x-picture--colored">
                       <template #placeholder>
                         <div style="padding-top: 100%; background-color: lightgray"></div>
                       </template>
@@ -283,7 +283,7 @@
                         <div style="padding-top: 100%; background-color: lightsalmon"></div>
                       </template>
                     </BaseResultImage>
-                    <BaseResultLink :result="partialResult" class="x-result-link">
+                    <BaseResultLink :result="item" class="x-result-link">
                       <span class="x-result__title" data-test="partial-result-item">
                         {{ item.name }}
                       </span>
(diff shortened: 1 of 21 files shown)


packages/x-components/src/views/Layout.vue
```diff
@@ -2,6 +2,37 @@
   <div>
     <UrlHandler />
     <BaseEventsModalOpen>Start</BaseEventsModalOpen>
+    <h1>Test controls</h1>
+    <ul class="x-test-controls">
```
pmareke, line 6: `<ul class="x-test-controls x-list x-list--gap-05">`

packages/x-components/src/views/Layout.vue
```diff
@@ -2,6 +2,37 @@
...
+      <li class="x-test-controls__control">
```
pmareke, line 7: `<li class="x-test-controls__item x-list__item">`

packages/x-components/src/views/Layout.vue
```diff
@@ -2,6 +2,37 @@
...
+        <label>
```
pmareke, line 8: Use the label as the resto of the `li`

packages/x-components/src/views/Layout.vue
```diff
@@ -261,6 +317,34 @@
               </PromotedsList>
             </BannersList>
           </ResultsList>
+
```
pmareke, line 320: `<!-- Partials -->`

packages/x-components/src/views/Layout.vue
```diff
@@ -189,11 +229,27 @@
         </template>
 
         <template #main-body>
+          <!-- IdentifierResults -->
+          <div v-if="$x.identifierResults.length">
```
pmareke, line 233: Remove this div as the `IdentifierResults` will not render if there are not sku results.

packages/x-components/src/views/Layout.vue
```diff
@@ -189,11 +229,27 @@
...
+            <IdentifierResults class="x-list x-list--horizontal">
+              <template #default="{ identifierResult }">
+                <article class="result" style="max-width: 300px">
```
pmareke, line 236: Remove the max-width please.

packages/x-components/tests/e2e/cucumber/recommendations/recommendations.feature
```diff
@@ -5,9 +5,11 @@ Feature: Recommendations component
 
   Scenario Outline:  1. Recommendations are displayed
     Given following config: max items to store is <maxItemsToRequest>
+    And   start button is clicked
     Then  number of rows requested in "<request>" is <maxItemsToRequest>
     And   number of displayed recommendations are equal or less than <maxItemsToRequest>
-    And   each recommendation has an associated hyperlink containing image and text
+  # to include once recommendations and results have hyperlinks
```
pmareke, line 11: Indent these lines please.

packages/x-components/tests/e2e/cucumber/related-tags/related-tags.spec.ts
```diff
@@ -87,12 +87,12 @@ And(
     if (addToSearchBox) {
       cy.getByDataTest('search-input').should(
         'contain',
-        this.searchedQuery + this.clickedRelatedTag
+        this.searchedQuery + ' ' + this.clickedRelatedTag
```
pmareke, line 90: Please use string interpolation ``${this.searchedQuery} ${this.clickedRelatedTag}``

packages/x-components/tests/e2e/cucumber/related-tags/related-tags.spec.ts
```diff
@@ -87,12 +87,12 @@ And(
...
       );
     } else {
       cy.getByDataTest('search-input').should(
         'not.contain',
-        this.searchedQuery + this.clickedRelatedTag
+        this.searchedQuery + ' ' + this.clickedRelatedTag
```
pmareke, line 95: Same here

pmareke (changes requested):
As we talk, we need to add `retries: 2` in the cypress.json in order to avoid the first exception in the search box tests.

Great job by the way!

packages/x-components/src/views/Layout.vue
tiborux: el resto 🇲🇽

tiborux:
The build is failing

packages/x-components/src/views/Layout.vue
mavmaf: Well, that's the point and it makes sense for me. What I am missing?

pmareke: You don't need an `v-if` in the `div` as the component only renders if there are identifier results, I don't know if yo got me.

packages/x-components/cypress.json
```diff
@@ -8,11 +8,15 @@
   "pluginsFile": "tests/plugins/index.js",
   "supportFile": "tests/support/index.ts",
   "integrationFolder": "tests/e2e",
-  "testFiles": ["cucumber/**/*.features"],
+  "testFiles": [
```
pmareke, line 11: Are we sure about this change?

pmareke: Rollback this change please and leave it in one line.

packages/x-components/src/views/Layout.vue
```diff
@@ -261,6 +315,35 @@
               </PromotedsList>
             </BannersList>
           </ResultsList>
+
+          <!-- Partials -->
+          <PartialResultsList :animation="resultsAnimation">
+            <template #default="{ partialResult }">
+              <span data-test="partial-query">{{ partialResult.query }}</span>
+              <BaseGrid :animation="resultsAnimation" :columns="4" :items="partialResult.results">
+                <template #Result="{ item }">
```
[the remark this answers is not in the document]
pmareke: You're right, never mind.

PR #182, opened 2021-10-07
refactor(dynamic-slots): change all dynamic slots to kebab-case

pmareke:
In this PR I renamed all the dynamic slots to kebab-case (ie: Result -> result or NextQueriesGroup -> next-queries-group)

Also I added a new test in order to valide this change.

EX-4807

👀  @javieri-empathy @tajespasarela


tajespasarela (dismissed):

packages/x-components/src/components/__tests__/base-grid.spec.ts
```diff
@@ -1,18 +1,19 @@
-import { Identifiable } from '@empathyco/x-types';
 import { mount, Wrapper, WrapperArray } from '@vue/test-utils';
```
javieri-empathy, line 1: In the `selected-filters-list.vue` component, we are also using the `filter.facetId` as the slot name. For consistency's shake, we should change it. 

However, as this is a property that the developer users of the library can modify through the adapter, and we haven't yet set any standards about how a facet id should be for us, I'm not sure if we should change it before deciding this.

javieri-empathy: Same happens in the `facets.vue` component

packages/x-components/src/components/__tests__/base-grid.spec.ts
```diff
@@ -53,8 +54,20 @@ function renderBaseGridComponent({
 
 describe('testing Base Grid', () => {
   const searchResponse = getSearchResponseStub();
+  const nextQueriesResponse = [
```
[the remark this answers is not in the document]
pmareke: Sure. I'm going to duplicate this interface in the spec file and add a TODO in order to remove it when the original one is exported.

packages/x-components/src/components/base-grid.vue
```diff
@@ -141,7 +141,10 @@
       cssClass: VueCSSClasses;
     }[] {
       return this.computedItems.map(item => ({
-        item,
+        item: {
+          ...item,
+          modelName: toKebabCase(item.modelName)
+        },
         cssClass: item.modelName
```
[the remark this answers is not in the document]
pmareke: Sure.

packages/x-components/src/components/__tests__/base-grid.spec.ts
javieri-empathy: I would take a decision in this PR, if not this is the classic change that we will never do. Either we force to write kebab case Id's for the facets, or we transform them just for these components.

packages/x-components/src/components/__tests__/base-grid.spec.ts
```diff
@@ -53,8 +54,20 @@ function renderBaseGridComponent({
...
+    {
+      modelName: 'NextQueries',
+      id: 'nextQueries',
+      nextQueries: getNextQueriesStub()
+    }
+  ];
   const columns = 3;
```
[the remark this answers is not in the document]
javieri-empathy: I would just remove it from every test it does not use it. This will hurt nobody, and it will leave the test simpler to read.

packages/x-components/src/utils/types.ts
```diff
@@ -163,6 +163,17 @@ export type EventsForDirectionLimit = {
  */
 export type ListItem = Identifiable & NamedModel;
 
+/**
+ * The type returned by the {@link gridItems} function. Basically it's a list of items with its CSS classes and a slotName
```
javieri-empathy, line 167: I would just leave this grid item as a private type for the base grid component. It should not be part from the public API.

packages/x-components/src/components/__tests__/base-grid.spec.ts
```diff
@@ -101,3 +131,9 @@ interface BaseGridComponentAPI {
   getDefaultSlot: () => WrapperArray<Vue>;
   getScopedSlot: (modelName: string) => WrapperArray<Vue>;
 }
+
+// TODO Remove this interface when the original from the next-queries.list.vue is exported.
+interface NextQueriesGroup extends ListItem {
```
javieri-empathy, line 136: Contrary to the `GridItem` type which should be private (just for the file it is being used in), this one should be public, so you can just move it to the utils file or the search-types project instead of duplicating it.

packages/x-components/src/utils/types.ts
javieri-empathy: Moreover the link does not work here, and causes the build to fail.

javieri-empathy (changes requested):

packages/x-components/src/utils/types.ts
```diff
@@ -197,3 +197,14 @@ export interface DebounceOptions {
...
   (...args: Params): void;
 }
+
+/**
+ * A list of next queries.
+ *
+ * @public
+ */
+// TODO Remove this interface when the original from the next-queries.list.vue is exported.
+export interface NextQueriesGroup extends ListItem {
```
tajespasarela, line 207: Why duplicate it? why not export the type in `NextQueriesList` component and use it?

packages/x-components/src/components/base-grid.vue
```diff
@@ -37,6 +37,18 @@
   import { XInject } from './decorators/injection.decorators';
   import { LIST_ITEMS_KEY } from './decorators/injection.consts';
 
+
+  /**
+   * The type returned by the gridItems function. Basically it's a list of items with its CSS classes and a slotName
```
tajespasarela: format doc line, it is too long

packages/x-components/src/components/__tests__/base-grid.spec.ts
tajespasarela: lets do it! kebabCase also with the `facetId`.

tajespasarela (changes requested):

packages/x-components/src/utils/types.ts
pmareke: We need to do it in another PR, release and alpha an import it using the library, right?.

packages/x-components/src/components/__tests__/base-grid.spec.ts
```diff
@@ -68,30 +76,48 @@ describe('testing Base Grid', () => {
...
     expect(getScopedSlot('banner').exists()).toBe(true);
   });
+
+  it('allows customizing named slots only using kebab case', () => {
+    const customValidItemSlot = `
+      <template #banner="{ item }">
+        <p data-test="banner-slot">{{ item.modelName }}</p>
+      </template>
+      <template #NextQueriesGroup="{ item }">
+        <p data-test="next-queries-group-slot">{{ item.modelName }}</p>
```
javieri-empathy, line 103: The data test for this should be `NextQueriesGroup-slot`. 

Below you are using a helper function that just appends `-slot` to the passed model name, and uses to select a DOM element. This means that when you do `getScopedSlot('NextQueriesGroup')` you are running `wrapper.findAll('[data-test="NextQueriesGroup-slot"]'))`, but in your render template there is no element with this data test that could match, therefore the test will always pass, and the assertion worthless.

Additionally, if you just inline the template into the `renderBaseGridComponent` call you would make me very happy. Why create variables that are only going to be used to be passed into a function below? Just inline them in the call, it is much easier.

packages/x-components/src/utils/types.ts
javieri-empathy: Why???, they are both types from the same package, even if this type belonged to another package of the mono-repo it could be done in this PR, that's the whole point of the mono-repo.

packages/x-components/src/x-modules/facets/components/facets/facets.vue
```diff
@@ -39,6 +39,7 @@
   import { Component, Prop, Vue } from 'vue-property-decorator';
   import { Getter } from '../../../../components/decorators/store.decorators';
   import { xComponentMixin } from '../../../../components/x-component.mixin';
+  import { toKebabCase } from '../../../../utils';
```
javieri-empathy, line 42: Import directly from the file of the function, let's try to avoid barrels inside our code please.

packages/x-components/src/x-modules/facets/components/facets/facets.vue
```diff
@@ -18,9 +18,9 @@
             @binding {Facet} facet - Facet to render
       -->
       <slot
-        v-if="$scopedSlots[facetId]"
+        v-if="$scopedSlots[slotName(facetId)]"
         v-bind="{ facet, selectedFilters: selectedFiltersByFacet[facetId] || [] }"
-        :name="facetId"
+        :name="slotName(facetId)"
```
javieri-empathy, line 23: Why not just map the items like you did previously, and avoid calling twice the  wrongly named `slotName` method from inside the render function?

packages/x-components/src/x-modules/facets/components/lists/__tests__/selected-filters-list.spec.ts
```diff
@@ -100,19 +104,24 @@ describe('testing SelectedFiltersList component', () => {
         <SelectedFiltersList>
           <template #default="{ filter }">{{ filter.label }} selected!</template>
           <template #brand="{ filter }">Which one? {{ filter.label }}</template>
+          <template #root-categories="{ filter }">Much better: {{ filter.label }}</template>
+          <template #rootCategories="{ filter }">Much better: {{ filter.label }}</template>
```
javieri-empathy, line 108: What are you testing with this? I see no assertions related with this line. It is exactly the same than the one above, so you will never know if it is the `root-categories` slot or the `rootCategories` slot the one it is being rendered.

 If you change the text of the `rootCategories` slot to something like `Shouldn't be rendered: {{ filter.label }}`, it is enough, because you are checking that exactly 3 filters are rendered with a different text each one.

packages/x-components/src/x-modules/facets/components/lists/selected-filters-list.vue
```diff
@@ -41,6 +41,7 @@
   import { Component, Prop, Vue } from 'vue-property-decorator';
   import { Facet } from '@empathyco/x-types';
   import { xComponentMixin } from '../../../../components/index';
+  import { toKebabCase } from '../../../../utils';
```
javieri-empathy, line 44: Import directly from the file instead of from the barrel. And please, fix also the above `/components/index` import. 
We should try avoid importing from barrels in our code. It is very easy to create cyclical dependencies this way, and even if we don't do so, we are saying that the importer depends on all the files exported by the barrel, which is not true.

packages/x-components/src/x-modules/facets/components/lists/selected-filters-list.vue
```diff
@@ -69,6 +70,19 @@
...
+    /**
+     * Kebab trans formation of the facet id.
+     *
+     * @param facetId - The id of the facet to transform in kebab case.
+     *
+     * @returns A kebab case id.
+     *
+     * @internal
+     */
+    protected slotName(facetId: string): string {
```
javieri-empathy, line 82: Same as the facets, why not just map the item like in every component. In any case remember that this is not a valid method name for our code style guidelines, and the documentation is also wrong spelled. It is also usually better to document in an active voice: `Transforms the facet id to kebab case`, rather than `kebab transformation of the facet id`, which sounds weird when reading it.

packages/x-components/src/components/base-grid.vue
```diff
@@ -37,6 +37,18 @@
   import { XInject } from './decorators/injection.decorators';
   import { LIST_ITEMS_KEY } from './decorators/injection.consts';
 
+  /**
+   * The type returned by the gridItems function. Basically it's a list of items with its CSS
+   * classes and a slotName.
+   *
+   * @public
```
javieri-empathy, line 44: When not exported, the public/internal/private decorators are worthless, they won't be used by API extractor. However, I think it still is a good practise to explicitly say this way if we intend something to be only for internal developers. So either remove this `public`, or change it to `internal`, whatever you prever.

packages/x-components/src/utils/types.ts
tajespasarela: hack-khaby

packages/x-components/src/components/base-grid.vue
pmareke: My fault, it's internal.

packages/x-components/src/components/__tests__/base-grid.spec.ts
pmareke: I left :
```html
   <template #NextQueriesGroup="{ item }">
        <p data-test="next-queries-group-slot">{{ item.modelName }}</p>
      </template>
      <template #next-queries-group="{ item }">
        <p data-test="next-queries-group-slot">{{ item.modelName }}</p>
      </template>
```

to demonstrate that an invalid slot is not render and the kebab case it's, it's ok if the `getScopedSlot` doesn't found it.

packages/x-components/src/x-modules/facets/components/lists/selected-filters.vue
```diff
@@ -62,6 +65,20 @@
...
+     * Transform a dictionary of Facets including the slot name.
+     *
+     * @returns A dictionary of facets with the slot name.
+     *
+     * @internal
+     */
+    protected get mappedSelectedFilters(): Filter[] {
+      return this.selectedFilters.map((filter: any) => ({
+        ...filter,
+        slotName: toKebabCase(filter['facetId'] ?? '')
```
javieri-empathy, line 78: Nice try casting filter to `any` type, but no one scapes the types police. As not every filter has to belong to a facet, you can check it with `isFacetFilter` function, and if it does not belong to a facet, use the `default` slot.

As this component is used by the `selected-filters-list` component, I won't complain about using a slightly different variation, but please, add the `slotName` property to an interface and use it as the return type here.

packages/x-components/src/x-modules/facets/components/lists/selected-filters.vue
```diff
@@ -62,6 +65,20 @@
     @Getter('facets', 'selectedFiltersByFacet')
     public selectedFiltersByFacet!: FiltersByFacet;
 
+    /**
+     * Transform a dictionary of Facets including the slot name.
```
javieri-empathy, line 69: Fix doc. This adds the slot name property to a filter, not a facet.

packages/x-components/src/x-modules/facets/components/facets/facets.vue
```diff
@@ -105,6 +106,20 @@
...
+     * Transform a dictionary of Facets including the slot name.
+     *
+     * @returns A dictionary of facets with the slot name.
+     *
+     * @internal
+     */
+    protected get mappedFacets(): Dictionary<Facet> {
+      return map(this.facetsToRender, (_, facet) => ({
+        ...facet,
+        slotName: toKebabCase(facet.id as string)
```
javieri-empathy, line 119: Follow the same pattern you did in other components, wrapping the facet into an object, and create an interface for the type. Facet type do not have `slotName`. It might make sense to extract this pattern to a utility type.

packages/x-components/src/utils/types.ts
```diff
@@ -197,3 +197,21 @@ export interface DebounceOptions {
 export interface ThrottleFunction<Params extends any[]> {
   (...args: Params): void;
 }
+
+/**
+ * Custom interface to provide a slot name to a Filter.
+ *
+ * @public
+ */
+export interface RenderFilter extends Filter {
```
javieri-empathy, line 206: I don't think this is the best place to have this interface. Also, unless we need them, they should be internal. I would just add them in the components you use them

javieri-empathy: Like we said, now I don't see any reason to not wrap the filter and facet into an object, instead of extending from them.

packages/x-components/src/x-modules/facets/components/lists/selected-filters-list.vue
```diff
@@ -88,6 +90,20 @@
...
+
+    /**
+     * Transform a dictionary of Filters including the slot name.
+     *
+     * @returns A dictionary of facets with the slot name.
+     *
+     * @internal
+     */
+    protected mapSlot(selectedFilters: Filter[]): RenderFilter[] {
+      return selectedFilters.map((filter: Filter) => ({
```
javieri-empathy, line 102: You don't need this `Filter` type, it is inferred.

packages/x-components/src/x-modules/facets/components/lists/selected-filters-list.vue
```diff
@@ -88,6 +90,20 @@
...
+        ...filter,
+        slotName: isFacetFilter(filter) ? toKebabCase(filter['facetId'] as string) :' default'
```
javieri-empathy, line 104: Unless you need it (kebab case properties, or dynamic properties for example), avoid using bracket syntax to access properties. It adds noise.

javieri-empathy (changes requested):

javieri-empathy (approved):

packages/x-components/src/x-modules/facets/components/facets/facets.vue
```diff
@@ -107,16 +126,16 @@
...
     /**
-     * Transform a dictionary of Facets including the slot name.
+     * Transforms a dictionary of Facets including the slot name.
      *
      * @returns A dictionary of facets with the slot name.
      *
      * @internal
      */
-    protected get mappedFacets(): Dictionary<Facet> {
+    protected get mappedFacets(): Dictionary<RenderFacet> {
```
tajespasarela, line 135: you already have the id as string in the dictionary:

```
      return map(this.facetsToRender, (facetId, facet) => ({
        slotName: toKebabCase(facetId),
        facet
      }));
```

javieri-empathy (approved):

PR #183, opened 2021-10-10
feat(url): syncronize url and state with the scroll 

tiborux:
**How to test it**

Scroll the results and check how the url updates its value with the first item in the scroll viewport. If you reload the webpage with a scroll in the url, the scroll element will scroll to the element that its set on the url.

### DIFF
--- a/packages/x-components/src/components/scroll/scroll.mixin.ts
+++ b/packages/x-components/src/components/scroll/scroll.mixin.ts
@@ -32,6 +32,14 @@ export default class ScrollMixin extends Vue {
   public distanceToBottom!: number;
 
   /**
+   * Percent of the element that has to be outside of the scroll viewport.
+   *
+   * @public
+   */
+  @Prop({ default: 0.25 })
+  public hiddenPercentOfFirstItem!: number;
+
+  /**
    * Property for getting the current position of scroll.
    *
    * @internal
@@ -138,7 +146,7 @@ export default class ScrollMixin extends Vue {
   emitScroll(_newScrollPosition: number, oldScrollPosition: number): void {
     this.$emit('scroll', this.currentPosition);
     this.previousPosition = oldScrollPosition;
-    this.setFirstGridItemIdInView();
+    this.setFirstItemInScrollView();
   }
 
   /**
@@ -192,12 +200,12 @@ export default class ScrollMixin extends Vue {
   }
 
   /**
-   * Emits 'FirstScrollItemChanged' event with the id of the first element that it is inside the
-   * scroll element.
+   * Emits 'FirstItemInScrollViewChanged' event with the id of the first element that it is inside
+   * the scroll element.
    *
    * @internal
    */
-  protected setFirstGridItemIdInView(): void {
+  protected setFirstItemInScrollView(): void {
     if (this.$el) {
       const scrollTargets = this.$el.querySelectorAll<HTMLElement>('[data-scroll-id]');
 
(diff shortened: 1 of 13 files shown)


packages/x-components/src/components/scroll/scroll.mixin.ts
```diff
@@ -186,4 +218,63 @@ export default class ScrollMixin extends Vue {
...
   }
+
+  /**
+   * Emits 'FirstItemInScrollViewChanged' event with the id of the first element that it is inside
+   * the scroll element.
+   *
+   * @internal
+   */
+  protected setFirstItemInScrollView(): void {
+    if (this.$el) {
```
pmareke, line 229: You have a problem here with `this.$el` as in the base main scroll test we're testing the scroll in the `html` and the `body`.

With this change everything works again:

```javascript
 /**
   * Emits 'FirstItemInScrollViewChanged' event with the id of the first element that it is inside
   * the scroll element.
   *
   * @internal
   */
  protected setFirstItemInScrollView(): void {
    let element: HTMLElement;
    if (this.$parent.$props['tag'] === 'html') {
      element = document.documentElement;
    } else if (this.$parent.$props['tag'] === 'body') {
      element = document.body;
    } else {
      element = this.$el as HTMLElement;
    }
    if (element) {
      const scrollTargets = element.querySelectorAll<HTMLElement>('[data-scroll-id]');

      const firstElementInView = Array.from(scrollTargets).find(element =>
        this.isAnyPartOfElementInViewport(element)
      );
      if (firstElementInView) {
        this.$x.emit('FirstItemInScrollViewChanged', firstElementInView.dataset.scrollId as string);
      }
    }
  }
```

But I'm not sure if it's the best solution, any idea?

packages/x-components/tests/e2e/cucumber/scroll/scroll.feature
```diff
@@ -0,0 +1,17 @@
+Feature: Scroll component
```
pmareke, line 1: My proposal for the missing Scenario:

```

Feature: Scroll component

  Background:
    Given a results scroll API with a known response
    Given a related tags API with a known response
    Given a next queries API with a known response
    Given a recommendations API with a known response
    Given a query suggestions API with a known response

  Scenario Outline: 1. The first element in the scroll viewport is set in the URL
    Given following basic config
    And start button is clicked
    When  "<query>" is searched
    Then  scroll down the results
    And   the first element id in the scroll viewport is in the url
    Examples:
    | query |
    | lego  |

  Scenario Outline: 2. The scroll url param is read and scroll into
    Given an url with scroll parameter
    Then  the first element in the view is the scroll <id>
    Examples:
    | id |
    | 17 |
```

packages/x-components/tests/e2e/cucumber/scroll/scroll.spect.ts
```diff
@@ -0,0 +1,26 @@
+import { Given, Then, And } from 'cypress-cucumber-preprocessor/steps';
```
pmareke, line 1: My proposal for the missing Scenario

```
import { Given, Then, And } from 'cypress-cucumber-preprocessor/steps';
import { getResultsStub } from '../../../../src/__stubs__';

// Background
Given('a results scroll API with a known response', () => {
  cy.intercept('https://api.empathy.co/search', req => {
    req.reply({
      results: getResultsStub(48),
      banners: [],
      promoteds: []
    });
  }).as('interceptedResults');
});

// Scenario 1
Given('following basic config', () => {
  cy.visit('/layout?useMockedAdapter=true', {});
});

Given('an url with scroll parameter', () => {
  cy.visit('/layout?useMockedAdapter=true&query=lego&page=4&scroll=product-25', {});
});

Then('scroll down the results', () => {
  cy.get('[data-scroll-id=product-32]').scrollIntoView();
});

And('the first element id in the scroll viewport is in the url', () => {
  cy.url().should('include', 'product-32');
});

And('the first element in the view is the scroll {int}', (id: number) => {
  cy.get(`[data-scroll-id=product-${id}]`)
    .should('be.visible')
    .then(() => {
      cy.get(`[data-scroll-id=product-${id -1}]`).should('not.be.visible');
    });
});
```

PR #184, opened 2021-10-12
feat(related-tags): Change the related tags mapper in search adapter

guillei10:
## Description

The aim of this PR is to change the mapping of the query of the related tags. Until now we just append all the tags on the right of the query. Now we are supporting tags to go on the left OR right of the query. To avoid some nonsense queries, we have changed the mapper.

The algorithm is based on @PabloCasia's [in this file.](https://github.com/empathyco/x-android/blob/main/adapter/src/main/java/co/empathy/x/adapter/network/ApiQueryBuilder.kt)

## Type of change

- [  ] Bug fix (non-breaking change that fixes an issue)
- [  ] Breaking change (fix or feature that causes existing functionality to not work as expected)
- [  ] Change requires a documentation update

### DIFF
--- a/packages/search-adapter/src/empathy/mappers/request/params/empathy-request-related-tags-query.mapper.ts
+++ b/packages/search-adapter/src/empathy/mappers/request/params/empathy-request-related-tags-query.mapper.ts
@@ -1,4 +1,3 @@
-import { RelatedTag } from '@empathyco/x-types';
 import { injectable } from 'inversify';
 import { QueryableRequest } from '../../../../types';
 import { RequestMapper } from '../../../empathy-adapter.types';
@@ -9,24 +8,16 @@ import { RequestMapper } from '../../../empathy-adapter.types';
  * @public
  */
 @injectable()
-export class EmpathyRequestRelatedTagsQueryMapper
-  implements RequestMapper<QueryableRequest, string>
+export class EmpathyRequestRelatedTagsQueryMapper implements RequestMapper<QueryableRequest, string>
 {
   map({ relatedTags = [] }: QueryableRequest, query: string): string {
-    const [leftRelatedTags, rightRelatedTags] = this.splitRelatedTagsByQueryPosition(relatedTags);
-    return `${leftRelatedTags} ${query} ${rightRelatedTags}`.trim();
-  }
-
-  splitRelatedTagsByQueryPosition(relatedTags: RelatedTag[]): string[] {
-    return relatedTags
-      .reduce(
-        ([left, right], { tag, query }) => {
-          return query.startsWith(tag)
-            ? [`${left} ${tag}`, right]
-            : [left, `${right} ${tag}`];
-        },
-        ['', '']
-      )
-      .map(sortedRelatedTags => sortedRelatedTags.trim());
+    return relatedTags.reduce(
+      (partialQuery, { tag, query: relatedTagQuery }) =>
+        relatedTagQuery.startsWith(tag)
+          ? `${tag} ${partialQuery}`
+          : `${partialQuery} ${tag}`,
+      query
+    )
+    .trim();
   }
 }


packages/search-adapter/src/empathy/mappers/request/params/empathy-request-related-tags-query.mapper.ts
```diff
@@ -8,9 +9,24 @@ import { RequestMapper } from '../../../empathy-adapter.types';
...
+export class EmpathyRequestRelatedTagsQueryMapper
+  implements RequestMapper<QueryableRequest, string>
+{
   map({ relatedTags = [] }: QueryableRequest, query: string): string {
-    return relatedTags.reduce((chain, rt) => `${chain} ${rt.tag}`, query).trim();
+    const [leftRts, rightRts] = this.splitRelatedTagsByQueryPosition(relatedTags, query);
+    return `${leftRts} ${query} ${rightRts}`.trim();
+  }
+
+  splitRelatedTagsByQueryPosition(array: RelatedTag[], query: string): string[] {
```
pmareke, line 20: `relatedTags` instead of `array` as it's to generic.

packages/search-adapter/src/empathy/mappers/request/params/empathy-request-related-tags-query.mapper.ts
```diff
@@ -8,9 +9,24 @@ import { RequestMapper } from '../../../empathy-adapter.types';
...
  */
 @injectable()
-export class EmpathyRequestRelatedTagsQueryMapper implements RequestMapper<QueryableRequest, string> {
-
+export class EmpathyRequestRelatedTagsQueryMapper
+  implements RequestMapper<QueryableRequest, string>
+{
   map({ relatedTags = [] }: QueryableRequest, query: string): string {
-    return relatedTags.reduce((chain, rt) => `${chain} ${rt.tag}`, query).trim();
+    const [leftRts, rightRts] = this.splitRelatedTagsByQueryPosition(relatedTags, query);
```
pmareke, line 16: There is a warning here because the IDE is complaining about the possible undefined of the `relatedTags`.

It's not a big deal but you can remove if adding

```javascript
  if (!relatedTags ){
      return query;
    }
```

pmareke (changes requested):

packages/search-adapter/src/empathy/mappers/request/params/empathy-request-related-tags-query.mapper.ts
```diff
@@ -13,12 +13,18 @@ export class EmpathyRequestRelatedTagsQueryMapper
   implements RequestMapper<QueryableRequest, string>
 {
   map({ relatedTags = [] }: QueryableRequest, query: string): string {
-    const [leftRts, rightRts] = this.splitRelatedTagsByQueryPosition(relatedTags, query);
+    const [leftRts, rightRts] = this.splitRelatedTagsByQueryPosition(
```
pmareke, line 16: I will use `leftRelatedTags` and `rightRelatedTags` instead.

packages/search-adapter/src/empathy/mappers/request/params/empathy-request-related-tags-query.mapper.ts
```diff
@@ -8,9 +9,30 @@ import { RequestMapper } from '../../../empathy-adapter.types';
...
+    return `${leftRts} ${query} ${rightRts}`.trim();
+  }
+
+  splitRelatedTagsByQueryPosition(
+    relatedTags: RelatedTag[],
+    query: string
+  ): string[] {
+    return relatedTags
+      .reduce(
+        ([left, right], rt) => {
```
javieri-empathy, line 29: Avoid using abbreviations as variable names `rt`, `rts` are not valid names following our naming conventions.

packages/search-adapter/src/empathy/mappers/request/params/empathy-request-related-tags-query.mapper.ts
```diff
@@ -8,9 +9,30 @@ import { RequestMapper } from '../../../empathy-adapter.types';
...
+          return rt.query.indexOf(query) > rt.query.indexOf(rt.tag)
```
javieri-empathy, line 30: Should we extract this to a method? `isQueryAfterTag` just to document what it does? I have to think for a few seconds, even minutes every time I read this code 😅

javieri-empathy: I think we should normalise this query, or as a minimum, making it lowercase. In phones, the keyboard tends to auto-capitalises the typed text usually, and we would end up comparing non matching strings. You can try changing `lego` to `lEgo` in one of your tests to see how it fails.

Another option would be to just use each `relatedTag`, `tag` and `query` properties to determine if is a tag that should go to the left or to the right. This might be a simpler approach with a simpler condition, but I'm not sure if it is a 100% bulletproof.

javieri-empathy (changes requested):

packages/search-adapter/src/empathy/__tests__/request-mappers/param-mappers/empathy-request-related-tags-query-mapper.spec.ts
```diff
@@ -4,26 +4,46 @@ import { EmpathyRequestRelatedTagsQueryMapper } from '../../../mappers/request/p
...
-    { modelName: 'RelatedTag', tag: 'city', previous: 'lego', query: 'lego city', selected: true },
-    { modelName: 'RelatedTag', tag: 'friends', previous: 'lego', query: 'lego friends', selected: true },
-    { modelName: 'RelatedTag', tag: 'playmobil farm', previous: 'lego', query: 'lego playmobil farm', selected: true }
-  ];
-  const rawRequest: QueryableRequest = {
-    query: query,
-    relatedTags: relatedTags
-  };
+describe('Related tags query mapper', () => {
+  it('Appends related tags to both sides of query', () => {
```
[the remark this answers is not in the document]
javieri-empathy: There are still trailing commas in this file after each related tag last property.

packages/search-adapter/src/empathy/__tests__/request-mappers/param-mappers/empathy-request-related-tags-query-mapper.spec.ts
```diff
@@ -4,26 +4,46 @@ import { EmpathyRequestRelatedTagsQueryMapper } from '../../../mappers/request/p
...
+    const query = 'lego';
```
javieri-empathy, line 9: Capitalise the query (`Lego`), to check that this new behavior still works.

packages/search-adapter/src/empathy/mappers/request/params/empathy-request-related-tags-query.mapper.ts
```diff
@@ -8,9 +9,29 @@ import { RequestMapper } from '../../../empathy-adapter.types';
...
+            : [left, `${right} ${tag}`];
+        },
+        ['', '']
+      )
+      .map(sortedRelatedTags => sortedRelatedTags.trim());
+  }
+
+  // Checks if the query is after the tag.
+  isTagBeforeQuery(tag: string, query: string): boolean {
+    return query.indexOf(tag) === 0;
```
javieri-empathy, line 35: I just realised that you have the `startsWith` method, which is more suited for this change.

javieri-empathy (changes requested):

packages/search-adapter/src/empathy/mappers/request/params/empathy-request-related-tags-query.mapper.ts
```diff
@@ -8,9 +9,29 @@ import { RequestMapper } from '../../../empathy-adapter.types';
  * @public
  */
 @injectable()
-export class EmpathyRequestRelatedTagsQueryMapper implements RequestMapper<QueryableRequest, string> {
-
+export class EmpathyRequestRelatedTagsQueryMapper
+  implements RequestMapper<QueryableRequest, string>
+{
   map({ relatedTags = [] }: QueryableRequest, query: string): string {
```
javieri-empathy, line 15: I think we could even simplify this a little more. I'm not 100% sure again, so give me some time to see if I can break it, but here is the proposal:

```
  map({ relatedTags = [] }: QueryableRequest, query: string): string {
    return relatedTags.reduce(
      (partialQuery, relatedTag) =>
        relatedTag.query.startsWith(relatedTag.tag)
          ? `${ relatedTag.tag } ${ partialQuery }`
          : `${ partialQuery } ${ relatedTag.tag }`,
      query
    ).trim();
  }
```

pmareke (approved):

PR #191, opened 2021-10-18
Add new layout into XComponents

pmareke:
EX-4821

### DIFF
--- a/packages/x-components/src/components/layouts/single-column-layout.vue
+++ b/packages/x-components/src/components/layouts/single-column-layout.vue
@@ -158,7 +158,9 @@
       z-index: 3;
 
       ::v-deep .x-modal__content {
-        margin-inline-start: var(-x-size-margin-left-layout-single-column, 0);
+        width: 100%;
+        height: 100%;
+        margin-inline-start: var(--x-size-margin-left-layout-single-column, 0);
       }
     }
     &__predictive,
--- a/packages/x-components/src/views/layouts/multi-column-layout.vue
+++ b/packages/x-components/src/views/layouts/multi-column-layout.vue
@@ -3,7 +3,7 @@
     <template #header-start>
       <span
         class="
-          x-title1
+          x-title
           x-list__item--expand
           x-list x-list--horizontal x-list--justify-center
           x-margin--bottom-03 x-margin--top-03
@@ -27,7 +27,7 @@
     <template #header-end>
       <span
         class="
-          x-title1
+          x-title2
           x-list__item--expand
           x-list x-list--horizontal x-list--justify-center
           x-margin--bottom-03 x-margin--top-03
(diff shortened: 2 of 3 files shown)


packages/x-components/src/components/layouts/fixed-header-and-asides-layout.vue
```diff
@@ -0,0 +1,289 @@
...
+<script lang="ts">
+  import { Component } from 'vue-property-decorator';
+  import { mixins } from 'vue-class-component';
+  import { TranslateFromLeft, TranslateFromRight } from '../animations';
+  import BaseIdModal from '../modals/base-id-modal.vue';
+  import BaseIdScroll from '../scroll/base-id-scroll.vue';
+  import LayoutsMixin from './layouts.mixin';
+
+  /*
+ <docs lang="mdx">
```
tiborux, line 91: Why this is commented an inside of the script tag?

packages/x-components/src/router.ts
```diff
@@ -9,6 +9,11 @@ const routes: RouteConfig[] = [
     name: 'Layout',
     component: () => import('./views/Layout.vue')
   },
+  {
+    path: '/jumbo',
+    name: 'Jumbo',
```
tiborux, line 14: Name it something like Fixed header layout?

tiborux (changes requested):

packages/x-components/src/components/layouts/multi-column-max-width-layout.vue
```diff
@@ -192,14 +168,6 @@
...
       [main-end scroll-to-top-end page-end];
 
-    // color
-    background-color: var(--x-color-background-layout-columns, white);
-    border-color: var(--x-color-border-layout-columns);
-
-    // border
-    border-style: solid;
-    border-width: var(--x-size-border-width-layout-columns, 0);
-
```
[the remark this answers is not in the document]
tajespasarela: Add documentation of the public tokens used in this layout, with what is the purpose of each one, and its default values.

Keep in mind that internal tokens like these calculated are not necessary to be documented:
image

packages/x-components/src/components/layouts/single-column-layout.vue
```diff
@@ -68,11 +68,13 @@
```
tajespasarela, line 68: move the tokens form `packages/x-components/src/design-system/components/layout/columns.tokens.scss`
to here, and remove those that are not necessary as it can be done by the slots content.

tajespasarela: Also document them with the use and default value

packages/x-components/src/router.ts
tajespasarela: +1

packages/x-components/src/views/fixed-header-and-asides-layout-demo.vue
```diff
@@ -0,0 +1,41 @@
+<template>
```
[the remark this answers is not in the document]
pmareke: Should we? Or a I'll delete the view.

PR #192, opened 2021-10-18
docs: Fix components documentation export

javieri-empathy:
Exports the documentation of the components under `src/components`, that were not in a nested folder. To reproduce the issue, just trigger a build from the `main` branch, and see how components like the `src/components/base-dropdown` have no exported documentation.

The regular expressions used for the components have been tweaked, a little bit to ensure that they match these missing components. Additionally, named capture groups have been used, but just to provide better readability of the code.

### DIFF
--- a/packages/x-components/build/docgen/utils.js
+++ b/packages/x-components/build/docgen/utils.js
@@ -52,7 +52,6 @@ function generateDestination(folder, regex, file) {
   const match = regex.exec(file);
   if (match) {
     const { path, componentName } = match.groups;
-    console.log(join(folder, path, `x-components.${componentName}.md`));
     return join(folder, path, `x-components.${componentName}.md`);
   } else {
     return '';


packages/x-components/build/docgen/utils.js
```diff
@@ -50,7 +50,7 @@ function getDocumentFileDestination(file, config) {
 function generateDestination(folder, regex, file) {
   const match = regex.exec(file);
   if (match) {
-    const [, path, componentName] = match;
+    const { path, componentName } = match.groups;
```
pmareke, line 53: Even though this change works perfectly I realised that if the component is the root of the `components` directory, the path is not matched:

image

Apparently it should fails as the method should returns:
```
return `${folder}//x-components.${componentName}.md`;
```
But it works.

pmareke (changes requested):
Review regex

pmareke (approved):

PR #193, opened 2021-10-18
Bugfix/ex 4878 fix components markdown

javieri-empathy:
Makes all components use `lang="mdx"` in its docs, so prettier can format it. 
Fixes some `#Example` headers in docs, which should be `# Example` (note the space between `#` and the text).
Fixes an issue where eslint was trying to lint `.md` files, which we don't support in our rules.

### DIFF
--- a/.github/contributing/components.md
+++ b/.github/contributing/components.md
@@ -2,8 +2,9 @@
 
 ## Naming components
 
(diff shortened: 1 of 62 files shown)


pmareke (changes requested):
There are 4 components using `Example` instead of `Examples`

image
image

PR #195, opened 2021-10-20
feat(extra-params): Integrate UrlHandler with the Extra Params.

tajespasarela:
[EX-4564]

This PR is to make work the URL module and the Extra params. Changes:

- Store Emitters:
    All the emitters have now a debounce (with 0ms). This means that even if you make several changes to the store asynchronously, the Event of the emitter will be emitted only once. Obviously, if the changes are made in different moments and not sequentially (even asynchronously), it will emit multiple Events.
    Furthermore, the `XxxRequestChanged` event Emitters are treated a little different from the rest. The Emitters of this kind of event will have an "extra" debounce (this means they will have two denounces with 0 ms). What for? well, the idea is to "sort" the events moving all the `XxxRequestChanged` to the end of the JS tasks queue. This way the rest of the Events, which are making changes in the store state, are executed before the request event is emitted. So we will "minimize" the repeated `XxxRequestChanged` events, so we will "minimize" the repeated requests to API. It is not a perfect solution and requests like "PopularSearches" and "Recommendations" are still repeated because they are requested as soon as the "app" is mounted.
   Finally, these `XxxRequestChanged` Emitters, will have a default filter to check if the values of the request have really changed. Once again to minimize the repeated requests. It only checks at the first level of the object, so it will not check inner Objects nor Arrays, to avoid heavy CPU tasks in complex requests (Filters for example).

- XInstaller:
    Removed the extra params from the Adapter config, because these parameters are already sent in the Request Object.

- Url Module:
  Unified extra params and initial params in the state. The mutation uses `Vue.set` in all the cases to deal with new fields (extra params) in the state.

- ExtraParams component:
    Emit initial `ExtraParamsProvided` with the already existing params in the store, to deal with the case these params were already there because the params loaded from URL.
    The next `ExtraParamsProvided` are emitted with the changes in the `values` props.

- RenderlessExtraParams:
 Removed default value. This default value can be configured through the `ExtraParams` component.

- SnippetConfigExtraParams:
    Changed the @XInject by default Vue injection, because the XInstaller injects it in the default way and the injected object it is already reactive.

- UrlHandler component:
    Removed `load` listener. The event is emitted as soon as the component is created. The URL is supposed to be loaded in this case.
    Emits new `ExtraParamsLoadedFromUrl` again.

- Bonus:
  Added some wiring for the URL params load to some XModules were missing.
  Added some filters to Emitters comparing Arrays to avoid emit events when different empty arrays are set in the state.
    Removed some unnecessary actions from Url and ExtraParams Modules.

[EX-4564]: https://searchbroker.atlassian.net/browse/EX-4564?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/x-components/src/x-modules/url/components/__tests__/url-handler.spec.ts
```diff
@@ -116,71 +107,60 @@ describe('testing UrlHandler component', () => {
...
     emit('PushableUrlStateChanged', {
-      ...initialUrlState.params,
+      ...initialUrlState,
       query: 'lego',
       page: 2,
-      store: '111'
+      store: '111',
+      warehouse: '222'
     });
     const urlSearchParams = getCurrentUrlParams();
```
tiborux, line 165: Add a expect to check that the warehouse is null:

    expect(urlSearchParams.get('warehouse')).toBeNull();

packages/x-components/src/x-modules/url/store/mutations/set-params.mutation.ts
```diff
@@ -0,0 +1,9 @@
+import Vue from 'vue';
+import { forEach } from '../../../../utils/object';
+import { UrlXStoreModule } from '../types';
+
+export const setParams: UrlXStoreModule['mutations']['setParams'] = (state, params) => {
```
tiborux, line 5: Should we document this mutation?

tiborux (changes requested):

packages/x-components/src/x-modules/url/components/url-handler.vue
```diff
@@ -4,10 +4,19 @@
   import { XOn } from '../../../components/decorators/bus.decorators';
```
javieri-empathy, line 4: There are some issues that I need to investigate deeper.

1. Extra params are added immediately to the URL, even if it is the default value. I guess this is because of the order
2. Extra param change causes unstable history:
	* Search `lego`
	* Select a filter
	* Change the store to `Italy` -> Filters are then deselected
	* Hit back in the browser -> Filter is selected back, but store is kept in `Italy`
	* This happens because the filters are cleared in another batch of url updates.
3. Related tag selection is not undoable:
	* Search `playmobil`
	* Select any related tag
	* Hit back -> The tag is removed from the URL, but no new request is made.

javieri-empathy: 4. Selected filters component should exclude raw filters from being paint.

packages/x-components/src/x-modules/search/store/emitters.ts
```diff
@@ -7,7 +8,10 @@ import { searchXStoreModule } from './module';
  * @internal
  */
 export const searchEmitters = createStoreEmitters(searchXStoreModule, {
-  FacetsChanged: state => state.facets,
+  FacetsChanged: {
+    selector: state => state.facets,
+    filter: createArrayComparator('id')
```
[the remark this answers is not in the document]
tajespasarela: because I've forgotten it 🤦🏻 
rollback...

PR #197, opened 2021-10-21
Add Redirection and AutoProgressBar components

pmareke:
EX-4297

### DIFF
--- a/packages/x-components/src/components/__tests__/auto-progress-bar.spec.ts
+++ b/packages/x-components/src/components/__tests__/auto-progress-bar.spec.ts
@@ -47,7 +47,7 @@ describe('testing AutoProgressBar component', () => {
   it('render a progress bar component with an animation', () => {
     const { wrapper } = renderAutoProgressBar({ durationInSeconds: 5 });
 
-    expect(wrapper.find(getDataTestSelector('progress-bar__line')).attributes().style).toBe(
+    expect(wrapper.find(getDataTestSelector('progress-bar-line')).attributes().style).toBe(
       'animation-duration: 5s;'
     );
   });
--- a/packages/x-components/src/components/auto-progress-bar.vue
+++ b/packages/x-components/src/components/auto-progress-bar.vue
@@ -1,6 +1,6 @@
 <template>
   <div v-if="isLoading" class="x-progress-bar" data-test="progress-bar" role="progressbar">
-    <div class="x-progress-bar__line" :style="cssStyles" data-test="progress-bar__line"></div>
+    <div class="x-progress-bar-line" :style="cssStyles" data-test="progress-bar-line"></div>
   </div>
 </template>
 
@@ -54,7 +54,7 @@
     background-color: #b3b3b3;
     border-radius: 24px;
     margin: 32px 216px 0 215px;
-    &__line {
+    &-line {
       flex: 1 0 auto;
       background-color: #1a1a1a;
       animation-name: slide;
--- a/packages/x-components/src/views/Layout.vue
+++ b/packages/x-components/src/views/Layout.vue
@@ -230,13 +230,13 @@
 
         <template #main-body>
           <Redirection
-            v-slot="{ redirection, redirect, abortRedirect, isLoading, delayInSeconds }"
+            #default="{ redirection, redirect, abortRedirect, isRedirecting, delayInSeconds }"
             delayInSeconds="5"
           >
             <span>{{ redirection.url }}</span>
             <button @click="redirect">Redirect now!</button>
             <button @click="abortRedirect">Abort redirection!</button>
-            <AutoProgressBar :isLoading="isLoading" :durationInSeconds="delayInSeconds" />
+            <AutoProgressBar :isLoading="isRedirecting" :durationInSeconds="delayInSeconds" />
           </Redirection>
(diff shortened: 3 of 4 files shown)


packages/x-components/src/x-modules/search/components/__tests__/redirection.spec.ts
```diff
@@ -0,0 +1,161 @@
...
+    expect(onUserClickedARedirection).toHaveBeenCalledTimes(1);
+  });
+
+  it('accepts the redirection in auto mode', () => {
+    const { wrapper } = renderRedirection();
+    const onUserClickedARedirection = jest.fn();
+
+    jest.advanceTimersByTime(2000);
+
+    wrapper.vm.$x.on('UserClickedARedirection', true).subscribe(onUserClickedARedirection);
```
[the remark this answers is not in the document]
pmareke: Yes, my fault.

tiborux (changes requested):

packages/x-components/src/views/base-config.ts
```diff
@@ -6,6 +6,7 @@ import { realAdapter } from '../adapter/real-adapter';
 export const baseSnippetConfig: SnippetConfig = {
   instance: 'juguettos',
   lang: 'es',
+  env: 'staging',
```
tiborux, line 9: Is this right? Be careful

packages/x-components/src/x-modules/search/components/__tests__/redirection.spec.ts
```diff
@@ -0,0 +1,167 @@
...
+  it('renders the redirection component slot', () => {
+    const { redirectionWrapper } = renderRedirection({
+      template: `
+        <Redirection :mode="mode" :delay="delay" v-slot="{ redirection, redirect, abortRedirect }">
+          <span data-test="redirection-url">{{ redirection.url }}</span>
+        </Redirection>`
+    });
+
+    expect(redirectionWrapper.get(getDataTestSelector('redirection-url')).text()).toEqual(
+      'https://picsum.photos/seed/redirection/500'
```
tiborux, line 99: instead of hardcoding the string, can we use redirections[0].url?

packages/x-components/src/x-modules/search/components/__tests__/redirection.spec.ts
```diff
@@ -0,0 +1,167 @@
...
+      localVue,
+      store
+    }
+  );
+
+  const redirectionWrapper = wrapper.findComponent(Redirection);
+
+  return {
+    wrapper,
+    redirectionWrapper,
```
tiborux, line 55:
```suggestion
    wrapper: redirectionWrapper,
```

tiborux: If you do this, you can remove the redirectionWrapper from the RenderRedirectionAPI, because right now you are returning 2 times the same thing.

packages/x-components/src/x-modules/search/components/__tests__/redirection.spec.ts
```diff
@@ -0,0 +1,167 @@
...
+  beforeAll(() => {
+    jest.useFakeTimers();
+  });
+
+  afterAll(() => {
+    jest.useFakeTimers();
+  });
+
+  it('is an XComponent', () => {
+    const { redirectionWrapper } = renderRedirection();
```
tiborux, line 75: redirectionWrapper change it by wrapper in all places.

tiborux (changes requested):

packages/x-components/src/components/__tests__/auto-progress-bar.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+
+function renderAutoProgressBar({
+  template = '<AutoProgressBar :duration="duration" :isWaiting="isWaiting"/>',
+  duration = 1,
+  isWaiting = true
+}: RenderBaseAutoProgressBarOptions = {}): RenderAutoProgressBarAPI {
+  const wrapper = mount(
+    {
+      components: { AutoProgressBar },
+      props: ['duration', 'isWaiting'],
```
tiborux, line 13:
```suggestion
```
it is not necessary

packages/x-components/src/components/auto-progress-bar.vue
```diff
@@ -0,0 +1,93 @@
...
+
+    /**
+     * The duration in seconds of the progress bar.
+     *
+     * @public
+     */
+    @Prop({ default: 5 })
+    public duration!: number;
+
+    protected get cssStyles(): Partial<CSSStyleDeclaration> {
```
tiborux, line 34: document this

packages/x-components/src/views/base-config.ts
pmareke: We need to use staging in order to have Redirections in the layout view, also I don't like to use the customer production as default.

pmareke:
> I found a problem:
> 
> 1. Search from black friday
> 2. Before the  UserClickedARedirection will be emitted, type a new query.
> 3. The event of the redirection will be launched anyways.

Maybe we can listen the event `UserAcceptedAQuery` and clear the setInterval? What do you think?

PR #200, opened 2021-10-28
fix(facets): Fix facets service using only the first filter entity to…

javieri-empathy:
Fixes an issue where only the first filter was used to create an entity when updating the facets. If the first filter had a modifier, or was from a "less common" type, all the filters from different facets were updated using this behavior. 

For example, in added regression test, we are only setting the `gender` facet to single select, but because it is the first facet in the list of facets that we pass to the `updateFacets` method, the `size` facet was using also this behavior wrongly, when it should be multi-selectable (remember that multi-select is our new default).

Additionally, as we are now creating more entities, I've added a simple cache to the `FilterEntityFactory`. Because of this, tests that rely on the filters service should reset also the service, which fortunately it is quite easy to do.


packages/x-components/tests/e2e/cucumber/global-definitions.ts
```diff
@@ -3,7 +3,15 @@ import { defineParameterType } from 'cypress-cucumber-preprocessor/steps';
...
   name: 'boolean',
   regexp: /true|false/,
-  transformer(s) {
-    return s === 'true';
+  transformer(string) {
+    return string === 'true';
+  }
+});
+
+defineParameterType({
```
javieri-empathy, line 11: Rollback this

packages/x-components/src/x-modules/facets/components/facets/__tests__/facets-provider.spec.ts
```diff
@@ -129,6 +130,8 @@ function renderFacetsProviderComponent({
   stateFacets = {},
   providedFacets
 }: FacetsRenderOptions = {}): FacetsComponentAPI {
+  DefaultFacetsService.instance = new DefaultFacetsService(new FilterEntityFactory());
```
pmareke, line 133: Just to give some context here. What it's the difference of passing the `FilterEntityFactory`?

packages/x-components/src/x-modules/facets/entities/filter-entity.factory.ts
```diff
@@ -38,6 +38,8 @@ export class FilterEntityFactory {
    */
   protected modifiers: Record<Facet['id'], FilterEntityModifier[]> = {};
 
+  protected cache: Record<Facet['id'] | '__unknown-facet__', FilterEntity> = {};
```
pmareke, line 41: Please add doc.

packages/x-components/src/x-modules/facets/entities/filter-entity.factory.ts
```diff
@@ -47,6 +49,13 @@ export class FilterEntityFactory {
    * @returns The {@link FilterEntity} created by the factory.
    */
   createFilterEntity(store: Store<RootXStoreState>, filter: Filter): FilterEntity {
+    const cacheKey = isFacetFilter(filter) ? filter.facetId : '__unknown-facet__';
+    return (
+      this.cache[cacheKey] ?? (this.cache[cacheKey] = this.instantiateFilterEntity(store, filter))
+    );
+  }
+
+  protected instantiateFilterEntity(store: Store<RootXStoreState>, filter: Filter): FilterEntity {
```
pmareke, line 58: Should we document this new method?

packages/x-components/src/x-modules/facets/service/__tests__/facets-service.spec.ts
```diff
@@ -575,25 +582,57 @@ describe('testing facets service', () => {
         id: 'backend',
         facets: [newSizeFacet]
       });
+
+      expect(getSelectedFilters()).toHaveLength(1);
```
pmareke, line 586: Should we check if the selected is the correct one (in our case the last one)?

`expect(getSelectedFilters()).toStrictEqual([newSizeFacet.filters[2]]);`

packages/x-components/tests/e2e/cucumber/common-steps.spec.ts
```diff
@@ -214,7 +221,6 @@ Given('a results API with a known response', () => {
           ...createFilter('Muñecas', false, createFilter => [
             ...createFilter('Peluches', false),
             ...createFilter('Ropa y accesorios', false),
-            ...createFilter('Muñecas', false),
```
pmareke, line 217: Why did you remove this filter?

packages/x-components/tests/e2e/cucumber/multiselect-filters.feature
```diff
@@ -13,18 +13,17 @@ Feature: MultiSelect filters component
...
-  # Then  selected filter is shown in the selected filters list
-  # And   search request contains selected filter
-  # TODO  https://searchbroker.atlassian.net/browse/EX-4866
-  # And   clear-filters button displays the number of selected filters
+    And   filter "<filter2>" is clicked in facet "<facetName>"
+    Then  filter "<filter1>" is shown in the selected filters list
+    And   filter "<filter2>" is shown in the selected filters list
+    And   search request contains "<filter1>" filter
+    And   search request contains "<filter2>" filter
+    And   clear-filters button should have 2 filters selected
```
pmareke, line 25: I'm not sure if I like this `2` here.

packages/x-components/tests/e2e/cucumber/multiselect-filters.spec.ts
```diff
@@ -1,46 +1,24 @@
...
+  cy.getByDataTest('selected-filters-list')
+    .getByDataTest('filter')
+    .contains(filter)
+    .should('exist');
+});
 
-And('search request contains selected filter', function (this: { clickedFilter: string }) {
-  cy.wait('@requestWithFilter').its('request.body').should('include', this.clickedFilter);
+And('search request contains {string} filter', (filter: string) => {
+  cy.wait('@requestWithFilter').its('request.body').should('include', filter);
```
pmareke, line 19: Should we use here the `json.Parse` as the other PR, I really like it.

pmareke (changes requested):

packages/x-components/tests/e2e/cucumber/common-steps.spec.ts
javieri-empathy: Because it was a duplicated filter which was causing an infinite loop issue. Filters are indexed by its id, and this filter has the same id  than its parent. You can imagine the rest of the story.

packages/x-components/src/x-modules/facets/components/facets/__tests__/facets-provider.spec.ts
javieri-empathy: Both the facets service and the entity factory are singletons, the problem is that they are stateless singletons, which is an issue to test different flow. In this line we are "resetting" the singleton by creating new instances.

packages/x-components/tests/e2e/cucumber/multiselect-filters.feature
javieri-empathy: Then think about it, and get back to this comment when you know it 🤣.

packages/x-components/tests/e2e/cucumber/multiselect-filters.spec.ts
javieri-empathy: I think they are both different cases. Here it would be more complicated because it is searching a filter by its label. I would leave it like this.

packages/x-components/src/x-modules/facets/service/__tests__/facets-service.spec.ts
javieri-empathy: This test is just to ensure that only one filter is selected. I don't think the "which one" matters, as the outcome is that the app keeps working properly.

packages/x-components/tests/e2e/cucumber/multiselect-filters.feature
pmareke: I was thinking about using a variable!

pmareke: `DON'T USE MAGIC NUMBERS`

pmareke (approved):

pmareke (changes requested):
Fix learn errors, please.

pmareke (dismissed):

packages/x-components/src/x-modules/facets/components/__tests__/clear-filters.spec.ts
```diff
@@ -29,6 +30,8 @@ function renderClearFilters({
     </ClearFilters>
   `
 }: RenderFiltersOptions = {}): RenderFiltersAPI {
+  DefaultFacetsService.instance = new DefaultFacetsService(new FilterEntityFactory());
```
tajespasarela, line 33: extract this to a tests utils method

packages/x-components/src/x-modules/facets/entities/filter-entity.factory.ts
```diff
@@ -38,15 +38,39 @@ export class FilterEntityFactory {
...
    * Creates a new FilterEntity from a filter.
    *
    * @param store - The {@link https://vuex.vuejs.org/guide/ | Vuex Store} that the entity will
    * use.
-   * @param filter - The filter necesary to know what Entity to create.
+   * @param filter - The {@link @empathyco/x-types#Filter | Filter} necessary to know what
+   * {@link FilterEntity} to create.
    * @returns The {@link FilterEntity} created by the factory.
    */
   createFilterEntity(store: Store<RootXStoreState>, filter: Filter): FilterEntity {
```
tajespasarela, line 57: this method doesn't create a FilterEntity anymore. Rename to `getFilterEntity`.

Therefore, rename the `instantiateFilterEntity` to `createFilterEntity`

packages/x-components/src/x-modules/facets/service/facets.service.ts
```diff
@@ -83,23 +85,19 @@ export class DefaultFacetsService implements FacetsService {
...
-    const previousFilters = this.removeGroupFilters(facetsGroup.id);
+    this.removeGroupFilters(facetsGroup.id);
     facetsGroup.facets.forEach(facet => {
       this.setFacetGroup({ facetId: facet.id, groupId: facetsGroup.id });
       this.setFacet(facet);
     });
     const newFilters = facetsGroup.facets.flatMap(facet => facet.filters);
     this.setFilters(newFilters);
-    return { newFilters, previousFilters };
+    return newFilters;
```
tajespasarela, line 100: I don't remember why we needed the `previousFilters` for.

tajespasarela (changes requested):

packages/x-components/src/x-modules/facets/service/facets.service.ts
javieri-empathy: We used them to keep the previous selected state of the filters in the `updateFacets` method, but then this method was changed to first retrieve the selected filters, then replacing the filters in the store, and finally updating the filters selected state. I don't know the reason of this change, but I prefer it, as it avoids us to return 2 things from the same method. I don't think there is any issue with this approach.

packages/x-components/tests/e2e/cucumber/multiselect-filters.feature
javieri-empathy: This is not a magic number, as per wikipedia, magic numbers are:

> Unique values with unexplained meaning or multiple occurrences which could (preferably) be replaced with named constants

Here I'm sure this number was perfectly understood, the human readable sentence gave enough context about it. Anyway, I've changed it just for consistency. Even if I prefer to hardcode values in tests with only one example.

tajespasarela (approved):

pmareke (approved):

PR #206, opened 2021-10-29
fix(x-plugin): Fix no results message flash on load.

tajespasarela:
EX-4951

This PR fixes the No results message flashes on load.

Steps to reproduce:

1- load with a query in URL multiple times
2- Sometimes you'll see the No Results Message flashes momentaneously

### DIFF
--- a/packages/x-components/src/store/utils/status-store.utils.ts
+++ b/packages/x-components/src/store/utils/status-store.utils.ts
@@ -31,7 +31,7 @@ export interface StatusMutations {
  *
  * @public
  */
-export type RequestStatus = 'success' | 'loading' | 'error';
+export type RequestStatus = 'success' | 'loading' | 'error' | '';
 
 /**
  * Sets the request status. Can be used as a mutation.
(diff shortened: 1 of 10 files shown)


packages/x-components/src/store/utils/status-store.utils.ts
```diff
@@ -31,7 +31,7 @@ export interface StatusMutations {
  *
  * @public
  */
-export type RequestStatus = 'success' | 'loading' | 'error';
+export type RequestStatus = 'success' | 'loading' | 'error' | '';
```
javieri-empathy, line 34: Just to keep the semantic naming, I would probably use `'initial'` instead of `''`

javieri-empathy (commented):
I think it works. Tested with CPU & network slowdown:

- From `success` with **no** results to `success` with no-results
- From `success` with **no** results to `success` with results
- From `success` with results to `success` with results
- From `success` with results to `success` with **no** results
- URL Load with **no** results
- URL Load with results

javieri-empathy (approved):

pmareke (approved):

PR #207, opened 2021-10-29
Add origin

javieri-empathy:
EX-4869

Adds support for the new tracking parameters. Both queries and results will have its own `origin`. This origin can be used to determine their effectiveness in relation to the tool that created them, and the location where this "tool" is placed.

- origin: It is composed by two parts
  - feature: The name of the tool that originated the query or the result
  - location: Where the visual representation of the tool that originated the query or result is placed

Components will have a hardcoded feature, and receive the injected location.


packages/x-components/src/plugins/__tests__/x-plugin-mixin.spec.ts
```diff
@@ -26,54 +28,91 @@ describe('testing $x component API global mixin', () => {
...
-      const listener = jest.fn();
-
-      @Component({
-        template: `<div><slot/></div>`,
-        provide: {
-          origin: 'origin-test'
-        }
-      })
-      class Provider extends Vue {}
+    interface RenderOriginComponentOptions {
```
pmareke, line 31: Can we document these interfaces please?

packages/x-components/src/plugins/x-plugin.mixin.ts
```diff
@@ -88,3 +89,21 @@ export function getRootXComponent(component: Vue): XComponent | undefined {
...
+/**
+ * Creates an {@link Origin} string given a {@link Feature} and a {@link FeatureLocation}. If it
+ * can't create it, it will return `undefined`.
+ *
+ * @param feature - The feature that originated the query, or `undefined` if the event is not
+ * related with a query.
+ * @param location - The location where the event has happened.
+ * @returns The composed origin, or `undefined` if it is not able to create the origin.
+ */
+function createOrigin(
```
pmareke, line 102: Should we add the @internal decorator here?

packages/x-components/src/types/origin.ts
```diff
@@ -0,0 +1,54 @@
...
+  | 'url_history_pdp'
+  | 'url_history'
+  | 'predictive_layer'
+  | 'no_query'
+  | 'results'
+  | 'no_results'
+  | 'partial_results';
+
+/**
+ * The tool that generated a query.
```
pmareke, line 26: `The tool that generates a query.`

packages/x-components/src/wiring/wiring.types.ts
```diff
@@ -127,8 +129,7 @@ export interface WireService<SomeService> {
    *
    * @param method - The method to invoke.
    * @returns A Wire that expects to receive the function parameter as payload.
-   */
-  <SomeMethod extends PropsWithType<SomeService, MonadicFunction>>(method: SomeMethod): Wire<
+   */ <SomeMethod extends PropsWithType<SomeService, MonadicFunction>>(method: SomeMethod): Wire<
```
pmareke, line 132: You're commenting the type, are you?

packages/x-components/src/wiring/wiring.types.ts
```diff
@@ -137,8 +138,7 @@ export interface WireService<SomeService> {
    * @param method - The method to invoke.
    * @param payload - The payload to invoke the service with.
    * @returns A Wire that can be used anywhere.
-   */
-  <SomeMethod extends PropsWithType<SomeService, MonadicFunction>>(
+   */ <SomeMethod extends PropsWithType<SomeService, MonadicFunction>>(
```
pmareke, line 141: Same here.

packages/x-components/src/wiring/wiring.types.ts
```diff
@@ -155,6 +155,5 @@ export interface WireServiceWithoutPayload<SomeService> {
    *
    * @param method - The method to invoke.
    * @returns A Wire that can be used anywhere.
-   */
-  <SomeMethod extends PropsWithType<SomeService, NiladicFunction>>(method: SomeMethod): AnyWire;
+   */ <SomeMethod extends PropsWithType<SomeService, NiladicFunction>>(method: SomeMethod): AnyWire;
```
pmareke, line 158: And here.

packages/x-components/src/x-modules/search/store/types.ts
```diff
@@ -32,7 +32,7 @@ export interface SearchState extends StatusState {
   /** A flag to indicate if new results are append to the current instead of replacing them. */
   isAppendResults: boolean;
   /** The origin property of the request. */
-  origin: QueryOrigin | null;
+  origin: Origin | null;
```
pmareke, line 35: Why here is `null` but later(line 109) is `undefined`?

packages/x-components/src/components/result/base-result-link.vue
```diff
@@ -37,15 +35,7 @@
...
-    /**
-     * The origin to be sent as part of the `params` property in the
-     * {@link @empathyco/x-types#Tagging | tagging} information.
-     *
-     * @public
-     */
-    @Inject({ from: 'origin', default: 'default' })
-    protected origin!: QueryOrigin;
-
+    public $el!: HTMLElement;
```
pmareke, line 38: Could you add some documentation here please?

packages/x-components/src/plugins/__tests__/x-plugin-mixin.spec.ts
```diff
@@ -26,54 +28,91 @@ describe('testing $x component API global mixin', () => {
...
+    });
 
-      wrapper
-        .findComponent(Injector)
-        .vm.$x.emit('UserIsTypingAQuery', 'So awesome, much quality, such skill');
-      componentInstance.vm.$x.on('UserIsTypingAQuery', true).subscribe(listener);
+    it('emits a valid origin when both location and feature are available', () => {
+      const { wrapper, busListener } = renderOriginComponent({
+        feature: 'search_box',
+        location: 'predictive_layer'
```
pmareke, line 106: I'm not sure what is the syntax for this `location` as I can see both `location: 'empty-search'`(in the `bus.decorators.spec.ts`  file) and ` location: 'predictive_layer'`.

pmareke (changes requested):
Great job mate, ❤️

herrardo (changes requested):

packages/x-components/src/wiring/wiring.types.ts
javieri-empathy: I did this change accidentally when removing unused imports. I rolled it back.

packages/x-components/src/x-modules/search/store/types.ts
javieri-empathy: Good catch, it should be `null`

packages/x-components/src/types/origin.ts
```diff
@@ -0,0 +1,54 @@
...
+export type Feature = QueryFeature | ClickFeature;
+
+/**
+ * The name of the feature that the query comes from.
+ *
+ * @public
+ */
+export type QueryFeature =
+  | 'search_box'
+  | 'linked'
```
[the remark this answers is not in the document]
javieri-empathy: Changed it. Any more pending changes? Should I talk with someone from the data or playboard team just to confirm?

packages/x-components/src/plugins/__tests__/x-plugin-mixin.spec.ts
javieri-empathy: `empty-search` shouldn't exist. Fixed it.

packages/x-components/src/plugins/__tests__/x-plugin-mixin.spec.ts
javieri-empathy: Done

pmareke (approved):

packages/x-components/src/plugins/x-plugin.mixin.ts
```diff
@@ -60,11 +61,15 @@ export function getBusAPI(bus: XBus, component: PrivateExtendedVueComponent): XC
...
-      metadata: Omit<WireMetadata, 'moduleName'> = {}
+      metadata: Omit<WireMetadata, 'moduleName' | 'origin'> = {}
     ) => {
       const xComponent = component.xComponent;
-      const moduleName = xComponent ? getXComponentXModuleName(xComponent) : null;
-      bus.emit(event, payload as any, { moduleName, origin: component.$origin, ...metadata });
+      bus.emit(event, payload as any, {
+        moduleName: xComponent ? getXComponentXModuleName(xComponent) : null,
+        origin: createQueryOrigin(metadata.feature, component.$location as QueryLocation),
+        location: component.$location,
```
javieri-empathy, line 70: I'm not sure about using the `location` for tracking result events. I think we are mixing again the feature with the location we place it. For me it is clearer to see `recommendations:predictive_layer`, `recommendations:no_results`, rather than `no_results`, and having to think what feature are we using in the no_results? Is it the recommendations? Or viceversa, seeing `recommendations` without an `origin` and thinking _"hmm, they haven't origin so they might have no query associated. I guess these are the recommendations that we paint in the `no_query` location"_

packages/x-components/src/types/result-location.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Indicates where the result where the event comes from is located.
+ *
+ * @public
+ */
+export type ResultLocation =
+  | 'recommendations'
```
javieri-empathy, line 7: I know we talked yesterday about using `recommendations` here because if the reccommendations come without origin they should be the ones in the `no_results` situation, but I would name them `no_query`. How are we going to distinguish them from the ones in the `predictive_layer`?

pmareke (dismissed):

packages/x-components/src/plugins/x-plugin.mixin.ts
```diff
@@ -60,11 +61,15 @@ export function getBusAPI(bus: XBus, component: PrivateExtendedVueComponent): XC
...
       payload?: XEventPayload<Event>,
-      metadata: Omit<WireMetadata, 'moduleName'> = {}
+      metadata: Omit<WireMetadata, 'moduleName' | 'origin'> = {}
     ) => {
       const xComponent = component.xComponent;
-      const moduleName = xComponent ? getXComponentXModuleName(xComponent) : null;
-      bus.emit(event, payload as any, { moduleName, origin: component.$origin, ...metadata });
+      bus.emit(event, payload as any, {
+        moduleName: xComponent ? getXComponentXModuleName(xComponent) : null,
+        origin: createQueryOrigin(metadata.feature, component.$location as QueryLocation),
```
tajespasarela, line 69: I'm not sure this is the right place to do this. The event should contain the feature and the location. Then the search  
 merge them before send to the API. Once the API responses with the tagging object, that info is captured by Tagging Module, and tracked to Tagging API.
When a result click/addToCart/... is done, the origin of the query is already in the Tagging Module, to add to the click track.

packages/x-components/src/plugins/x-plugin.mixin.ts
```diff
@@ -60,11 +61,15 @@ export function getBusAPI(bus: XBus, component: PrivateExtendedVueComponent): XC
     emit: <Event extends XEvent>(
       event: Event,
       payload?: XEventPayload<Event>,
-      metadata: Omit<WireMetadata, 'moduleName'> = {}
+      metadata: Omit<WireMetadata, 'moduleName' | 'origin'> = {}
```
tajespasarela, line 64: shouldn't be the location omitted too?

packages/x-components/src/components/result/base-result-link.vue
```diff
@@ -62,40 +57,26 @@
     @Prop({ required: true })
     protected result!: Result;
 
-    /**
```
tajespasarela, line 65: remove the origin mentioned in the Events list down on the docs block

packages/x-components/src/x-modules/search/store/types.ts
```diff
@@ -106,7 +106,7 @@ export interface SearchMutations extends StatusMutations {
    *
    * @param origin - The new origin.
    */
-  setOrigin(origin: QueryOrigin | undefined): void;
+  setOrigin(origin: QueryOrigin | undefined | null): void;
```
tajespasarela, line 109: what's the difference between having `undefined` origin and having `null` origin? do we need both?

tajespasarela (changes requested):

packages/x-components/src/plugins/x-plugin.mixin.ts
```diff
@@ -15,8 +16,8 @@ declare module 'vue/types/vue' {
 }
 
 interface PrivateExtendedVueComponent extends Vue {
+  $location?: QueryLocation | ResultLocation;
```
tajespasarela, line 19: If we want this injected location to be reactive, we need the @XProvide and @XInject

tajespasarela (changes requested):

packages/x-components/src/types/result-location.ts
javieri-empathy: Done, merged both now that it is tracked in the same way

packages/x-components/src/plugins/x-plugin.mixin.ts
javieri-empathy: I think this is the right place, the components are the only place you can retrieve both the location and the feature, and merge them into an origin. I see this as "enriching the metadata". If not you would have to do this in a mutation.

packages/x-components/src/x-modules/search/store/types.ts
javieri-empathy: They are equivalent, indeed they should be transformed both to `null`. The point is that the `origin` is optional in the metadata, and this mutation params were extended to support also that case.

packages/x-components/src/x-modules/search/wiring.ts
```diff
@@ -46,7 +47,7 @@ export const cancelFetchAndSaveSearchResponseWire = wireDispatchWithoutPayload(
  *
  * @public
  */
-export const setOrigin = wireCommit('setOrigin', ({ metadata }) => metadata.origin);
+export const setOrigin = wireCommit('setOrigin', ({ metadata }) => metadata.origin as QueryOrigin);
```
tajespasarela, line 50: We can do this:
```
export const setOrigin = wireCommit('setOrigin', 
  ({ metadata: { feature, location } }) => createQueryOrigin(feature, location)
);
```
or we can create an action to do that, to not overload the wire with logic

packages/x-components/src/x-modules/search/store/types.ts
tajespasarela: with the change I suggest below, the `createQueryOrigin` can always return an origin or `null`

tajespasarela (changes requested):

packages/x-components/src/plugins/x-plugin.mixin.ts
```diff
@@ -88,3 +91,22 @@ export function getRootXComponent(component: Vue): XComponent | undefined {
...
+ * Creates a {@link QueryOrigin} string given a {@link QueryFeature} and a {@link Location}.
+ * If the {@link QueryOrigin} can't be created, it returns `undefined`.
+ *
+ * @param feature - The feature that originated the query, or `undefined` if the event is not
+ * related with a query.
+ * @param location - The location where the event has happened.
+ * @returns The composed origin, or `undefined` if it is not able to create the origin.
+ * @internal
+ */
+function createQueryOrigin(
```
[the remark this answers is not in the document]
javieri-empathy: single-responsibility

Jokes aside, I agree that the bus is not the place to do it, even if I don't consider the `x-plugin.mixin` the bus itself, but a mixin that integrates the bus into the components adding features.

But I thought that the `map` operator in the wiring was supposed to be used as a "pick", not for creating new elements. If I have to move the createOrigin from where now it is, I'd rather move it into the `setOrigin` mutation that we have.

packages/x-components/src/x-modules/search/wiring.ts
javieri-empathy: Done with action

packages/x-components/src/plugins/x-plugin.mixin.ts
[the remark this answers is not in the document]
javieri-empathy: I hadn't read that you agree to do it in an action. I did in in that place instead, much better.

packages/x-components/src/x-modules/scroll/config.types.ts
```diff
@@ -0,0 +1,6 @@
+/**
```
tajespasarela, line 1: remove this file from this PR

packages/x-components/src/x-modules/search/wiring.ts
```diff
@@ -46,7 +46,7 @@ export const cancelFetchAndSaveSearchResponseWire = wireDispatchWithoutPayload(
  *
  * @public
  */
-export const setOrigin = wireCommit('setOrigin', ({ metadata }) => metadata.origin);
+export const setOrigin = wireDispatch('saveOrigin', ({ metadata }) => metadata);
```
tajespasarela, line 49: rename the const to `saveOrigin`

tajespasarela (changes requested):

tajespasarela (approved):

PR #208, opened 2021-10-29
docs(eslint-plugin-x): Update readme to reference the correct package

ivantxu:
Updated Readme on the eslint-plugin-x package to reference the correct package (@empathyco), was previously mentioning @empathy.


pmareke (approved):
Great, thanks!

PR #218, opened 2021-11-10
feat(tagging): add `SessionService` [EX-4875]

pmareke:
# Pull request template

EX-4875

## Motivation and context

This PR adds a new session id service, which allows us to get and remove the current session id.

## Type of change

- [x] `Main`
- [ ] Other. Specify:

## How has this been tested?

In order to see the new property in the `localstorage` you should register the tagging module in the `Layout` and add the `wireServices` in the tagging wiring.

## Checklist:

### DIFF
--- a/packages/x-components/.eslintrc.js
+++ b/packages/x-components/.eslintrc.js
@@ -5,6 +5,7 @@ module.exports = {
     project: 'tsconfig.eslint.json'
   },
   rules: {
-    'no-dupe-class-members': 'off'
+    'no-dupe-class-members': 'off',
+    '@typescript-eslint/no-unused-vars-experimental': 'off'
   }
 };
--- a/packages/x-components/src/x-modules/tagging/service/__tests__/session.service.spec.ts
+++ b/packages/x-components/src/x-modules/tagging/service/__tests__/session.service.spec.ts
@@ -1,0 +1,89 @@
+import { DefaultSessionService } from '../session.service';
+import { StorageService } from '../types';
+
+/**
+ * Storage Service mock version.
+ *
+ * @internal
+ */
+class MockedStorageService implements StorageService {
+  getSessionIdSpy = jest.fn();
+  setSessionIdSpy = jest.fn();
+  removeSessionIdSpy = jest.fn();
+
+  constructor(protected storage: Storage, protected prefix: string) {
+  }
+
+  getPrefix(): string {
+    return this.prefix;
+  }
+
+  injectGetResponse(key: string): void {
+    this.getSessionIdSpy.mockReturnValue(key);
+  }
+
+  getItem<T = any>(key: string): any {
+    return this.getSessionIdSpy(key);
+  }
+
+  setItem(key: string, item: any, ttlInMs?: number) {
+    this.setSessionIdSpy(key, item, ttlInMs);
+  }
+
+  removeItem<T = any>(key: string): any {
+    this.removeSessionIdSpy(key);
+    return key;
+  }
+
+  clear(): number {
+    return  0;
+  }
+}
+
+const prefix = 'test';
+const mockedStorageService = new MockedStorageService(localStorage, prefix);
+const sessionService  = new DefaultSessionService(mockedStorageService, 1);
+
+describe('testing session id service', () => {
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('creates a new session id', () => {
+    const session = sessionService.getSessionId();
+    expect(mockedStorageService.getPrefix()).toBe(prefix);
+    expect(mockedStorageService.getSessionIdSpy).toHaveBeenCalledTimes(1);
+    expect(mockedStorageService.setSessionIdSpy).toHaveBeenCalledTimes(1);
+    expect(mockedStorageService.removeSessionIdSpy).not.toHaveBeenCalled();
+    expect(mockedStorageService.setSessionIdSpy).toHaveBeenCalledWith(
+      'session-id',
+      expect.any(String),
+      expect.any(Number)
+    );
+    expect(session).toHaveLength(21);
+    expect(session).toMatch(/[A-Za-z0-9_-]/);
+  });
+
+  it('returns an existing session id', () => {
+    mockedStorageService.injectGetResponse('session-id');
+    const session = sessionService.getSessionId();
+    expect(mockedStorageService.getPrefix()).toBe(prefix);
+    expect(mockedStorageService.getSessionIdSpy).toHaveBeenCalledTimes(1);
+    expect(mockedStorageService.setSessionIdSpy).toHaveBeenCalledTimes(1);
+    expect(mockedStorageService.removeSessionIdSpy).not.toHaveBeenCalled();
+    expect(mockedStorageService.setSessionIdSpy).toHaveBeenCalledWith(
+      'session-id',
+      'session-id',
+      expect.any(Number)
+    );
+    expect(session).toBe('session-id');
+  });
+
+  it('removes an existing session id', () => {
+    sessionService.clearSessionId();
+    expect(mockedStorageService.getPrefix()).toBe(prefix);
+    expect(mockedStorageService.getSessionIdSpy).not.toHaveBeenCalled();
+    expect(mockedStorageService.setSessionIdSpy).not.toHaveBeenCalled();
+    expect(mockedStorageService.removeSessionIdSpy).toHaveBeenCalledTimes(1);
+  });
+});
--- a/packages/x-components/src/x-modules/tagging/service/__tests__/session.service.spec.ts
+++ b/packages/x-components/src/x-modules/tagging/service/__tests__/session.service.spec.ts
@@ -26,7 +26,7 @@ describe('testing session id service', () => {
   });
 
   it('returns an existing session id', () => {
-    mockedStorageService.injectGetResponse('session-id');
+    mockedStorageService.injectGetResponse('abcd-1234');
     const session = sessionService.getSessionId();
     expect(mockedStorageService.getPrefix()).toBe(prefix);
     expect(mockedStorageService.getSessionIdSpy).toHaveBeenCalledTimes(1);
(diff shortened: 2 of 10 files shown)


packages/x-components/.eslintrc.js
```diff
@@ -5,6 +5,7 @@ module.exports = {
     project: 'tsconfig.eslint.json'
   },
   rules: {
-    'no-dupe-class-members': 'off'
+    'no-dupe-class-members': 'off',
+    '@typescript-eslint/no-unused-vars-experimental': 'off'
```
pmareke, line 9: I don't know why, but the lint is falling and I'm not able to fix it.




Is something related with the `MockedStorageService`

packages/x-components/src/x-modules/tagging/service/__tests__/session.service.spec.ts
```diff
@@ -0,0 +1,50 @@
+import { DefaultSessionService } from '../session.service';
+import { MockedStorageService } from './utils/mock-storage.service';
+
+describe('testing session id service', () => {
+  const prefix = 'test';
+  const mockedStorageService = new MockedStorageService(prefix);
```
pmareke, line 6: In order to test the SessionService I mocked a custom service in order to spy the calls and the payloads.

packages/x-components/.eslintrc.js
pmareke: After some research I've discovered that we're using `@typescript-eslint/eslint-plugin": "^3.1.0"` and this rule was remove in the version [5](https://github.com/typescript-eslint/typescript-eslint/blob/4875cd6d224db67337f066d2a2cfb15e00587cf8/CHANGELOG.md#500-2021-10-11)

image

packages/x-components/src/x-modules/tagging/service/types.ts
```diff
@@ -0,0 +1,21 @@
...
+export interface SessionService {
+  /**
+   * Get the current session id from the storage:
+   * * If it's present then:
+   * * * If it's valid just return it.
+   * * * If it's outdated it will generate a new one.
+   * * If not creates a new one.
+   *
+   * @returns The session id.
+   */
```
tiborux, line 15:
```suggestion
  /**
   * Get the current session id from the storage.
   *
   * @returns The session id.
   */
```

We are explaining what it's doing in the other file. I think that its enough to say that is returning you the session id.

packages/x-components/src/x-modules/tagging/service/__tests__/session.service.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+  const sessionService = new DefaultSessionService(mockedStorageService, 1);
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('creates a new session id', () => {
+    const session = sessionService.getSessionId();
+    expect(mockedStorageService.getPrefix()).toBe(prefix);
+    expect(mockedStorageService.getSessionIdSpy).toHaveBeenCalledTimes(1);
```
tiborux, line 16: can we extract the equal part to an auxiliary function? Like:

```
check(value: string) {
    expect(mockedStorageService.getPrefix()).toBe(prefix);
    expect(mockedStorageService.getSessionIdSpy).toHaveBeenCalledTimes(1);
    expect(mockedStorageService.setSessionIdSpy).toHaveBeenCalledTimes(1);
    expect(mockedStorageService.removeSessionIdSpy).not.toHaveBeenCalled();
    expect(mockedStorageService.setSessionIdSpy).toHaveBeenCalledWith(
      'session-id',
      value,
      expect.any(Number)
    );


}
```

packages/x-components/src/x-modules/tagging/service/__tests__/session.service.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+    expect(mockedStorageService.setSessionIdSpy).toHaveBeenCalledTimes(1);
+    expect(mockedStorageService.removeSessionIdSpy).not.toHaveBeenCalled();
```
LuisMartinez15, line 18: I don't think this check is necessary here. Why should we check `removeSessionId` if has been called or not here?

```suggestion
```

packages/x-components/src/x-modules/tagging/service/__tests__/session.service.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+describe('testing session id service', () => {
+  const prefix = 'test';
+  const mockedStorageService = new MockedStorageService(prefix);
+  const sessionService = new DefaultSessionService(mockedStorageService, 1);
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('creates a new session id', () => {
```
LuisMartinez15, line 13:
```suggestion
  it('creates a new session id when calling getSessionId and a session does not exist', () => {
```

tiborux (changes requested):
Good job!

packages/x-components/src/x-modules/tagging/service/__tests__/session.service.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+    expect(session).toMatch(/[A-Za-z0-9_-]/);
+  });
+
+  it('returns an existing session id', () => {
+    mockedStorageService.injectGetResponse('abcd-1234');
+    const session = sessionService.getSessionId();
+    expect(mockedStorageService.getPrefix()).toBe(prefix);
+    expect(mockedStorageService.getSessionIdSpy).toHaveBeenCalledTimes(1);
+    expect(mockedStorageService.setSessionIdSpy).toHaveBeenCalledTimes(1);
+    expect(mockedStorageService.removeSessionIdSpy).not.toHaveBeenCalled();
```
LuisMartinez15, line 34: Same as above. I don't think checking this provides any value to this test.

```suggestion
```

packages/x-components/src/x-modules/tagging/service/__tests__/session.service.spec.ts
```diff
@@ -0,0 +1,50 @@
+import { DefaultSessionService } from '../session.service';
+import { MockedStorageService } from './utils/mock-storage.service';
+
+describe('testing session id service', () => {
+  const prefix = 'test';
+  const mockedStorageService = new MockedStorageService(prefix);
+  const sessionService = new DefaultSessionService(mockedStorageService, 1);
+
+  afterEach(() => {
```
LuisMartinez15, line 9: Use `beforeEach` instead of `afterEach` whenever possible.

```suggestion
  beforeEach(() => {
```
```suggestion
  afterEach(() => {
```

packages/x-components/src/x-modules/tagging/service/__tests__/utils/mock-storage.service.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  public constructor(protected _prefix: string) {
+    super(undefined, _prefix);
+  }
+
+  getPrefix(): string {
+    return this._prefix;
+  }
+
+  injectGetResponse(key: string): void {
```
LuisMartinez15, line 21: This name is a little bit weird as I needed to look to the code to see the real intention, what about `mockSessionIdValue`?

packages/x-components/src/x-modules/tagging/service/__tests__/session.service.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+      expect.any(Number)
+    );
+    expect(session).toBe('abcd-1234');
+  });
+
+  it('removes an existing session id', () => {
+    sessionService.clearSessionId();
+    expect(mockedStorageService.getPrefix()).toBe(prefix);
+    expect(mockedStorageService.getSessionIdSpy).not.toHaveBeenCalled();
+    expect(mockedStorageService.setSessionIdSpy).not.toHaveBeenCalled();
```
LuisMartinez15, line 47: These checks do not provide any value for this test.
Where and when was the sessionId created? Shouldn't we check after calling `clearSessionId()` that there is no `session` in the `storage`?

packages/x-components/src/x-modules/tagging/service/__tests__/session.service.spec.ts
```diff
@@ -0,0 +1,50 @@
...
+    expect(mockedStorageService.removeSessionIdSpy).toHaveBeenCalledTimes(1);
+  });
+});
```
LuisMartinez15, line 50: Shouldn't we also test the refresh `session` scenario?

LuisMartinez15 (changes requested):

packages/x-components/src/x-modules/tagging/service/session.service.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  /**
+   * Generates an unique session id.
+   *
+   * @returns The new session id.
+   *
+   * @internal
+   */
+  private createSessionId = (): string => nanoid();
```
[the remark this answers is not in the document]
pmareke: Honestly right now it doesn't but I wasn't sure if in the future this generation could change.

For now I'm going to remove it, thanks!.

packages/x-components/src/x-modules/tagging/service/__tests__/session.service.spec.ts
pmareke: I'm not sure if we need to test anything related to the `storage`, I don't care what happens in the storage package.

I'll remove this two `not.toHaveBeenCalled` lines.

packages/x-components/src/x-modules/tagging/service/__tests__/session.service.spec.ts
pmareke: It's a nice idea but it only repeats two times so I prefer to keep like it is.

packages/x-components/src/x-modules/tagging/service/__tests__/session.service.spec.ts
LuisMartinez15: As you said it is "implicitly" tested in the second test case 👍🏿

LuisMartinez15 (approved):

LuisMartinez15:
amazing job, keep it up avilesino! 💪🏿

tiborux (approved):

PR #219, opened 2021-11-11
feat(scroll): add empty `XModule`

javieri-empathy:
# Description

Adds a basic scroll x module. The scroll module is needed to solve the problems derived from using observables to keep the state.

Adding a scroll module will also help storing the scroll position in the URL. This is the 1st one of 3 expected PRs.

### DIFF
--- a/packages/x-components/src/wiring/events.types.ts
+++ b/packages/x-components/src/wiring/events.types.ts
@@ -34,14 +34,16 @@ import { UrlXEvents } from '../x-modules/url/events.types';
  * * {@link QuerySuggestionsXEvents},
  * * {@link RecommendationsXEvents}
  * * {@link RelatedTagsXEvents}
- * * {@link SearchXEvents},
+ * * {@link ScrollXEvents},
  * * {@link SearchBoxXEvents}
+ * * {@link SearchXEvents}
  * * {@link UrlXEvents}
  *
  * @public
  */
 export interface XEventsTypes
-  extends EmpathizeXEvents,
+  extends DeviceXEvents,
+    EmpathizeXEvents,
     ExtraParamsXEvents,
     FacetsXEvents,
     HistoryQueriesXEvents,
@@ -54,8 +56,7 @@ export interface XEventsTypes
     ScrollXEvents,
     SearchBoxXEvents,
     SearchXEvents,
-    UrlXEvents,
-    DeviceXEvents {
+    UrlXEvents {
   /**
    * The search adapter configuration has changed
    * * Payload: The new search adapter configuration.
--- a/packages/x-components/src/wiring/events.types.ts
+++ b/packages/x-components/src/wiring/events.types.ts
@@ -1,5 +1,4 @@
 import { Result, Suggestion } from '@empathyco/x-types';
-import { ScrollDirection } from '../components/scroll/scroll.types';
 import { ArrowKey, PropsWithType } from '../utils';
 import { DeviceXEvents } from '../x-modules/device';
 import { EmpathizeXEvents } from '../x-modules/empathize/events.types';
@@ -12,6 +11,7 @@ import { PopularSearchesXEvents } from '../x-modules/popular-searches/events.typ
 import { QuerySuggestionsXEvents } from '../x-modules/query-suggestions/events.types';
 import { RecommendationsXEvents } from '../x-modules/recommendations/events.types';
 import { RelatedTagsXEvents } from '../x-modules/related-tags/events.types';
+import { ScrollXEvents } from '../x-modules/scroll/events.types';
 import { SearchBoxXEvents } from '../x-modules/search-box/events.types';
 import { SearchXEvents } from '../x-modules/search/events.types';
 import { UrlXEvents } from '../x-modules/url/events.types';
@@ -41,8 +41,7 @@ import { UrlXEvents } from '../x-modules/url/events.types';
  * @public
  */
 export interface XEventsTypes
-  extends DeviceXEvents,
-    EmpathizeXEvents,
+  extends EmpathizeXEvents,
     ExtraParamsXEvents,
     FacetsXEvents,
     HistoryQueriesXEvents,
@@ -52,9 +51,11 @@ export interface XEventsTypes
     QuerySuggestionsXEvents,
     RecommendationsXEvents,
     RelatedTagsXEvents,
-    SearchXEvents,
+    ScrollXEvents,
     SearchBoxXEvents,
-    UrlXEvents {
+    SearchXEvents,
+    UrlXEvents,
+    DeviceXEvents {
   /**
    * The search adapter configuration has changed
    * * Payload: The new search adapter configuration.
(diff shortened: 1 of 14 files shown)


packages/x-components/src/x-modules/scroll/store/index.ts
```diff
@@ -0,0 +1,4 @@
+export * from './emitters';
+export * from './module';
+export * from './types';
+export { initScrollComponentState } from './utils';
```
pmareke, line 4: Should we export * just in case we add more utils in the future?

packages/x-components/src/x-modules/scroll/store/types.ts
```diff
@@ -0,0 +1,159 @@
...
+/**
+ * Contains all the state of a scroll component.
+ *
+ * @public
+ */
+export interface ScrollComponentState {
+  /**
+   * The position in pixels the user has scrolled down.
+   */
+  position: number;
```
pmareke, line 30: Maybe `positionInPx`?

packages/x-components/src/x-modules/scroll/store/types.ts
```diff
@@ -0,0 +1,159 @@
...
+ * @public
+ */
+export interface ScrollMutations {
+  /**
+   * Sets the identifier of the element that is pending to be scrolled into the view.
+   *
+   * @param pendingScrollTo - The identifier of the element that should be scrolled into view
+   * whenever it is loaded.
+   */
+  setPendingScrollTo(pendingScrollTo: string): void;
```
pmareke, line 64: I would prefer to use `setPendingScrollTo(elementId: string): void;`

packages/x-components/src/x-modules/scroll/store/types.ts
```diff
@@ -0,0 +1,159 @@
...
+ * it has reached certain position.
+ *
+ * @public
+ */
+export interface ScrollPositionReachedPayload {
+  /**
+   * True if it has reached certain position. False otherwise.
+   */
+  value: boolean;
+  /**
```
pmareke, line 136: Can we create a basic interface with this `id` and use it in the other 4 interfaces?

packages/x-components/src/x-modules/scroll/store/utils.ts
```diff
@@ -0,0 +1,25 @@
+import Vue from 'vue';
+import { Dictionary } from '../../../utils/types';
+import { ScrollComponentState } from './types';
+
+/**
+ * Initialises an {@link ScrollComponentState} object if it has not already been created.
+ *
+ * @param state - The state shere the new data should be added.
```
pmareke, line 8: `shere `?

packages/x-components/src/wiring/events.types.ts
```diff
@@ -52,9 +51,11 @@ export interface XEventsTypes
     QuerySuggestionsXEvents,
     RecommendationsXEvents,
     RelatedTagsXEvents,
-    SearchXEvents,
+    ScrollXEvents,
     SearchBoxXEvents,
-    UrlXEvents {
+    SearchXEvents,
+    UrlXEvents,
+    DeviceXEvents {
```
LuisMartinez15, line 58: You reordered this alphabetically but moved `DeviceXEvents` to the end, why? 🤷🏿‍♂️

pmareke, line 58: If this `XEvents` are order from A to Z, this one should be the first one?

pmareke (changes requested):

packages/x-components/src/x-modules/scroll/store/index.ts
LuisMartinez15, line 4: I know the content of this file is just this function, but, what's the reason behind only exporting it instead of whole file? If we add more utilities we would need to refactor this to be able to use them outside our library.

In case it was unintended, I´ll leave the suggestion here:

```suggestion
export * from './utils';
```

packages/x-components/src/x-modules/scroll/store/types.ts
LuisMartinez15, line 64: I´d change this param name to `scrollId`.

```suggestion
   * @param scrollId - The identifier of the element that should be scrolled into view
   * whenever it is loaded.
   */
  setPendingScrollTo(scrollId: string): void;
```

packages/x-components/src/x-modules/scroll/store/utils.ts
```diff
@@ -0,0 +1,25 @@
...
+ * @param id - The identifier for the {@link ScrollComponentState}.
+ *
+ * @internal
+ */
+export function initScrollComponentState(
+  state: Dictionary<ScrollComponentState>,
+  id: string
+): void {
+  if (!state[id]) {
+    Vue.set<ScrollComponentState>(state, id, {
```
LuisMartinez15, line 18: I'd change this param name to `componentsState` to avoid confusions with the "whole" `store` `state` of the module.

```suggestion
 * Initialises an {@link ScrollComponentState} object if it has not already been created.
 *
 * @param componentsState - The componentsState where the new data should be added.
 * @param id - The identifier for the {@link ScrollComponentState}.
 *
 * @internal
 */
export function initScrollComponentState(
  componentsState: Dictionary<ScrollComponentState>,
  id: string
): void {
  if (!componentsState[id]) {
    Vue.set<ScrollComponentState>(componentsState, id, {
```

packages/x-components/src/x-modules/scroll/wiring.ts
```diff
@@ -0,0 +1,106 @@
...
+    id: metadata.id as string
+  })
+);
+
+/**
+ * Saves a boolean indicating if the scroll has reached the end of a container to the store.
+ *
+ * @public
+ */
+export const setScrollReachedEndWire = wireCommit(
```
LuisMartinez15, line 41:
```suggestion
export const setScrollHasReachedEndWire = wireCommit(
```

packages/x-components/src/x-modules/scroll/wiring.ts
```diff
@@ -0,0 +1,106 @@
...
+    setScrollPositionWire
+  },
+  UserChangedScrollDirection: {
+    setScrollDirectionWire
+  },
+  UserReachedScrollStart: {
+    setScrollHasReachedStartWire
+  },
+  UserReachedScrollEnd: {
+    setScrollReachedEndWire
```
LuisMartinez15, line 95:
```suggestion
    setScrollHasReachedEndWire
```

LuisMartinez15 (changes requested):

packages/x-components/src/x-modules/scroll/store/types.ts
javieri-empathy: I don't think we need to be that specific here, but if @LuisMartinez15 also tells me to add it I won't fight you both. In any case, it is `positionPx`, no `positionInPx`

packages/x-components/src/x-modules/scroll/store/types.ts
```diff
@@ -0,0 +1,159 @@
...
+   */
+  pendingScrollTo: string;
+}
+
+/**
+ * Contains all the state of a scroll component.
+ *
+ * @public
+ */
+export interface ScrollComponentState {
```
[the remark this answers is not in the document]
pmareke: I like it.

LuisMartinez15 (approved):

pmareke (approved):

PR #220, opened 2021-11-12
feat: Make `Identifiable` type generic

javieri-empathy:
Just makes the Identifiable type generic so we can use it with another ID types

### DIFF
--- a/packages/search-types/src/identifiable.model.ts
+++ b/packages/search-types/src/identifiable.model.ts
@@ -3,7 +3,7 @@
  *
  * @public
  */
-export interface Identifiable {
+export interface Identifiable<ID = string | number> {
   /** A unique ID that identifies the Object. */
-  id: string | number;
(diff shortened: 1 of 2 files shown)


pmareke (approved):

packages/search-types/src/identifiable.model.ts
```diff
@@ -3,7 +3,7 @@
  *
  * @public
  */
-export interface Identifiable {
+export interface Identifiable<ID = string | number> {
```
pmareke, line 6: I would use T instead of ID, but you're the master here.

pmareke (approved):

LuisMartinez15 (approved):

CachedaCodes (approved):

LuisMartinez15 (approved):

pmareke (approved):

PR #221, opened 2021-11-12
Feature/ex 4997 move scroll components

javieri-empathy:
This PR moves the scroll components to its own module.

The `base-scroll` and `scroll-mixin` serve as utilities to build more advanced scroll components. This components remain in `src/components`, the do not depend on any module.

The rest of te scroll components have been moved to the `scroll` module, and adjusted to work with it.

I've also simplified store management. In the end if we assume `ComponentScrollState` is an entity we can allow modifying it in any way, we do not need 5 individual mutations to do so. Wires were already mapping the value, so there is no big change here either.

❌Be careful, this PR has breaking changes ❌

> feat(scroll): Move scroll related components to the `scroll` X Module.
>
> BREAKING-CHANGE: `BaseIdScroll` has been renamed to `Scroll`. It should be imported from `@empathyco/x-components/scroll`
> BREAKING-CHANGE: `BaseMainScroll` has been renamed to `WindowScroll`. It should be imported from `@empathyco/x-components/scroll`
> BREAKING-CHANGE: `BaseScrollToTop` has been renamed to `ScrollToTop`. It should be imported from `@empathyco/x-components/scroll`
> BREAKING-CHANGE: `UserReachedScrollStart` event now emits a boolean indicating the change of this state.
> BREAKING-CHANGE: `UserAlmostReachedScrollEnd` event now emits a boolean indicating the change of this state.
> BREAKING-CHANGE: `UserReachedScrollEnd` event now emits a boolean indicating the change of this state.
> 
> EX-4997


packages/x-components/src/components/layouts/fixed-header-and-asides-layout.vue
```diff
@@ -95,7 +95,7 @@
   @Component({
     components: {
       BaseIdModal,
-      BaseIdScroll
+      Scroll
```
javieri-empathy, line 98: Importing a component from a module in a non module component shouldn't be allowed. But I think we can discuss the solution to this problem and move it into another PR. One solution would be to allow dynamic root components in our layouts sections.

pmareke:
Please change the Scroll To Top in the Layout view.

packages/x-components/src/x-modules/scroll/components/__tests__/scroll-to-top.spec.ts
```diff
@@ -1,60 +1,63 @@
...
+    const { scrollToTopWrapper, emitXEvent } = renderScrollToTop();
+    await emitXEvent('UserAlmostReachedScrollEnd', true);
+    await emitXEvent('UserChangedScrollDirection', 'DOWN');
     expect(scrollToTopWrapper.text()).toEqual('Top');
   });
 
   it('shows if a scroll is almost reaching the end and there is no threshold', async () => {
-    const { scrollToTopWrapper, emitXEvent } = renderBaseScrollToTop();
+    const { scrollToTopWrapper, emitXEvent } = renderScrollToTop();
     expect(scrollToTopWrapper.html()).toBe('');
```
pmareke, line 53: Is confusing, at least for me, that when the component is not visible it's rendered.

I mean, I would expect to test something like `expect(scrollToTopWrapper).toBeUndefined();` instead of an empty string.

packages/x-components/src/x-modules/scroll/components/__tests__/scroll.spec.ts
```diff
@@ -29,37 +31,41 @@ function renderBaseIdScroll({
...
+      jest.advanceTimersByTime(durationMs);
       await wrapper.vm.$nextTick();
     }
   };
 }
 
-describe('testing Base Scroll Component', () => {
+describe('testing Scroll Component', () => {
   beforeAll(jest.useFakeTimers);
   afterEach(jest.clearAllTimers);
```
pmareke, line 59: Same here as the other comment, if you remove this afterEach everything works the same.

packages/x-components/src/x-modules/scroll/components/__tests__/scroll.spec.ts
```diff
@@ -29,37 +31,41 @@ function renderBaseIdScroll({
...
+    const { wrapper } = await renderScroll({
       template: `
-        <BaseIdScroll :id="id">
+        <Scroll :id="id">
           <div data-test="content-scroll">
             <p>scroll content</p>
           </div>
-        </BaseIdScroll>`,
+        </Scroll>`,
       id: 'scrollResults'
```
pmareke, line 69: This `id` parameter already has a `scrollResults` value.

packages/x-components/src/x-modules/scroll/components/__tests__/scroll.spec.ts
```diff
@@ -68,14 +74,14 @@ describe('testing Base Scroll Component', () => {
   });
 
   it('throttles the scroll event', async () => {
-    const { wrapper, scroll, scrollElement } = renderBaseIdScroll({
+    const { wrapper, scroll, scrollElement } = await renderScroll({
       throttleMs: 200,
       id: 'scrollResults'
```
pmareke, line 79: Same here.

packages/x-components/src/x-modules/scroll/components/__tests__/window-scroll.spec.ts
```diff
@@ -32,28 +47,23 @@ function renderBaseMainScroll({
...
     },
     async scrollBody(options: { to: number; durationMs: number }) {
       body.scrollTop = options.to;
       body.dispatchEvent(new Event('scroll'));
       jest.advanceTimersByTime(options.durationMs);
-      await wrapperContainer.vm.$nextTick();
+      await localVue.nextTick();
     },
     async scrollContent(options: { to: number; durationMs: number }) {
       const scrollContent = wrapperContainer.find(getDataTestSelector('content')).element;
```
pmareke, line 69: Review the method as the following code it's not longer needed.

```javascript
jest.spyOn(scrollContent, 'clientHeight', 'get').mockImplementation(() => 700);
jest.spyOn(scrollContent, 'scrollHeight', 'get').mockImplementation(() => 150);
```

packages/x-components/src/x-modules/scroll/components/__tests__/window-scroll.spec.ts
```diff
@@ -78,14 +88,14 @@ describe('testing Main Scroll Component', () => {
```
pmareke, line 88: Same here with the `afterEach(jest.clearAllTimers);`

packages/x-components/src/x-modules/scroll/components/__tests__/window-scroll.spec.ts
```diff
@@ -1,22 +1,37 @@
...
 
-function renderBaseMainScroll({
-  template = `<BaseMainScroll :throttleMs="throttleMs" :tag="tag"/>`,
+async function renderWindowScroll({
+  template = `<WindowScroll :throttleMs="throttleMs" :tag="tag"/>`,
   throttleMs = 200,
   scrollHeight = 800,
   clientHeight = 200,
   distanceToBottom = 100,
   tag,
```
pmareke, line 13: You can use here a default `body` or `html` and remove the half of `tag` in the calls to the `renderBaseMainScroll` function.

packages/x-components/src/x-modules/scroll/components/scroll-to-top.vue
```diff
@@ -15,12 +15,14 @@
...
+  import { State, xComponentMixin } from '../../../components';
+  import BaseEventButton from '../../../components/base-event-button.vue';
+  import { NoElement } from '../../../components/no-element';
+  import { Dictionary } from '../../../utils';
+  import { XEventsTypes } from '../../../wiring';
+  import { ScrollComponentState } from '../store';
+  import { scrollXModule } from '../x-module';
 
   /**
    * Renders a button with a default slot.
```
pmareke, line 28: `Renders a button with a default slot and an animation.` ?

packages/x-components/src/x-modules/scroll/components/scroll-to-top.vue
```diff
@@ -107,7 +117,7 @@
      * @internal
      */
     protected get isThresholdReached(): boolean {
-      return this.useThresholdStrategy && this.scrollTop > (this.thresholdPx as number);
+      return this.useThresholdStrategy && this.scrollData.position > this.thresholdPx!;
```
pmareke, line 120: This `useThresholdStrategy` computed property name is confusing isn't it? Maybe something like `isThresholdANumber`

packages/x-components/src/x-modules/scroll/components/scroll.vue
```diff
@@ -60,23 +62,14 @@
...
     public id!: string;
 
-    /**
-     * If true (default), sets the scroll position of the {@link BaseScroll} component to top when
-     * an {@link XEventsTypes.UserAcceptedAQuery} event is emitted.
-     *
-     * @public
-     */
-    @Prop({ default: true })
-    protected resetOnQueryChange!: boolean;
```
pmareke: What happened with this prop? If we're going to remove it please remove it also in the `BaseScroll` and `Scroll` doc examples

packages/x-components/src/x-modules/scroll/components/window-scroll.vue
```diff
@@ -33,135 +33,78 @@
...
-      this.$x.emit(event, payload as any, { target: this.element, id: this.id });
-    }
-
-    // eslint-disable-next-line @typescript-eslint/no-empty-function
-    render(): void {}
-
-    beforeDestroy(): void {
-      this.elementListener.removeEventListener('scroll', this.throttledStoreScrollData);
+    protected createXEventMetadata(): Partial<WireMetadata> {
+      return { target: this.$el, id: this.id };
```
pmareke, line 88: This is the same as the createEventMetadata from the `Scroll` component right? Can we extract to the utils folder?

If not, please add the type `as HTMLElement` to the `$el`.

packages/x-components/src/x-modules/scroll/store/module.ts
```diff
@@ -13,21 +13,17 @@ export const scrollXStoreModule: ScrollXStoreModule = {
...
-      initScrollComponentState(state.data, id);
-      state.data[id].hasReachedEnd = value;
-    },
-    setScrollHasReachedStart(state, { id, value }) {
-      initScrollComponentState(state.data, id);
-      state.data[id].hasReachedStart = value;
+    setScrollComponentState(state, { id, newState }) {
+      if (!state.data[id]) {
+        Vue.set<ScrollComponentState>(state.data, id, {
+          hasReachedStart: false,
```
pmareke, line 19: I think we can create a helper method called `createScrollInitialState`, or something like that, and move this to the utils folder.

packages/x-components/src/x-modules/scroll/wiring.ts
```diff
@@ -26,9 +26,22 @@ export const setScrollPositionWire = wireCommit(
  * @public
  */
 export const setScrollDirectionWire = wireCommit(
-  'setScrollDirection',
+  'setScrollComponentState',
   ({ metadata, eventPayload }: WirePayload<ScrollDirection>) => ({
-    direction: eventPayload,
+    newState: { direction: eventPayload },
+    id: metadata.id as string
```
pmareke, line 32: Can you use `id: metadata.id!` here? At this point you can asume that you're going to have and id right?

pmareke (changes requested):
Amazing job mate, acojonante.

packages/x-components/src/components/layouts/fixed-header-and-asides-layout.vue
pmareke: I think in this case in particularly we're always going to have scroll right? It's for me but thanks for pointing out.

packages/x-components/src/x-modules/scroll/components/scroll-to-top.vue
LuisMartinez15: But there is not default animation, right? You have to pass in the animation.

packages/x-components/src/x-modules/scroll/components/window-scroll.vue
```diff
@@ -0,0 +1,227 @@
...
+   * Main scroll component that depending on the user interactivity emits different events for
+   * knowing when the user scrolls, the direction of scroll and if user reaches the start or end.
+   * All this about main scroll.
+   *
+   * @public
+   */
+  @Component({
+    mixins: [xComponentMixin(scrollXModule)]
+  })
+  export default class WindowScroll extends mixins(ScrollMixin) {
```
LuisMartinez15, line 20: Why is this component called `WindowScroll` instead of `MainScroll` if the documentation says Main scroll component? just wondering 😅

packages/x-components/src/x-modules/scroll/components/window-scroll.vue
LuisMartinez15: +1 for the type

packages/x-components/src/x-modules/scroll/store/index.ts
```diff
@@ -1,4 +1,3 @@
 export * from './emitters';
 export * from './module';
 export * from './types';
-export * from './utils';
```
LuisMartinez15, line 4: RIP (15/11-15/11)

packages/x-components/src/x-modules/scroll/store/module.ts
LuisMartinez15: I think he did the opposite here 😅 
This was previously in a utility and he removed it and did it here. I wouldn't move it back to the utils folder as this is the only place where this is being used.

LuisMartinez15 (changes requested):

packages/x-components/src/components/layouts/fixed-header-and-asides-layout.vue
javieri-empathy: The problem is that even if you don't use our layouts you will still have the scroll module loaded.

packages/x-components/src/components/scroll/__tests__/base-scroll.spec.ts
```diff
@@ -60,8 +62,8 @@ describe('testing Base Scroll Component', () => {
   beforeAll(jest.useFakeTimers);
```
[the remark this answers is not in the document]
javieri-empathy: That is just to cleanup in case there are any pending timers. We are using fake timers (check line 55 `jest.advanceTimersByTime(durationMs)`.

packages/x-components/src/x-modules/scroll/components/__tests__/scroll-to-top.spec.ts
```diff
@@ -1,60 +1,63 @@
...
-  const scrollToTopWrapper = wrapper.findComponent(BaseScrollToTop);
+  const scrollToTopWrapper = wrapper.findComponent(ScrollToTop);
 
   return {
     scrollToTopWrapper,
-    async click() {
+    click() {
       scrollToTopWrapper.trigger('click');
-      await localVue.nextTick();
+      return localVue.nextTick();
```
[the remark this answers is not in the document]
javieri-empathy: Right, I forgot the second approach, it is easier.

packages/x-components/src/x-modules/scroll/components/__tests__/scroll-to-top.spec.ts
```diff
@@ -1,60 +1,63 @@
...
     },
-    async emitXEvent<Event extends XEvent>(event: Event, payload: XEventPayload<Event>) {
```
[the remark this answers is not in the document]
javieri-empathy: Both approaches are doing the same. `async` / `await` is almost all syntactic sugar over the promises. Why do I want to do an `await` of a promise when I can just return it?

packages/x-components/src/x-modules/scroll/components/__tests__/scroll-to-top.spec.ts
javieri-empathy: I don't think this confusing. Maybe a `expect(xxx).toRenderNothing()` would be more readable, but reading that a component is rendering no HTML is fine for me. In any case, that's Vue test utils problem, which forces you to test it like this. And you can't use `expect(scrollToTopWrapper.element).toBeUndefined()` because the element is actually a comment node, so it is not undefined.

packages/x-components/src/x-modules/scroll/components/__tests__/scroll-to-top.spec.ts
```diff
@@ -1,60 +1,63 @@
...
-    const { scrollToTopWrapper, emitXEvent } = renderBaseScrollToTop();
+    const { scrollToTopWrapper, emitXEvent } = renderScrollToTop();
     expect(scrollToTopWrapper.html()).toBe('');
-    await emitXEvent('UserAlmostReachedScrollEnd', 100);
+    await emitXEvent('UserAlmostReachedScrollEnd', true);
+    await emitXEvent('UserChangedScrollDirection', 'DOWN');
     expect(scrollToTopWrapper.html()).not.toBe('');
   });
 
   it('shows if the scroll is over the threshold and user scrolls up', async () => {
```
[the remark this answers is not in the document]
javieri-empathy: Why do you say it doesn't render at all? IT renders a button, with an span...

packages/x-components/src/x-modules/scroll/components/__tests__/scroll-to-top.spec.ts
```diff
@@ -77,18 +81,18 @@ describe('testing Base Scroll To Top component', () => {
 
   // eslint-disable-next-line max-len
   it('hides when the scroll direction is up once the scroll has almost reached the end', async () => {
-    const { scrollToTopWrapper, emitXEvent } = renderBaseScrollToTop();
+    const { scrollToTopWrapper, emitXEvent } = renderScrollToTop();
 
     await emitXEvent('UserChangedScrollDirection', 'DOWN');
-    await emitXEvent('UserAlmostReachedScrollEnd', 100);
+    await emitXEvent('UserAlmostReachedScrollEnd', true);
     expect(scrollToTopWrapper.html()).not.toBe('');
```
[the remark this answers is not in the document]
javieri-empathy: If you pass the `Top` string explicitly to the `renderBaseScrollToTop` function I would do it too, but as it is not passed, knowing that it is rendering something it is enough.

packages/x-components/src/x-modules/scroll/components/__tests__/scroll.spec.ts
javieri-empathy: Same answer, this is just to ensure that each test is executed isolated and there are no pending calls from other tests.

packages/x-components/src/x-modules/scroll/components/__tests__/scroll.spec.ts
javieri-empathy: Changed it. It is all about being explicit

packages/x-components/src/x-modules/scroll/components/scroll-to-top.vue
javieri-empathy: I agree with @LuisMartinez15 The animation optional. However I have simplified the documentation, too much detail.

packages/x-components/src/x-modules/scroll/components/scroll-to-top.vue
javieri-empathy: I don't think so. This component has two ways of working, rendering itself when a certain threshold is reached, or when the user is about the end the scroll and keeps scrolling down. I don't even know if this last way is necessary, as it does not work properly with the infinite scroll, maybe it only makes sense for pagination. 

The `useThresholdStrategy` name exactly tells this, if the component should render after a threshold. `isThresholdANumber` says nothing about the intention for me.

packages/x-components/src/x-modules/scroll/components/window-scroll.vue
javieri-empathy: Because the documentation is written in chino mandarin and it wasn't affected by the global replace I did. Fixed it. Thanks

packages/x-components/src/x-modules/scroll/components/window-scroll.vue
javieri-empathy: I don't think it makes sense to make any of this. 
The `this.$el` is already typed thanks to the `ScrollMixin` component. In the `Scroll.vue` component, it has to be casted, as it does not use the `ScrollMixin`.

I don't see any value in the utility, appart from other layer of abstraction. If this was more complex code with decisions, loops it might make more sense, but for creating a typed object with 2 properties? Unnecessary FMPOV.

packages/x-components/src/x-modules/scroll/store/module.ts
javieri-empathy: I did exactly the opposite yeah, as it is now just one `if` and one statement, I think it is easier to keep it here. 

I had wrote a comment here in the PR explaining this, but forgot to send it 😅.

packages/x-components/src/x-modules/scroll/components/scroll.vue
pmareke: Remove it in the BaseScroll docs also please.

packages/x-components/src/x-modules/scroll/components/__tests__/window-scroll.spec.ts
```diff
@@ -522,42 +601,63 @@ describe('testing Main Scroll Component', () => {
...
+          moduleName: 'scroll',
           target: document.body,
           id: 'main-scroll'
         }
       });
     });
 
     it('does not trigger event when scrolling in document or other element', async () => {
-      const { wrapperContainer, scrollHtml, scrollContent } = renderBaseMainScroll({
+      const { wrapper, scrollHtml, scrollContent } = await renderWindowScroll({
```
javieri-empathy, line 649: Should this test be removed? I feel like it is testing a native browser behavior (the scroll event does not bubble), and it is doing so by mocking all the JSDOM elements in the test with random values....

pmareke: I like the validation that our scrolls are not scrolling, for me is ok to leave it but I'll buy you if you prefer to delete it.

pmareke (approved):

packages/x-components/src/components/scroll/__tests__/base-scroll.spec.ts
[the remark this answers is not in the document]
LuisMartinez15: I think we said the same, but got downvoted :feelsbadman:

packages/x-components/src/x-modules/scroll/components/__tests__/window-scroll.spec.ts
LuisMartinez15: We can delete it if it doesn't provide any value. We can check it later after the whole module's rework or in another pr, as I'm going to merge this one.

LuisMartinez15 (approved):

PR #224, opened 2021-11-16
feat(scroll):  add `MainScroll` and `MainScrollItem` components

javieri-empathy:
[EX-4998](https://searchbroker.atlassian.net/browse/EX-4998)

# Description

This PR enables us to have support to store scroll in the URL.

- Adds `MainScroll` component. This component injects a custom intersection observer that will record the elements that are currently visible. Its bounds and threshold can be configured.
- Adds `MainScrollItem` component. This component wraps other components (like the result ones) that should be used to store/restore the scroll in the URL. It is the responsible for registering itself into the injected observer.
- Modifies the store module with proper types. We are now storing an identifier for the scroll, not the scroll position.

## Updating the URL
1. `MainScroll` component creates and provides an intersection observer with the given parameters. This intersection observer will help to keep a list of the visible elements.
2. `MainScrollItem` register themselves into this observer.
3.  A computed property in `MainScroll` selects the first visible element.
4.  When the first visible element changes, the `UserScrolledToElement` event is emitted.

## Restoring from the URL
1. `URLParams` event is emitted initially with a `scroll` property
2. The  `scroll` module updates its `pendingScrollTo` property.
3. As soon as the `MainScroll` component is rendered, each  `MainScrollItem` detects if it is the one that should be scrolled into.
4. When a `MainScrollItem` detects that it is the one that should be scrolled into the view, it scrolls itself, and emits the `ScrollRestoreSucceeded` event. If no element emits this event after a defined time, the `ScrollRestoreFailed` is emitted instead. Currently both events do the same, they are just differentiated for debugging purposes.

# Known issues

- Scroll is restored a few pixels wrong due to entering transitions.


package.json
```diff
@@ -6,7 +6,7 @@
     "postci": "lerna bootstrap --hoist --ci && npm run build",
     "build": "lerna run build",
     "lint": "eslint packages --cache",
-    "lint:fix": "eslint packages --cache --fix",
```
javieri-empathy, line 9: eslint cache is slowing down webstorm, making it even freeze for long times.

tajespasarela (changes requested):
The subscription to `UserClicedScrollToTop` should be in the scroll mixin not in `Scroll.vue`

packages/x-components/src/x-modules/scroll/components/main-scroll.vue
```diff
@@ -0,0 +1,403 @@
...
+### Play with props
+
+#### Window scroll
+
+In case you aren't using a custom scrolling element like the `Scroll` panel, and want to use the
+default browser scroll, you can do so by using the `use-window-scroll` prop:
+
+```vue
+<template>
+  <MainScroll use-window>
```
tajespasarela, line 321: use camel case for props

packages/x-components/src/x-modules/scroll/components/main-scroll.vue
```diff
@@ -0,0 +1,403 @@
...
+restore the scroll to the provided position (if available). Because of the dynamic nature of
+JavaScript applications, the element that it tries to restore the scroll to might not be available
+anymore. For this reason after a defined time, the scroll restoration will be considered failed.
+
+This time can be configured through the `restore-scroll-timeout-ms` prop. This is specially useful
+when combined with the URL X Module.
+
+```vue
+<template>
+  <MainScroll :restore-scroll-timeout-ms="1000">
```
tajespasarela, line 353: use camel case for props

packages/x-components/src/x-modules/scroll/components/main-scroll.vue
```diff
@@ -0,0 +1,403 @@
...
+    mixins: [xComponentMixin(scrollXModule)]
+  })
+  /* eslint-disable @typescript-eslint/unbound-method */
+  export default class MainScroll extends NoElement {
+    /**
+     * The reference to the HTML node that has the scroll.
+     *
+     * @public
+     */
+    public $el!: HTMLElement;
```
tajespasarela, line 31: Is this attribute really needed here? `NoElement` is a Vue instance with `$el` defined

packages/x-components/src/x-modules/scroll/components/main-scroll.vue
```diff
@@ -0,0 +1,403 @@
...
+    /**
+     * If true (default), sets the scroll position to top when an
+     * {@link XEventsTypes.UserAcceptedAQuery} event is emitted.
+     *
+     * @public
+     */
+    @Prop({
+      type: Boolean,
+      default: true
+    })
```
tajespasarela, line 77:
```suggestion
    @Prop({ type: Boolean, default: true })
```

packages/x-components/src/x-modules/scroll/components/main-scroll.vue
```diff
@@ -0,0 +1,403 @@
...
+    protected intersectionObserver: IntersectionObserver | null = null;
+
+    /**
+     * Creates an `IntersectionObserver` to detect the first visible elements. Children of this
+     * component should register themselves if they want to be observed.
+     *
+     * @returns The intersection observer.
+     * @public
+     */
+    @XProvide(FirstVisibleItemObserverKey)
```
tajespasarela, line 118: this key confused me twice already reviewing the PR. Wouldn't be clearer something like `MainScrollObserver` or just `ScrollObserver`?
we already know what is for

tajespasarela (changes requested):

packages/x-components/src/components/layouts/multi-column-max-width-layout.vue
```diff
@@ -69,14 +69,16 @@
...
 
-      <Scroll v-if="hasContent('main-body')" id="main-scroll" class="x-layout__body-scroll">
-        <section class="x-layout__main-body x-list x-list--vertical">
-          <!-- @slot Slot that can be used to insert the body content. -->
-          <slot name="main-body">
-            <span v-if="devMode" class="slot-helper" style="height: 110vh">MAIN BODY</span>
-          </slot>
-        </section>
-      </Scroll>
+      <MainScroll>
```
tajespasarela, line 72: The other layout components are jealous, they deserve a `MainScroll` too

tajespasarela (changes requested):

packages/x-components/src/x-modules/scroll/components/main-scroll.vue
javieri-empathy: Vue `$el` has type `Element`, I'm narrowing that type to a more concrete one to avoid the `as HTMLElement` casts everywhere.

packages/x-components/src/x-modules/scroll/components/main-scroll.vue
javieri-empathy: I thought that prettier took care of this, but it respects both ways... weird.

packages/x-components/src/x-modules/scroll/components/main-scroll-item.vue
```diff
@@ -0,0 +1,166 @@
...
+
+  /**
+   * Wrapper for elements contained in the {@link MainScroll} that should store/restore its
+   * position.
+   *
+   * @public
+   */
+  @Component({
+    mixins: [xComponentMixin(scrollXModule)],
+    components: { UrlHandler, MainScroll, NoElement }
```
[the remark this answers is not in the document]
javieri-empathy: Because if you are not careful when writing the doc, webstorm auto-imports them in the component T.T. If I recall correctly Vue's eslint rules detects unused components in its last version.

packages/x-components/src/components/layouts/multi-column-max-width-layout.vue
javieri-empathy: I was asking in the other PR if we should treat the scroll as a special case and allow it to be used inside layouts. This makes that the `scroll` module is always present, even if you don't use our layouts or a scroll component, until the tree-shaking step happens and removes the unused code, which will only happen in production builds. 

To be honest I don't like this, I would like to fix this in another task. These are some options that came to my mind:

- Leave the layouts without the configured scroll.
- Move the layouts to its own independent module that would depend on the scroll module.
- Make the current HTMLElements in the layout configurable, so you can pass the component you want to render using a prop.

packages/x-components/src/x-modules/scroll/components/main-scroll.vue
```diff
@@ -0,0 +1,400 @@
...
+
+    /**
+     * Emits the first visible element `[data-scroll]`.
+     *
+     * @param element - The first visible element or `null` if there aren't any.
+     * @internal
+     */
+    @Watch('firstVisibleElement')
+    protected emitFirstVisibleElement(element: HTMLElement | null): void {
+      this.$x.emit('UserScrolledToElement', element?.dataset.scroll ?? '');
```
tajespasarela, line 177: Is it possible to check if the scroll element is at the start (scrollTop===0) and avoid to emit the event. Also clear the current scroll from URL when scroll top is reached?

Because when you search, the first id is set on URL even when the scrollTop is 0. It would be great to avoid that.

tajespasarela: right now if you search, and reload, we have the issue of the animation and the grid is scrolled a little. We can try to avoid at least if the scroll is on top

packages/x-components/tests/unit/main-scroll.spec.ts
```diff
@@ -0,0 +1,178 @@
...
+
+  mount({
+    components: {
+      Scroll,
+      MainScroll,
+      MainScrollItem
+    },
+    template: `
+      <MainScroll style="height:200px" v-bind="{ threshold, margin }">
+      <Scroll>
```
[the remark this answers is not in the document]
tajespasarela: If it is for the styles, we can add an element with inline styles.

packages/x-components/src/components/layouts/multi-column-max-width-layout.vue
tajespasarela: We leave changed at the moment, and we can refactor later in other PR, but we need the current layouts working with URL scroll

tajespasarela (changes requested):

packages/x-components/src/x-modules/scroll/components/main-scroll.vue
```diff
@@ -0,0 +1,400 @@
...
+     */
+    @Prop({ default: 5000 })
+    public restoreScrollTimeoutMs!: number;
+
+    /**
+     * Intersection percentage to consider an element visible.
+     *
+     * @public
+     */
+    @Prop({ default: 0.5 })
```
tajespasarela, line 57: reduce the default value to 0.3 to minimize the animation issue, add a TODO if you want to remember restore it when animation issue fixed

tajespasarela (changes requested):

packages/x-components/src/x-modules/scroll/components/main-scroll.vue
```diff
@@ -0,0 +1,400 @@
...
+     * The first visible element contained in this component.
+     *
+     * @returns The first visible element in this component.
+     * @internal
+     */
+    protected get firstVisibleElement(): HTMLElement | null {
+      return this.intersectingElements.length === 0
+        ? null
+        : this.intersectingElements.reduce((firstVisibleElement, anotherElement) => {
+            // FIXME: This algorithm only takes into account LTR layouts
```
guillei10, line 139: As far as I understood, adapting to RTL layouts will only require to change the `left` to `right` in the `anotherElementBounds.left <= firstVisibleElementBounds.left` depending on the value from the snippet.

packages/x-components/src/x-modules/scroll/components/main-scroll.vue
```diff
@@ -0,0 +1,400 @@
...
+  }
+</style>
+```
+
+### Play with props
+
+#### Window scroll
+
+In case you aren't using a custom scrolling element like the `Scroll` panel, and want to use the
+default browser scroll, you can do so by using the `use-window-scroll` prop:
```
guillei10, line 314: `useWindow` prop, isn't it?

packages/x-components/tests/unit/main-scroll.spec.ts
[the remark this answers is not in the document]
guillei10: and fix identation.... TOC things

packages/x-components/tests/unit/main-scroll.spec.ts
```diff
@@ -0,0 +1,178 @@
...
+        expect(item4Bottom).to.be.lte($scroll.position().top);
+      });
+    });
+  });
+
+  it('allows configuring when to consider an element visible', () => {
+    const { scrollToItem, userScrolledToElementSpy } = renderMainScroll({ threshold: 1 });
+
+    scrollToItem(5);
+    /* The 5th element should be top aligned now. Because of the thrshold=1, it still is the first
```
guillei10, line 104: threshold*

guillei10 (changes requested):

packages/x-components/tests/unit/main-scroll.spec.ts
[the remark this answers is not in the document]
javieri-empathy: you are right in everything.

packages/x-components/src/x-modules/scroll/components/main-scroll.vue
```diff
@@ -0,0 +1,400 @@
...
+     * Adjusts the size of the scroll container bounds.
+     *
+     * @public
+     */
+    @Prop({ default: '' })
+    public margin!: string;
+
+    /**
+     * If true (default), sets the scroll position to top when an
+     * {@link XEventsTypes.UserAcceptedAQuery} event is emitted.
```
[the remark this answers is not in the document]
javieri-empathy: We need to refactor this, we can't add every event that can possibly make the scroll contents change. I'll remove that lie, and leave it for this [task](https://searchbroker.atlassian.net/browse/EX-5016) that I've just created.

packages/x-components/src/x-modules/scroll/components/main-scroll.vue
```diff
@@ -0,0 +1,400 @@
...
+     */
+    @XProvide(FirstVisibleItemObserverKey)
+    public get firstVisibleElementObserver(): ScrollVisibilityObserver | null {
+      const observer = this.intersectionObserver;
+      return observer
+        ? {
+            observe: observer.observe.bind(observer),
+            unobserve: element => {
+              this.removeVisibleElement(element);
+              observer.observe(element);
```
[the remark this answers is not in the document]
javieri-empathy: Wow, sorry, good catch

packages/x-components/src/x-modules/scroll/components/main-scroll-item.vue
```diff
@@ -0,0 +1,163 @@
...
+     */
+    @Prop({ required: true })
+    public item!: Identifiable;
+
+    /**
+     * The tag to render.
+     *
+     * @public
+     */
+    @Prop({ default: 'NoElement' })
```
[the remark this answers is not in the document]
javieri-empathy: You are goddam right. I removed the import when the issue with the docs, thanks again.

packages/x-components/src/x-modules/scroll/components/main-scroll.vue
javieri-empathy: Yep, the only property that you should change is that `left` by a `right`. My problem is how to detect the RTL layout. I don't want to inject the `SnippetConfig` everywhere. Remember that this component might be used without it. And so far the only solution I know is recursively querying the DOM until you find a `dir` attribute.

packages/x-components/tests/unit/main-scroll.spec.ts
```diff
@@ -0,0 +1,264 @@
...
+});
+
+describe('testing MainScroll component', () => {
+  const cases: Array<Partial<RenderMainScrollOptions> & { description: string }> = [
+    { description: 'when using the html element', useWindow: true, windowScrollingElement: 'html' },
+    { description: 'when using the body element', useWindow: true, windowScrollingElement: 'body' },
+    { description: 'when using a custom scrolling element' }
+  ];
+
+  cases.forEach(({ description, ...defaultParameters }) => {
```
javieri-empathy, line 122: Before anyone says something, I hate the debugging experience of using this approach, you have to comment first the case, and then flag the test you want to debug with `.only`, but I hate more having to triplicate the tests.

packages/x-components/src/x-modules/scroll/components/main-scroll.vue
tajespasarela: while we don't have elements of 2 rows of width, there is no problem with this, I don't care if I take the first or the last of the row. I only want the vertical position for the scroll.

packages/x-components/tests/unit/main-scroll.spec.ts
tajespasarela: maybe this article could help: https://dev.to/tinesoft/be-smart-write-parameterized-tests-with-jest-3eab

packages/x-components/src/x-modules/scroll/components/main-scroll-item.vue
```diff
@@ -0,0 +1,163 @@
...
+
+To do so, it must be wrapped with the `MainScroll` component. In the following example we make use
+of all of these components. The URL is modified as the user scrolls.
+
+```vue
+<template>
+  <div>
+    <UrlHandler />
+    <SearchInput />
+    <BaseResultLink />
```
tajespasarela, line 130: The `BaseResultLink` is not in the correct place

packages/x-components/src/x-modules/scroll/components/scroll.const.ts
```diff
@@ -0,0 +1,10 @@
+import { XInjectKey } from '../../../components/decorators/injection.decorators';
+import { ScrollVisibilityObserver } from './scroll.types';
+
+/**
+ * Constant to inject the first visible item observer.
+ *
+ * @internal
+ */
+export const ScrollObserverKey: XInjectKey<ScrollVisibilityObserver | null> =
+  'firstVisibleItemObserverKey';
```
tajespasarela, line 10: update also the value

packages/x-components/tests/unit/main-scroll.css
```diff
@@ -0,0 +1 @@
+
```
tajespasarela, line 1: nice blank line! I was just saying this is an empty file, but not! it has that nice blank line 😍

tajespasarela (changes requested):

packages/x-components/tests/unit/main-scroll.spec.ts
javieri-empathy: I didn't know that Jest supported this by default, which is awesome if we need to do this (And I'm planing to do so in the scroll component). However I still have the same complains: I can't trigger a single test scenario without having to comment the rest.

tajespasarela: https://github.com/facebook/jest/tree/main/packages/jest-each#tagged-template-literal-of-rows

packages/x-components/tests/unit/main-scroll.css
javieri-empathy: If you remove this file the whole x-components break.

tajespasarela (approved):

guillei10 (approved):

PR #225, opened 2021-11-16
[EX-5009] Fix sliding panel test

mavmaf:
- `related-tags-stubs.factory.ts` have been modified in order to generate 12 RTs.
- doesn't seem necessary to modify `cy.wait(100)`
- Took advantage from the bug and added check for `sliding-panel-left-button`

## Motivation and context
Sliding panel e2e test is failing in Jenkins. Reproducible in local if CPU performance is slowdown x4.

## Type of change

- [x] `Main`
- [ ] Other. Specify:

## How has this been tested?

CPU performance have been slowdown x6 and test still passes.

## Checklist:

### DIFF
--- a/packages/x-components/src/__stubs__/related-tags-stubs.factory.ts
+++ b/packages/x-components/src/__stubs__/related-tags-stubs.factory.ts
@@ -9,7 +9,7 @@ import { RelatedTag } from '@empathyco/x-types';
  *
  * @internal
  */
-export function getRelatedTagsStub(amount = 9): RelatedTag[] {
+export function getRelatedTagsStub(amount = 12): RelatedTag[] {
   return Array.from({ length: amount }, (_, index) =>
     createRelatedTagStub('Related Tag', `Related Tag ${index + 1}`)
   );
--- a/packages/x-components/tests/e2e/cucumber/sliding-panel.feature
+++ b/packages/x-components/tests/e2e/cucumber/sliding-panel.feature
@@ -12,17 +12,14 @@ Feature: Sliding panel component
     Then  "right" sliding panel arrow is displayed
     And   only some related tags are visible
     When  "right" sliding panel arrow is clicked
-    And   wait for the movement of the elements inside the sliding panel
     Then  visible related tags have changed
     And   "both" sliding panel arrow is displayed
     And   only some related tags are visible
     When  "right" sliding panel arrow is clicked
-    And   wait for the movement of the elements inside the sliding panel
     Then  visible related tags have changed
     And   "left" sliding panel arrow is displayed
     And   only some related tags are visible
     When  "left" sliding panel arrow is clicked
-    And   wait for the movement of the elements inside the sliding panel
     Then  visible related tags have changed
 
     Examples:
--- a/packages/x-components/tests/e2e/cucumber/sliding-panel.feature
+++ b/packages/x-components/tests/e2e/cucumber/sliding-panel.feature
@@ -11,15 +11,19 @@ Feature: Sliding panel component
     When  "<query>" is searched
     Then  "right" sliding panel arrow is displayed
     And   only some related tags are visible
-    When  right sliding panel arrow is clicked
+    When  "right" sliding panel arrow is clicked
     And   wait for the movement of the elements inside the sliding panel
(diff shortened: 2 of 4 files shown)


packages/x-components/src/__stubs__/related-tags-stubs.factory.ts
```diff
@@ -9,7 +9,7 @@ import { RelatedTag } from '@empathyco/x-types';
  *
  * @internal
  */
-export function getRelatedTagsStub(amount = 9): RelatedTag[] {
+export function getRelatedTagsStub(amount = 12): RelatedTag[] {
```
javieri-empathy, line 12: Was the issue that there weren't enough related tags, and because of that the sliding panel wasn't able to perform a second scroll? That's what I've understood from the PR and from trying to break it myself, but I'm not sure I've got it right.

In any case I was also testing removing the `wait` just for fun, and slightly changing the `visible related tags have changed` allows us to do so. The change is just to use a `.should` inside the scroll `.then`, instead of another `then`, and  mappping the related tags there. This won't wait 100 ms, Cypress will keep re-executing the code until at least 1 related tag has changed or the 30s timeout happens.

```ts
Then('visible related tags have changed', () => {
  cy.getByDataTest('sliding-panel-scroll').then($scroll => {
    cy.getByDataTest('related-tag-item')
      .filter((_index, $element) => isElementVisible($element, $scroll.get(0)))
      .should($newRelatedTags => {
        const newRelatedTags = $newRelatedTags
          .map((_index, $relatedTag) => $relatedTag.textContent ?? '')
          .toArray();
        const areRelatedTagsDifferent = initialRelatedTags.some(relatedTag => !newRelatedTags.includes(relatedTag));
        expect(areRelatedTagsDifferent).to.be.true;
      });
  });
});
```

javieri-empathy (dismissed):

pmareke:
We have an error in the related tag spec file:

image

It should be 
`createRelatedTagStub('lego', 'bombero', { selected: true  }),`

packages/x-components/tests/e2e/cucumber/sliding-panel.feature
```diff
@@ -11,15 +11,19 @@ Feature: Sliding panel component
     When  "<query>" is searched
     Then  "right" sliding panel arrow is displayed
     And   only some related tags are visible
-    When  right sliding panel arrow is clicked
+    When  "right" sliding panel arrow is clicked
     And   wait for the movement of the elements inside the sliding panel
```
pmareke, line 15: If we use the Javi's approach we can remove the wait step and simplify a lot the scenario.

packages/x-components/tests/e2e/cucumber/sliding-panel.spec.ts
```diff
@@ -44,8 +44,8 @@ Then('{string} sliding panel arrow is displayed', (displayedArrows: string) => {
   }
 });
 
-When('right sliding panel arrow is clicked', () => {
-  cy.getByDataTest('sliding-panel-right-button').click();
+When('{string} sliding panel arrow is clicked', (direction: string) => {
+  cy.getByDataTest(`sliding-panel-${direction}-button`).click();
 });
 
 Then('wait for the movement of the elements inside the sliding panel', () => {
```
pmareke, line 51: Remove this wait, update the scenario and the `visible related tags have changed` step please.

packages/x-components/src/__stubs__/related-tags-stubs.factory.ts
pmareke: I'm not sure either as with 9 RTs still works.

pmareke (changes requested):

packages/x-components/src/__stubs__/related-tags-stubs.factory.ts
mavmaf: For what I have seen I don't even know why it is/was working with 9 RTs. When manually tested, with only one click in the `sliding-panel-right-button` the end of the RTs is reached, so why is a second click possible?

javieri-empathy (approved):

pmareke (approved):

PR #227, opened 2021-11-16
 feat(tagging): add `tagging` component

tiborux:
EX-3523


packages/x-components/src/x-modules/tagging/components/__tests__/tagging.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+        consent,
+        snippetConfig
+      },
+      localVue
+    }
+  );
+
+  return {
+    wrapper: wrapper.findComponent(Tagging),
+    on: wrapper.vm.$x.on.bind(wrapper.vm.$x)
```
pmareke, line 47: I'll remove this `on` and use the wrapper in the rest of the tests:

```javascript
const { wrapper } = renderTagging({ consent: true });
const eventSpy = jest.fn();

await wrapper.vm.$x.on('ConsentProvided').subscribe(eventSpy);
expect(eventSpy).toHaveBeenNthCalledWith(1, true);
```

pmareke (changes requested):
🔝

packages/x-components/src/x-modules/tagging/components/tagging.vue
```diff
@@ -0,0 +1,122 @@
...
+    components: {
+      Tagging
+    }
+  };
+</script>
+```
+
+### Play with props
+
+In this example, the `Tagging` component will emit `ConsentProvided` only if the consent is set to true.
```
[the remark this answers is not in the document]
tiborux: yes

packages/x-components/src/x-modules/tagging/components/tagging.vue
```diff
@@ -0,0 +1,122 @@
...
+  mixins: [xComponentMixin(taggingXModule)]
+})
+export default class Tagging extends Vue {
+  /**
+   * It injects {@link SnippetConfig} provided by an ancestor as snippetConfig.
+   *
+   * @internal
+   */
+  @Inject('snippetConfig')
+  protected snippetConfig!: SnippetConfig;
```
javieri-empathy, line 18: Make this `snippetConfig` optional. We might not always have this snippet config, and the tagging module should still work without it.

```suggestion
  @Inject('snippetConfig')
  protected snippetConfig!: SnippetConfig;
```

packages/x-components/src/x-modules/tagging/components/__tests__/tagging.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+@Component({
+  template: `<div><slot/></div>`
+})
+class Provider extends Vue {
+  @Prop({ default: baseSnippetConfig })
+  @Provide()
+  public snippetConfig!: SnippetConfig;
+}
+
+function renderTagging({
```
javieri-empathy, line 20: You don't need the `Provider` component, you can do that provide in the wrapper that you are creating. Also, I would suggest to also make the `snippetConfig` reactive here, and clone it just to avoid mutating the `baseSnippetConfig` object.

```ts
function renderTagging({
  template = `<Tagging :consent="consent"/>`,
  consent = undefined,
  snippetConfig = { ...baseSnippetConfig }
}: RenderTaggingOptions = {}): RenderTaggingAPI {
  const [, localVue] = installNewXPlugin();
  snippetConfig = localVue.observable(snippetConfig);

  const wrapper = mount(
    {
      components: {
        Tagging
      },
      props: ['consent', 'snippetConfig'],
      template
    },
    {
      provide: {
        snippetConfig
      },
      propsData: {
        consent,
        snippetConfig
      },
      localVue
    }
  );

  return {
    wrapper: wrapper.findComponent(Tagging),
    on: wrapper.vm.$x.on.bind(wrapper.vm.$x)
  };
}
```

packages/x-components/src/x-modules/tagging/components/__tests__/tagging.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+  it('emits ConsentProvided when the consent is set and send it by prop', () => {
+    const { on } = renderTagging({ consent: true });
+    const eventSpy = jest.fn();
+
+    on('ConsentProvided').subscribe(eventSpy);
+    expect(eventSpy).toHaveBeenNthCalledWith(1, true);
+  });
+
+  it("doesn't emit ConsentProvided when the consent is undefined send it by prop", () => {
+    const { on } = renderTagging();
```
javieri-empathy, line 67: As I proposed below with the changes in the component, I would say that the component should emit a `ConsentProvided: false` by default, if neither the snippet config consent, or the prop consent is passed.

In both cases I would make that situation explicit here in the `renderTagging` function, passing both a snippet config and a prop with consent = `undefined`.

packages/x-components/src/x-modules/tagging/components/__tests__/tagging.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+    on('ConsentProvided').subscribe(eventSpy);
+    expect(eventSpy).not.toHaveBeenCalled();
+  });
+
+  // eslint-disable-next-line max-len
+  it('emits ConsentProvided when the consent is provided and send it using the snippet config', async () => {
+    const snippet = Vue.observable({
+      ...baseSnippetConfig,
+      consent: false
+    } as SnippetConfig & { consent?: boolean });
```
javieri-empathy, line 79: Why do you need that ugly cast?

```suggestion
const snippet = Vue.observable<SnippetConfig>({
      ...baseSnippetConfig,
      consent: false
    });
```

packages/x-components/src/x-modules/tagging/components/__tests__/tagging.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+    expect(isXComponent(wrapper.vm)).toEqual(true);
+    expect(getXComponentXModuleName(wrapper.vm)).toEqual('tagging');
+  });
+
+  it('emits ConsentProvided when the consent is set and send it by prop', () => {
+    const { on } = renderTagging({ consent: true });
+    const eventSpy = jest.fn();
+
+    on('ConsentProvided').subscribe(eventSpy);
+    expect(eventSpy).toHaveBeenNthCalledWith(1, true);
```
javieri-empathy, line 63: When asserting calls, try to be exhaustive, and assert also the number of times it should have been called.

```suggestion
    expect(eventSpy).toHaveBeenCalledTimes(1);
    expect(eventSpy).toHaveBeenNthCalledWith(1, true);
```

packages/x-components/src/x-modules/tagging/components/__tests__/tagging.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+}
+
+describe('testing Tagging component', () => {
+  it('is an XComponent which has an XModule', () => {
+    const { wrapper } = renderTagging();
+    expect(isXComponent(wrapper.vm)).toEqual(true);
+    expect(getXComponentXModuleName(wrapper.vm)).toEqual('tagging');
+  });
+
+  it('emits ConsentProvided when the consent is set and send it by prop', () => {
```
javieri-empathy, line 58:
```suggestion
  it('emits ConsentProvided when the consent is set using the prop', () => {
```

packages/x-components/src/x-modules/tagging/components/__tests__/tagging.spec.ts
```diff
@@ -0,0 +1,116 @@
...
+  it("doesn't emit ConsentProvided when the consent is undefined send it by prop", () => {
+    const { on } = renderTagging();
+    const eventSpy = jest.fn();
+
+    on('ConsentProvided').subscribe(eventSpy);
+    expect(eventSpy).not.toHaveBeenCalled();
+  });
+
+  // eslint-disable-next-line max-len
+  it('emits ConsentProvided when the consent is provided and send it using the snippet config', async () => {
```
javieri-empathy, line 75:
```suggestion
  it('emits ConsentProvided when the consent is set using the snippet config', async () => {
```

packages/x-components/src/x-modules/tagging/store/emitters.ts
```diff
@@ -6,4 +6,9 @@ import { taggingXStoreModule } from './module';
  *
  * @internal
  */
-export const taggingEmitters = createStoreEmitters(taggingXStoreModule, {});
+export const taggingEmitters = createStoreEmitters(taggingXStoreModule, {
+  ConsentChanged: {
+    selector: state => state.consent!,
+    filter: consent => !consent
```
javieri-empathy, line 12: Hmm, I think this event is lying, it says that the consent has changed, and by that I understand to any possible value, but in reality, it will only be emitted when the consent is false. Maybe we can do that `filter` in the wire instead?

javieri-empathy (changes requested):

packages/x-components/src/x-modules/tagging/components/__tests__/tagging.spec.ts
tiborux: ill prefer to return the on in the function, like in the url tests

packages/x-components/src/views/Layout.vue
```diff
@@ -1,5 +1,6 @@
 <template>
   <div>
+    <Tagging :consent="false" />
```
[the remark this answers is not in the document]
tiborux: The session id is created when doing a tracking request, at this moment there are no tracking request so you are not going to see it.

javieri-empathy (dismissed):

javieri-empathy:
The buid is failing due to api extractor unable to find the `TaggingConfig` links. If instead of the state you directly use the `TaggingConfig` interface for links it works. For example, `TaggingConfig.sessionTTLMs` instead of `TaggingState.config.sessionTTLMs`

packages/x-components/src/views/Layout.vue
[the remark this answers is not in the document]
pmareke: I know, I know and because of that we could even remove this component from the PR. What I meant was to leave it as true in order to see later in the requests, but whatever you want.

packages/x-components/src/x-modules/tagging/store/module.ts
```diff
@@ -7,9 +7,23 @@ import { TaggingXStoreModule } from './types';
...
+    config: {
+      sessionTTLMs: 30 * 60 * 1000,
+      queryTaggingDebounceMs: 200
+    },
+    consent: null
   }),
   getters: {},
-  mutations: {},
+  mutations: {
+    setSessionDuration(state, sessionTTLMs) {
```
pmareke, line 18: We can destruct the object here, what do you think?

packages/x-components/src/x-modules/tagging/components/tagging.vue
```diff
@@ -0,0 +1,128 @@
...
+     * @public
+     */
+    @Prop()
+    protected consent?: boolean;
+
+    /**
+     * The active consent, selected from the `consent` prop and the `snippetConfig.consent`
+     * property. False by default.
+     *
+     * @returns A boolean that represents if the consent is accepted or not.
```
pmareke, line 51: I would like to have a `@remaks` explaining that if every consent is undefined we'l return false.

packages/x-components/src/x-modules/tagging/components/tagging.vue
```diff
@@ -0,0 +1,128 @@
...
+    components: {
+      Tagging
+    }
+  };
+</script>
+```
+
+### Play with props
+
+In this example, the `Tagging` component will emit `ConsentProvided` only if the consent is set to
```
pmareke, line 97: Review this text please, is not correct.

packages/x-components/src/x-modules/tagging/components/__tests__/tagging.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+import Vue from 'vue';
+import { installNewXPlugin } from '../../../../__tests__/utils';
+import { getXComponentXModuleName, isXComponent } from '../../../../components/x-component.utils';
+import { XComponentBusAPI } from '../../../../plugins/x-plugin.types';
+import { baseSnippetConfig } from '../../../../views/base-config';
+import { SnippetConfig } from '../../../../x-installer/api/api.types';
+import Tagging from '../../../tagging/components/tagging.vue';
+
+function renderTagging({
+  template = `<Tagging
```
pmareke, line 11: :trollface: 

```javascript
  template = `<Tagging
                :consent="consent"
                :sessionTTLMs="sessionTTLMs"
                :queryTaggingDebounceMs="queryTaggingDebounceMs"
              />`,
```

packages/x-components/src/x-modules/tagging/components/__tests__/tagging.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+}: RenderTaggingOptions = {}): RenderTaggingAPI {
+  const [, localVue] = installNewXPlugin();
+  snippetConfig = localVue.observable(snippetConfig);
+
+  const wrapper = mount(
+    {
+      components: {
+        Tagging
+      },
+      props: ['consent', 'snippetConfig', 'sessionTTLMs', 'queryTaggingDebounceMs'],
```
pmareke, line 28: To be consistent with the `propsData`

```
props: ['consent', 'queryTaggingDebounceMs', 'sessionTTLMs', 'snippetConfig'],
```

packages/x-components/src/x-modules/tagging/components/__tests__/tagging.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+    });
+    const eventSpy = jest.fn();
+
+    on('ConsentProvided').subscribe(eventSpy);
+    expect(eventSpy).toHaveBeenCalledTimes(1);
+    expect(eventSpy).toHaveBeenNthCalledWith(1, false);
+  });
+
+  it('emits ConsentProvided when the consent is set using the snippet config', async () => {
+    const snippet = Vue.observable<SnippetConfig>({
```
pmareke, line 105: You can remove this `Vue.observable` as in the render function you'll use the observable again.

packages/x-components/src/x-modules/tagging/components/__tests__/tagging.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+
+    snippet.consent = true;
+    await wrapper.vm.$nextTick();
+    expect(eventSpy).toHaveBeenCalledTimes(2);
+    expect(eventSpy).toHaveBeenNthCalledWith(2, true);
+  });
+});
+
+interface RenderTaggingOptions {
+  /** The consent value. */
```
pmareke, line 127: Would you mind to order the props in both interfaces?

pmareke (changes requested):
As @javieri-empathy mention, please fix the build

packages/x-components/src/views/Layout.vue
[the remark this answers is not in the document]
tiborux: I will leave it as it is and we can change it in the future if we need it

pmareke (approved):

javieri-empathy (approved):

PR #232, opened 2021-11-18
feat(search): add `SearchTaggingChanged`  emitter and save coming `queryTagging` from the search response

tiborux:
Make a request and check the store state in the search module, you will see there the tagging object stored.

EX-3525

### DIFF
--- a/packages/x-components/src/__stubs__/search-response-stubs.factory.ts
+++ b/packages/x-components/src/__stubs__/search-response-stubs.factory.ts
@@ -19,8 +19,13 @@ export function getSearchResponseStub(): SearchResponse {
     partialResults: [],
     promoteds: getPromotedsStub(),
     queryTagging: {
-      params: {},
-      url: ''
+      params: {
+        q: 'lego',
+        totalHits: '789',
+        lang: 'es',
+        follow: false
+      },
+      url: 'https://api.empathybroker.com/tagging/v1/track/query'
     },
     redirections: getRedirectionsStub(),
     results: getResultsStub(),
--- a/packages/x-components/src/x-modules/search/store/__tests__/actions.spec.ts
+++ b/packages/x-components/src/x-modules/search/store/__tests__/actions.spec.ts
@@ -70,7 +70,7 @@ describe('testing search module actions', () => {
 
   describe('fetchAndSaveSearchResponse', () => {
     // eslint-disable-next-line max-len
-    it('should request and store results, facets, banners, promoteds and redirections in the state', async () => {
+    it('should request and store results, facets, banners, promoteds, redirections and query tagging in the state', async () => {
       resetSearchStateWith(store, {
         query: 'lego'
       });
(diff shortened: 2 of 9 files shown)


packages/x-components/src/x-modules/search/store/module.ts
```diff
@@ -103,6 +107,9 @@ export const searchXStoreModule: SearchXStoreModule = {
     },
     setRedirections(state, redirections) {
       state.redirections = redirections;
+    },
+    setQueryTagging(state, queryTagging) {
+      state.queryTagging = queryTagging;
```
pmareke, line 112: As this is an an object I would rather use `Vue.set()` as the docs says

https://vuex.vuejs.org/guide/mutations.html#mutations-follow-vue-s-reactivity-rules

Am I missing something here? Because is working perfectly.

packages/x-components/src/x-modules/search/store/actions/fetch-and-save-search-response.action.ts
```diff
@@ -29,6 +39,7 @@ const { fetchAndSave, cancelPrevious } = createFetchAndSaveActions<
       commit('setFacets', facets);
     }
 
+    commit('setQueryTagging', queryTagging);
```
pmareke, line 42: I would like to see this new info in the `fetchAndSaveSearchResponse` action test, please update it.

packages/x-components/src/__stubs__/search-response-stubs.factory.ts
```diff
@@ -19,8 +19,13 @@ export function getSearchResponseStub(): SearchResponse {
     partialResults: [],
     promoteds: getPromotedsStub(),
     queryTagging: {
-      params: {},
-      url: ''
+      params: {
```
pmareke, line 22: Why do we need this info right now? As far as I understand we're not testing it right?

pmareke (changes requested):
Nice job, we almost have the tagging!

packages/x-components/src/x-modules/search/store/__tests__/actions.spec.ts
```diff
@@ -70,7 +70,7 @@ describe('testing search module actions', () => {
```
tajespasarela, line 70: I think the `mockedEmptySearchResponse` is not really a mock. It is not replacing a real "thing" and acting like it (As the MockedAdapter).

To me is another stub. I think we should create a `getEmptySearchResponseStub` along the other stub factories, and use here like any other stub

tajespasarela (changes requested):

packages/x-components/src/x-modules/search/store/module.ts
pmareke: We're adding new properties to an object, as is empty by default. So for me is exactly what the docs says.

pmareke (approved):

tajespasarela (approved):

PR #237, opened 2021-11-19
feat(scroll): ignore first visible element when it is also the first one observed

javieri-empathy:
EX-5041

This checks the element is not the first node matching the `[data-scroll]` selector, so when the scroll is at the top we don't have the first visible element stored in the URL.

It also uses a constant for the default main scroll, so the scroll to top button works by default with both components, and the scroll id is only needed to be passed if there are more scroll component instances.

### DIFF
--- a/packages/x-components/src/x-modules/scroll/components/window-scroll.vue
+++ b/packages/x-components/src/x-modules/scroll/components/window-scroll.vue
@@ -4,6 +4,7 @@
   import { ScrollDirection, ScrollMixin, xComponentMixin } from '../../../components';
   import { WireMetadata } from '../../../wiring';
   import { scrollXModule } from '../x-module';
+  import { MainScrollId } from './scroll.const';
 
   type ScrollableTag = 'html' | 'body';
 
@@ -30,7 +31,7 @@
      *
      * @public
      */
-    @Prop({ default: 'main-scroll' })
+    @Prop({ default: MainScrollId })
     protected id!: string;
 
     mounted(): void {
@@ -64,7 +65,7 @@
     protected initAndListenElement(): void {
       // eslint-disable-next-line @typescript-eslint/ban-ts-comment
       // @ts-expect-error
-      this.$el = this.tag === 'html' ? document.documentElement : document.body;
+      this.$el = this.tag === 'body' ? document.body : document.documentElement;
       this.$el.addEventListener('scroll', this.throttledStoreScrollData);
     }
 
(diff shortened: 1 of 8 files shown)


packages/x-components/tests/unit/main-scroll.spec.ts
```diff
@@ -141,18 +139,17 @@ describe('testing MainScroll component', () => {
         userScrolledToElementSpy().should('have.been.calledWith', 'item-5');
       });
 
-      it('ignores the first pixels of scroll', () => {
+      it.only('ignores the first element of scroll', () => {
```
pmareke, line 142: Please remove the `only` keyword, nothing is going to happen, believe in yourself!

packages/x-components/tests/unit/main-scroll.spec.ts
```diff
@@ -141,18 +139,17 @@ describe('testing MainScroll component', () => {
...
         const { scrollToItem, userScrolledToElementSpy } = renderMainScroll({
           ...defaultParameters,
-          minScrollTrigger: 75,
           itemHeight: '50px'
```
pmareke, line 145: You're using the same value as the default, so please remove it or change the value to express why are you sending it to the render function.

pmareke (changes requested):
When you reload the page the scroll position cuts the images of the results in the first row:

image

tajespasarela:
> When you reload the page the scroll position cuts the images of the results in the first row:
> 
> image

This is a known issue. We will deal with it in https://searchbroker.atlassian.net/browse/EX-5024

tajespasarela (changes requested):

packages/x-components/tests/unit/main-scroll.spec.ts
javieri-empathy: I'm going to remove it because this is no longer relevant for the test. The test does no longer depend on element heights. 
With the previous approach it was dependant, so I would rather be explicit about my test inputs and repeat them than let them be implicit and confusing.

packages/x-components/src/x-modules/scroll/components/window-scroll.vue
```diff
@@ -4,6 +4,7 @@
   import { ScrollDirection, ScrollMixin, xComponentMixin } from '../../../components';
   import { WireMetadata } from '../../../wiring';
   import { scrollXModule } from '../x-module';
+  import { MainScrollId } from './scroll.const';
 
   type ScrollableTag = 'html' | 'body';
```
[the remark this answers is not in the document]
javieri-empathy: Dislike but done

pmareke (approved):

tajespasarela (approved):

PR #238, opened 2021-11-19
ci: add new auto fix action

pmareke:
Right now if you edit a file in the website there is not lint applied.

This new action will lint every PR no matter the origin.


herrardo (commented):
Shouldn't this be a step of `build.yml - Check branch is releasable` action?

In that action that is launched for every PR `npm run lint` is in fact already run, it's prettier what is missing.

pmareke:
> Shouldn't this be a step of `build.yml - Check branch is releasable` action?
> 
> In that action that is launched for every PR `npm run lint` is in fact already run, it's prettier what is missing.

I didn't think about this approach and has more sense, I'll update the PR with this change thanks.

PR #239, opened 2021-11-22
[EX-4923] reorder mocked responses

mavmaf:
## Motivation and context
Mocked responses are now ordered in `mocked-responses.spec.ts` for convenience .

## Type of change

- [x] `Main`
- [ ] Other. Specify:

## How has this been tested?

As usual, if needed please change `cypress.json` in order to run tests independently.

## Checklist:


packages/x-components/tests/e2e/cucumber/mocked-responses.spec.ts
```diff
@@ -0,0 +1,558 @@
+import {
```
pmareke, line 1: I'm seeing that we have one Results intercept after the spellcheck:

image

pmareke: Also I would try to improve the steps names, as the `...with a known response` is quite confuse, what do you think?

packages/x-components/tests/e2e/cucumber/mocked-responses.spec.ts
```diff
@@ -0,0 +1,558 @@
...
+        url: 'https://tagging.empathy.co/',
+        params: {}
+      },
+      redirections: [],
+      results: [
+        createResultStub('LEGO Super Mario Pack Inicial: Aventuras con Mario - 71360', {
+          images: ['https://picsum.photos/seed/1/100/100']
+        })
+      ],
+      facets: getFacetsStub(),
```
pmareke, line 104: You can remove this faces as they're not used in the tests.

`facets: [],`

packages/x-components/tests/e2e/cucumber/mocked-responses.spec.ts
```diff
@@ -0,0 +1,558 @@
...
+      queryTagging: {
+        url: 'https://tagging.empathy.co',
+        params: {}
+      },
+      spellcheck: ''
+    });
+  }).as('interceptedNewResults');
+});
+
+Given('a results API with {int} results', (resultsLength: number) => {
```
pmareke, line 375: This is only used in the scroll feature, because of that you can remove the facets from the response and simplify it.

pmareke (changes requested):
Nice job!

packages/x-components/tests/e2e/cucumber/mocked-responses.spec.ts
```diff
@@ -0,0 +1,558 @@
...
+  getNextQueriesStub,
+  getPopularSearchesStub,
+  getQuerySuggestionsStub,
+  getRelatedTagsStub,
+  getResultsStub
+} from '../../../src/__stubs__';
+
+// ID Results
+Given('an ID results API', () => {
+  cy.intercept('https://api.empathy.co/searchById', req => {
```
CachedaCodes, line 29: Now that the `intercept`are all in the same place we can extract the mocked API URL to a constant and make all the `intercept` use that instead.
Also, extract the API endpoints that are repeated. I'm thinking of something like:

```javascript
const mockedApiUrl = 'https://api.empathy.co'

// ID Results
const searchByIdEndpoint = `${mockedApiUrl}/searchById`;

Given('an ID results API', () => {
  cy.intercept(searchByIdEndpoint, req => {
[...]

// Partial Results
Given('a results API with partial results', () => {
  cy.intercept(`${mockedApiUrl}/search`, req => {
[...]
```

packages/x-components/tests/e2e/cucumber/mocked-responses.spec.ts
CachedaCodes: The `...with a known response` is there to tell that the response is specified in the test (because we're mocking it). I recall we wrote it this way because we had to avoid using `...with a mocked response`. Do you have any alternative to say the same? I can't think of any 😞 .

CachedaCodes (changes requested):

pmareke (approved):

packages/x-components/tests/e2e/cucumber/mocked-responses.spec.ts
pmareke: The thing is that in the `a results API` you're also mocking the response 😆.

I'm the worst with the names so we can think about this one in the future.

CachedaCodes (approved):

PR #240, opened 2021-11-22
[EX-5015] E2E tests retries

mavmaf:
## Motivation and context
Right now, the timeout until one test is considered fail is 30s. Moreover, the number of retries in case of failure is 3. This means a total of 30s x 3 = 1m30s waiting in case a test is failing.
The aim of this task is to avoid wasting time by finding the best possible combination between these two.

**### SOLUTION
7s, 1 retry. Build ran successfully 10/10 times.**

## Type of change

- [ ] `Main`
- [ ] Other. Specify:

## How has this been tested?
Multiple builds will be launched to determine which is the limit the tests start failing

## Checklist:

### DIFF
--- a/packages/x-components/cypress.json
+++ b/packages/x-components/cypress.json
@@ -1,6 +1,6 @@
 {
   "baseUrl": "http://localhost:8080",
-  "defaultCommandTimeout": 30000,
+  "defaultCommandTimeout": 7000,
   "viewportHeight": 1080,
   "viewportWidth": 1920,
   "screenshotOnRunFailure": false,
@@ -15,5 +15,5 @@
     "componentFolder": "tests/unit",
     "testFiles": "**/*.spec.ts"
   },
-  "retries": 2
+  "retries": 1
 }


packages/x-components/cypress.json
```diff
@@ -1,6 +1,6 @@
 {
   "baseUrl": "http://localhost:8080",
-  "defaultCommandTimeout": 30000,
+  "defaultCommandTimeout": 7000,
```
pmareke, line 3: This is a silly question but...why is 7 seconds and not 8 or 6? Is guess is the minimum you found to pass the tests?

pmareke: Can we remove it and try with the default (4s)?

pmareke (approved):

javieri-empathy (approved):

packages/x-components/cypress.json
javieri-empathy: With the default values, and different configurations (ci, throttled 4x, throttled 6x) I've  only seen it fail on the search-box test. I would merge this as it is. It is a big improvement. In case it fails  in the e2e tests we only have to wait 14 seconds, not 1 minute  and a half like before.

PR #242, opened 2021-11-24
[EX-4902] Set keyboard navigation in header and Empathize

mavmaf:
## Motivation and context
Although the first idea was to include keyboard-navigation in all the layout including the grid, the approach changed to make it only work in the Empathize. For consistency, it has been added as well in part of the header (`SearchInput`, `ClearSearchInput `and `SearchButton`). It was not possible to added it to the RTs, though. However, that's not essential (although it is desirable) since it's not mandatory for RTs to be placed just below the `search-box`

## Type of change

- [x] `Main`
- [ ] Other. Specify:

## How has this been tested?

As usual, change `cypress.json` file in order to run tests independently. 

## Checklist:

### DIFF
--- a/packages/x-components/tests/e2e/cucumber/keyboard-navigation.spec.ts
+++ b/packages/x-components/tests/e2e/cucumber/keyboard-navigation.spec.ts
@@ -11,7 +11,9 @@ type Direction = 'above' | 'below' | 'on the left' | 'on the right';
 
 // Scenario 1
 And('{string} element position is stored', (focusableElement: string) => {
-  cy.getByDataTest(focusableElement).focus().as('originalElement');
+  return focusableElement === 'search-input'
+    ? cy.getByDataTest(focusableElement).focus().as('originalElement')
+    : cy.getByDataTest(focusableElement).last().focus().as('originalElement');
 });
 
 Then('next element position is "{direction}"', (expectedPosition: Direction) => {
--- a/packages/x-components/tests/e2e/cucumber/keyboard-navigation.spec.ts
+++ b/packages/x-components/tests/e2e/cucumber/keyboard-navigation.spec.ts
@@ -8,12 +8,11 @@ defineParameterType({
   }
 });
 type Direction = 'above' | 'below' | 'on the left' | 'on the right';
+type Move = 'up' | 'right' | 'bottom' | 'left';
 
 // Scenario 1
 And('{string} element position is stored', (focusableElement: string) => {
-  return focusableElement === 'search-input'
-    ? cy.getByDataTest(focusableElement).focus().as('originalElement')
-    : cy.getByDataTest(focusableElement).last().focus().as('originalElement');
+  return cy.getByDataTest(focusableElement).last().focus().as('originalElement');
 });
 
 Then('next element position is "{direction}"', (expectedPosition: Direction) => {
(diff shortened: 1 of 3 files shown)


pmareke (changes requested):
As the query is not relevant here I think you can move to the Background steps the first four steps of each scenario

```cucumber
Given no special config for layout view
And   start button is clicked
When  "lego" is searched
And   related results are displayed
```

And remove the query from the examples.

guillei10:
I agree with @pmareke. Searching is not relevant for testing the `BaseKeyboardNavigation`.

For the rest, looks awesome.

packages/x-components/tests/e2e/cucumber/keyboard-navigation.spec.ts
```diff
@@ -11,7 +11,9 @@ type Direction = 'above' | 'below' | 'on the left' | 'on the right';
 
 // Scenario 1
 And('{string} element position is stored', (focusableElement: string) => {
```
pmareke, line 13: I've just discovered that we've an error in the following step

```cucumber
When(
  '{string} arrow is pressed {int} times',
  (direction: 'up' | 'right' | 'bottom' | 'right', pressedTimes: number) => {
    for (let i = 0; i < pressedTimes; i++) {
      cy.focused().type(`{${direction}arrow}`);
    }
  }
);
```

As you can see `right` is duplicated and left is not present, I would like to extract this 4 options to a type and also simplify the loop:

`type Move = 'up' | 'right' | 'bottom' | 'left';`

```cucumber
When('{string} arrow is pressed {int} times', (direction: Move, pressedTimes: number) => {
  Array.from({ length: pressedTimes }).forEach(() => {
    cy.focused().type(`{${direction}arrow}`);
  });
});

File: packages/x-components/tests/e2e/cucumber/keyboard-navigation.spec.ts [hunk 2]
```diff
@@ -11,7 +11,9 @@ type Direction = 'above' | 'below' | 'on the left' | 'on the right';
...
-  cy.getByDataTest(focusableElement).focus().as('originalElement');
+  return focusableElement === 'search-input'
```
pmareke, line 14: When the `focusableElement` is the `search-box` and there is only one, the `last()` is the `first` one too :trollface: , so you can remove the if statement:

```cucumber
// Scenario 1
And('{string} element position is stored', (focusableElement: string) => {
  return cy.getByDataTest(focusableElement).last().focus().as('originalElement');
});
```

packages/x-components/tests/e2e/cucumber/keyboard-navigation.spec.ts
```diff
@@ -11,7 +11,9 @@ type Direction = 'above' | 'below' | 'on the left' | 'on the right';
...
+    ? cy.getByDataTest(focusableElement).focus().as('originalElement')
+    : cy.getByDataTest(focusableElement).last().focus().as('originalElement');
 });
 
 Then('next element position is "{direction}"', (expectedPosition: Direction) => {
```
pmareke, line 19: Something is wrong with the `XXX out of bounds is reached` as if you remove the 

```javascript
.trigger('keydown', { key: 'ArrowDown', force: true })
.trigger('keydown', { key: 'ArrowRight', force: true })
```

They tests pass exactly the same, please take a look.

pmareke (changes requested):

guillei10 (changes requested):
Haven't manage to find any other issues. Review the ones @pmareke pointed out and we are good from my part.

packages/x-components/tests/e2e/cucumber/keyboard-navigation.spec.ts
mavmaf: Hahaha that's what I had at first and I changed it because I thought you were going to complain

packages/x-components/tests/e2e/cucumber/keyboard-navigation.spec.ts
mavmaf: Mmmm don't know if I understood correctly, but that's the point. What I am trying to check here is that once you have the focus on the first / last possible navigable position, pressing the pertinent arrow keys will be the same as not pressing anything at all since the bound have been reached. Thus, if you are on the first position (`originalElement`) and you use left and/or up keys to move to a _previous_ one (`targetElement`), it won't be possible and `originalElement` and `targetElement` will have the same value, so the equivalent is as if you pressed nothing at all.

pmareke (approved):

guillei10 (approved):

PR #243, opened 2021-11-24
Feature/ex 3535 create globalxbus component

guillei10:
# Description

The goal of this PR is to create the `GlobalXBus` component. The purpose of this component is to receive events each with an associated function that will execute once the event is emitted from the bus. This way we allow the user to configure its own callbacks. For this, we allow the user to pass this config using the `$listeners` of the component

Once the component is destroyed, it unsubscribes to all the listeners to avoid leaks.

## Motivation and context

This is a part of the Snippet behaviour. Once this component is created, other components (like the SnippetConfig for instance), will be able to pass the listeners to this component.

## Test it

You can actually implement this in the `Layout` view and play with it. Example: 

_template:_
` <GlobalXBus @UserAcceptedAQuery="printQuery" @UserClickedColumnPicker="printColumns" />`

_script:_
```
     protected printQuery(query: string, metadata: any): string {
         console.log('query: ', query);
         console.log('metadata: ', metadata);
         return query;
    }
```
```
    protected printColumns(columns: string, _: any): void {
         console.log('columns: ', columns);
    }
```
```
    @XOn('SnippetCallbackExecuted')
    printThings(payload: any): void {
         console.log('event emitted with payload:', payload);
    }
```

Have fun!


packages/x-components/src/components/global-x-bus.vue
```diff
@@ -0,0 +1,122 @@
...
+      });
+    }
+
+    /**
+     * Unsubscribe to all subscriptions in {@link GlobalXBus.eventSubscriptions}.
+     *
+     * @internal
+     */
+    protected unsubscribeEvents(): void {
+      forEach(this.eventSubscriptions, (_, subscription) => {
```
pmareke, line 67: I think we can do this forEach in the beforeDestroy method instead of adding a new one.

packages/x-components/src/components/__tests__/global-x-bus.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Options to configure how the progress bar component should be rendered.
+ */
+interface RenderGlobalXBusOptions {
+  /** The listeners object in the component.*/
+  listeners?: ComponentOptions<Vue>['methods'];
+}
+
+/**
+ * Options to configure how the progress bar component should be rendered.
```
pmareke, line 77: Review comment.

packages/x-components/src/components/global-x-bus.vue
```diff
@@ -0,0 +1,122 @@
...
+  import { WireMetadata, XEvent, XEventPayload } from '../wiring';
+  import { NoElement } from './no-element';
+
+  type EventListeners = Record<
+    XEvent,
+    (payload: XEventPayload<XEvent>, metadata?: WireMetadata) => unknown
+  >;
+
+  /**
+   * A component which purpose is to receive callbacks associated with events and handle the
```
pmareke, line 14:
```
  /**
   * A component which receives a list of listeners, a key-value pair of an {@link XEvent} and a callback,
   * and iterate over them in order to subscribe each of them to the bus.
   *
   * When an XEvent is registered in the {@link XBus} this component keeps track of the callback returns
   *  in order to send it back letter using the  `SnippetCallbackExecuted` event.
   *
   * @remarks
   *
   * @public
   */
```

Something like this?

pmareke (changes requested):

pmareke:
It's a really nice PR man, I'm so proud of you ❤️ .

packages/x-components/src/components/global-x-bus.vue
```diff
@@ -0,0 +1,122 @@
...
+  import { Component } from 'vue-property-decorator';
+  import { Subscription } from 'rxjs';
+  import { forEach } from '../utils';
+  import { WireMetadata, XEvent, XEventPayload } from '../wiring';
+  import { NoElement } from './no-element';
+
+  type EventListeners = Record<
+    XEvent,
+    (payload: XEventPayload<XEvent>, metadata?: WireMetadata) => unknown
+  >;
```
javieri-empathy, line 11: The type is wrong. You can easily see this by invoking any callback with any value. TypeScript should complain.  Also the metadata is not optional.

I suggest to rename it to `XEventListeners`, and make the following change.

```ts
/**
* Map type of every {@link XEvent} and a callback with the payload and metadata for that event.
*/  
type XEventListeners = {
    [Event in XEvent]: (payload: XEventPayload<Event>, metadata: WireMetadata) => void;
  };
```

This change will force you below to cast the payload: `callback(eventPayload as never, metadata);`, we can leave a TODO there remembering us to improve the types. I've been trying to fix this, but I don't know why it infers never. The type is fine this way though, so the issue might be with the `object.ts` utilities. I also tried to improve their types even if it is out of the scope of this PR and failed.

packages/x-components/src/components/global-x-bus.vue
```diff
@@ -0,0 +1,122 @@
...
+     *
+     * @internal
+     */
+    protected unsubscribeEvents(): void {
+      forEach(this.eventSubscriptions, (_, subscription) => {
+        subscription.unsubscribe();
+      });
+    }
+
+    mounted(): void {
```
javieri-empathy, line 72: Why do you register the listeners in the `mounted` hook?

packages/x-components/src/components/__tests__/global-x-bus.spec.ts
```diff
@@ -0,0 +1,84 @@
...
+    const clickedColumnPickerCallback = jest.fn(payload => payload);
+    const { emit } = renderGlobalXBus({
+      listeners: {
+        UserAcceptedAQuery: acceptedAQueryCallback,
+        UserClickedColumnPicker: clickedColumnPickerCallback
+      }
+    });
+
+    emit('UserAcceptedAQuery', 'lego');
+    emit('UserClickedColumnPicker', 2);
```
javieri-empathy, line 29: I would only emit one of those two events, and assert that the other one hasn't been called.

packages/x-components/src/components/global-x-bus.vue
javieri-empathy: I would suggest something much more simple.
```ts
  /**
   * This component helps subscribing to any {@link XEvent} using Vue listeners API.
   *
   * @public
   */
```

javieri-empathy (changes requested):

packages/x-components/src/components/global-x-bus.vue
```diff
@@ -0,0 +1,122 @@
...
+     *
+     * @internal
+     */
+    protected handleEventSubscriptions(): void {
+      forEach(this.listeners, (eventName, callback) => {
+        this.eventSubscriptions[eventName] = this.$x
+          .on(eventName, true)
+          .subscribe(({ eventPayload, metadata }) => {
+            const callbackReturn = callback(eventPayload, metadata);
+            this.$x.emit('SnippetCallbackExecuted', { event: eventName, callbackReturn });
```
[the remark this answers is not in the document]
guillei10: This was @tajespasarela suggestion to complete the task, but I think he was mistaken. I've checked the task and this should be added in the `SnippetCallback` component. Removing it.

packages/x-components/src/components/global-x-bus.vue
```diff
@@ -0,0 +1,122 @@
...
+
+    /**
+     * Subscribes to all the events provided in the listeners with the function that will
+     * execute the callback and will emit a {@link XEventsTypes.SnippetCallbackExecuted} event.
+     * All the subscriptions get stored in {@link GlobalXBus.eventSubscriptions}.
+     *
+     * @internal
+     */
+    protected handleEventSubscriptions(): void {
+      forEach(this.listeners, (eventName, callback) => {
```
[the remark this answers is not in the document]
guillei10: Well I didn't know you could add a subscription in a subscription. This way makes a lot of sense.

packages/x-components/src/components/global-x-bus.vue
```diff
@@ -0,0 +1,122 @@
...
+</script>
+```
+
+## Events
+
+A list of events that the component will emit:
+
+- `SnippetCallbackExecuted`: the event is emitted when any of the provided callbacks is fired
+  because its associated event has been emitted.
+</docs>
```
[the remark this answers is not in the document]
guillei10: Done!

packages/x-components/src/components/__tests__/global-x-bus.spec.ts
```diff
@@ -0,0 +1,53 @@
...
+import { installNewXPlugin } from '../../__tests__/utils';
+import { XComponentBusAPI } from '../../plugins/x-plugin.types';
+import GlobalXBus from '../global-x-bus.vue';
+
+function renderGlobalXBus({ listeners = {} }: RenderGlobalXBusOptions = {}): RenderGlobalXBusAPI {
+  const [, localVue] = installNewXPlugin();
+  const wrapper = mount(GlobalXBus, { listeners, localVue });
+
+  return {
+    emit: wrapper.vm.$x.emit.bind(wrapper.vm.$x)
```
pmareke, line 12: You don't need to bind the emit event to $x, you can just return the emit

```javascript
return {
  emit: wrapper.vm.$x.emit
};
```

pmareke (changes requested):

pmareke (approved):

packages/x-components/src/components/__tests__/global-x-bus.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+interface RenderGlobalXBusOptions {
+  /** The listeners object in the component.*/
+  listeners?: ComponentOptions<Vue>['methods'];
+}
+
+/**
+ * Options to configure how the global X bus component should be rendered.
+ */
+interface RenderGlobalXBusAPI {
+  /** The wrapper for the modal component. */
```
pmareke, line 50: Change comment please.

pmareke (changes requested):

pmareke (approved):

javieri-empathy (approved):

PR #247, opened 2021-11-26
feat(storage-service): Rename StorageService to BrowserStorageService

tajespasarela:
- Add new InMemoryStorageService for testing purposes.
- Add StorageService interface.

EX-5073

# Pull request template
Remove MockedStorageService, created InMemoryStorageService for testing purposes

## Motivation and context
Not use localstorage depending implementation for testing

## Type of change

- [x] `Main`
- [ ] Other. Specify:


pmareke (dismissed):

packages/x-components/src/x-modules/tagging/service/session.service.ts
```diff
@@ -8,24 +8,41 @@ import { SessionService } from './types';
  * @public
  */
 export class DefaultSessionService implements SessionService {
-  protected readonly SESSION_ID_KEY = 'session-id';
+  /**
+   * Session id key to use as key in the storage.
+   *
+   * @public
+   */
+  public readonly SESSION_ID_KEY = 'session-id';
```
javieri-empathy, line 16: Shouldn't this be a static property?

pmareke (dismissed):

packages/x-components/src/x-modules/tagging/service/__tests__/session.service.spec.ts
```diff
@@ -1,46 +1,38 @@
...
-    expect(mockedStorageService.setSessionIdSpy).toHaveBeenCalledWith(
-      'session-id',
-      expect.any(String),
-      expect.any(Number)
-    );
+    expect(getItemSpy).toHaveBeenCalledTimes(1);
+    expect(setItemSpy).toHaveBeenCalledTimes(1);
+    expect(setItemSpy).toHaveBeenCalledWith(storageKey, expect.any(String), expect.any(Number));
     expect(session).toHaveLength(21);
     expect(session).toMatch(/[A-Za-z0-9_-]/);
```
javieri-empathy, line 23: Why these two assertions? Do we have any restriction with the values of the sessionId? I can only think of using safe URL characters, but why the 21 characters length?

packages/x-components/src/x-modules/tagging/service/__tests__/session.service.spec.ts
```diff
@@ -1,46 +1,38 @@
...
-    expect(mockedStorageService.getPrefix()).toBe(prefix);
-    expect(mockedStorageService.getSessionIdSpy).toHaveBeenCalledTimes(1);
-    expect(mockedStorageService.setSessionIdSpy).toHaveBeenCalledTimes(1);
-    expect(mockedStorageService.setSessionIdSpy).toHaveBeenCalledWith(
-      'session-id',
-      'abcd-1234',
-      expect.any(Number)
-    );
-    expect(session).toBe('abcd-1234');
+    sessionService.getSessionId();
```
javieri-empathy, line 27: What's the difference between this and above test? They say different things, but they are testing the same.

packages/x-components/src/x-modules/tagging/service/__tests__/session.service.spec.ts
```diff
@@ -1,46 +1,38 @@
+import { InMemoryStorageService } from '@empathyco/x-storage-service';
 import { DefaultSessionService } from '../session.service';
-import { MockedStorageService } from './utils/mock-storage.service';
 
 describe('testing session id service', () => {
-  const prefix = 'test';
-  const mockedStorageService = new MockedStorageService(prefix);
+  const mockedStorageService = new InMemoryStorageService();
   const sessionService = new DefaultSessionService(mockedStorageService, 1);
```
javieri-empathy, line 6: I don't think this way of testing gives us enough confidence. This test is reusing the same `InMemoryStorageService` for all tests, without resetting it.

javieri-empathy (changes requested):

packages/search-adapter/src/empathy/services/empathy-cache.service.ts
```diff
@@ -10,13 +10,16 @@ import { CacheService } from './cache-service.types';
  */
 @injectable()
 export class EmpathyCacheService implements CacheService {
-
   constructor(
-    @inject(DEPENDENCIES.storageService) protected readonly storageService: StorageService
+    @inject(DEPENDENCIES.storageService) protected readonly storageService: BrowserStorageService
```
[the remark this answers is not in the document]
tajespasarela: Done

packages/storage-service/src/in-memory-storage-service.ts
```diff
@@ -0,0 +1,63 @@
+import { StorageService } from './storage-service';
+
+type Dictionary<T = any> = Record<string, T>;
+
+/**
+ * In memory implementation of the storage service.
+ *
+ * @public
+ */
+export class InMemoryStorageService<Item = any> implements StorageService<Item> {
```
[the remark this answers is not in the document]
tajespasarela: You're right, for a moment I was thinking that all elements had the same type.

packages/x-components/src/utils/storage.ts
```diff
@@ -11,7 +11,7 @@ import { noOp } from './function';
  */
 export const localStorageService =
```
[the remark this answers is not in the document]
tajespasarela: done

packages/x-components/src/x-modules/tagging/service/session.service.ts
tajespasarela: I is exposed for testing purposes, but yes it can be static.

packages/x-components/src/x-modules/tagging/service/__tests__/session.service.spec.ts
tajespasarela: good question. I would only leave the last one.

packages/x-components/src/x-modules/tagging/service/__tests__/session.service.spec.ts
tajespasarela: absolute nothing I think. This is contained above.

packages/x-components/src/x-modules/tagging/service/__tests__/session.service.spec.ts
tajespasarela: reseting it in the beforeEach

packages/x-components/src/x-modules/tagging/service/__tests__/session.service.spec.ts
javieri-empathy: Sorry for my lack of words. I meant that with your change you made this test equal to the above one. In the origin they were testing different things.

With this simple test I think it should be fine. I tried to extend it to also test the TTL which wasn't tested before, but as the `InMemoryStorageService` ignores it, this would imply more changes.

```ts
  it('returns an existing session id', () => {
    const session1 = sessionService.getSessionId();
    const session2 = sessionService.getSessionId();
    expect(session1).toEqual(session2);
  });
```

packages/storage-service/src/browser-storage-service.ts
```diff
@@ -0,0 +1,127 @@
...
+      this.storage.removeItem(key);
+      return ++removedCount;
+    }, 0);
+  }
+
+  protected prefixKey(key: string): string {
+    return `${this.prefix}-${key}`;
+  }
+
+  protected createExpirableItem(item: any, ttlInMs?: number): any {
```
pmareke, line 88: If you're using the `Item = any` in the order methods when the function returns any, please add it here too.

packages/storage-service/src/browser-storage-service.ts
```diff
@@ -0,0 +1,127 @@
...
+      ...(!!ttlInMs && { ttl: ttlInMs + this.currentTimestamp() }),
+      value: item
+    };
+  }
+
+  protected currentTimestamp(): number {
+    return Date.now();
+  }
+
+  protected getItemValue(item: any): any {
```
pmareke, line 100: Same.

pmareke (changes requested):

packages/storage-service/src/browser-storage-service.ts
tajespasarela: no, because we use the generic type only to "cast" the returned item type from outside the service. Inside the service is an any, because it doesn't need the type, the service can store anything

packages/storage-service/src/browser-storage-service.ts
tajespasarela: same as the other comment

pmareke (approved):

javieri-empathy (approved):

packages/x-components/src/x-modules/tagging/service/__tests__/session.service.spec.ts
tajespasarela: added that

PR #248, opened 2021-11-26
feature: Add SnippetCallbacks component.

guillei10:
A new componen has been created called `SnippetCallbacks`, when importing this component and used in a view, it will get all the listeners from the `callbacks` property of the `snippetConfig` and will map them to add the `SnippetCallbackExecuted` emit.

## Motivation and context

This is part of the development that will allow the user to load its own callbacks that will listen to any desired `XEvents` using the `snippetConfig`.

## Type of change

- [x] `Main`

## How has this been tested?

You can try it by yourself in two simple steps:

1. Add a `callbacks` property in the `snippetConfig` with something like:
```
      callbacks: {
        UserAcceptedAQuery: (query) => query,
      }
```
2. Add the component in the view and put a listener in the `SnippetCallbackExecuted` event:
`    <SnippetCallbacks />
`
```
    @XOn('SnippetCallbackExecuted')
    printThings(payload: any): void {
      console.log('event emited with payload:', payload);
    }
```

## Checklist:

### DIFF
--- a/packages/x-components/src/components/__tests__/snippet-callbacks.spec.ts
+++ b/packages/x-components/src/components/__tests__/snippet-callbacks.spec.ts
@@ -8,18 +8,12 @@ function renderSnippetCallbacks({
   callbacks = {}
 }: RenderSnippetCallbacksOptions = {}): RenderSnippetCallbacksAPI {
   const [, localVue] = installNewXPlugin();
-  const wrapper = mount(
-    {
-      components: { SnippetCallbacks },
-      template: '<SnippetCallbacks />'
+  const wrapper = mount(SnippetCallbacks, {
+    provide: {
+      snippetConfig: localVue.observable({ ...baseSnippetConfig, callbacks })
     },
-    {
-      provide: {
-        snippetConfig: localVue.observable({ ...baseSnippetConfig, callbacks })
-      },
-      localVue
-    }
-  );
+    localVue
+  });
 
   return {
     wrapper
@@ -81,8 +75,8 @@ describe('testing SnippetCallbacks component', () => {
  * Options to configure how the snippet callbacks component should be rendered.
  */
 interface RenderSnippetCallbacksOptions {
-  /** The snippet config value. */
-  callbacks?: Partial<XEventListeners>;
+  /** The callbacks value to be provided. */
+  callbacks?: XEventListeners;
 }
 
 /**
--- a/packages/x-components/src/components/__tests__/snippet-callbacks.spec.ts
+++ b/packages/x-components/src/components/__tests__/snippet-callbacks.spec.ts
@@ -1,0 +1,94 @@
+import { mount, Wrapper } from '@vue/test-utils';
+import { installNewXPlugin } from '../../__tests__/utils';
+import { baseSnippetConfig } from '../../views/base-config';
+import { XEventListeners } from '../../x-installer/api/api.types';
+import SnippetCallbacks from '../snippet-callbacks.vue';
+
+function renderSnippetCallbacks({
+  callbacks = {}
+}: RenderSnippetCallbacksOptions = {}): RenderSnippetCallbacksAPI {
+  const [, localVue] = installNewXPlugin();
+  const wrapper = mount(
+    {
+      components: { SnippetCallbacks },
+      template: '<SnippetCallbacks />'
+    },
+    {
+      provide: {
+        snippetConfig: localVue.observable({ ...baseSnippetConfig, callbacks })
+      },
+      localVue
+    }
+  );
+
+  return {
+    wrapper
+  };
+}
+
+describe('testing SnippetCallbacks component', () => {
+  it('executes a callback injected from the snippetConfig', () => {
+    const acceptedAQueryCallback = jest.fn(payload => payload);
+    const clickedColumnPickerCallback = jest.fn(payload => payload);
+    const { wrapper } = renderSnippetCallbacks({
+      callbacks: {
+        UserAcceptedAQuery: acceptedAQueryCallback,
+        UserClickedColumnPicker: clickedColumnPickerCallback
+      }
+    });
+
+    wrapper.vm.$x.emit('UserAcceptedAQuery', 'lego');
+
+    expect(acceptedAQueryCallback).toHaveBeenCalledTimes(1);
+    expect(acceptedAQueryCallback).toHaveBeenCalledWith('lego', {
+      location: undefined,
+      moduleName: null
+    });
+
+    expect(clickedColumnPickerCallback).not.toHaveBeenCalled();
+  });
+
+  it('emits a SnippetCallbackExecuted event when a callback is executed', () => {
+    const acceptedAQueryCallback = jest.fn(payload => payload);
+    const clickedColumnPickerCallback = jest.fn(payload => payload);
+    const { wrapper } = renderSnippetCallbacks({
+      callbacks: {
+        UserAcceptedAQuery: acceptedAQueryCallback,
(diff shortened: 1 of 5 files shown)


packages/x-components/src/components/__tests__/snippet-callbacks.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+import { installNewXPlugin } from '../../__tests__/utils';
+import { baseSnippetConfig } from '../../views/base-config';
+import { XEventListeners } from '../../x-installer/api/api.types';
+import SnippetCallbacks from '../snippet-callbacks.vue';
+
+function renderSnippetCallbacks({
+  callbacks = {}
+}: RenderSnippetCallbacksOptions = {}): RenderSnippetCallbacksAPI {
+  const [, localVue] = installNewXPlugin();
+  const wrapper = mount(
```
pmareke, line 11: You don't need to use the template here

```javascript
  const wrapper = mount(SnippetCallbacks, {
    provide: {
      snippetConfig: localVue.observable({ ...baseSnippetConfig, callbacks })
    },
    localVue
  });
```

packages/x-components/src/components/__tests__/snippet-callbacks.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+      callbackReturn: 3
+    });
+  });
+});
+
+/**
+ * Options to configure how the snippet callbacks component should be rendered.
+ */
+interface RenderSnippetCallbacksOptions {
+  /** The snippet config value. */
```
pmareke, line 84: Review this comment please.

packages/x-components/src/components/snippet-callbacks.vue
```diff
@@ -0,0 +1,86 @@
...
+    /**
+     * Injects {@link SnippetConfig} provided by an ancestor as snippetConfig.
+     *
+     * @internal
+     */
+    @Inject('snippetConfig')
+    public snippetConfig!: SnippetConfig;
+
+    /**
+     * It maps all the callbacks provided by the {@link snippetConfig} and adds an emit to each one.
```
pmareke: `{@link SnippetConfig}`

packages/x-components/src/components/snippet-callbacks.vue
```diff
@@ -0,0 +1,86 @@
...
+              });
+            };
+          })
+        : ({} as XEventListeners);
+    }
+  }
+</script>
+
+<docs lang="mdx">
+## Events
```
pmareke, line 61: Would you mind to show and example with snippet config please?

packages/x-components/src/components/snippet-callbacks.vue
```diff
@@ -0,0 +1,86 @@
+<template>
```
pmareke, line 1: You can extend from NoElement instead of Vue and remove this tag

```html
<template>
  <GlobalXBus v-on="eventListeners" />
</template>
```

`export default class SnippetCallbacks extends NoElement`

packages/x-components/src/x-installer/api/api.types.ts
```diff
@@ -67,6 +78,8 @@ export interface SnippetConfig {
   documentDirection?: DocumentDirection;
   /** The currency name. There should be a currency format associated to this name in the app. */
   currency?: string;
+  /** Callbacks to be triggered when an XEvent is emitted. */
+  callbacks?: XEventListeners;
```
pmareke, line 82: I think this type has to be a `Partial`:

image

pmareke (changes requested):
Really good job, the feature is almost complete!

packages/x-components/src/components/snippet-callbacks.vue
```diff
@@ -0,0 +1,86 @@
...
+     *
+     * @returns The event listeners with the {@link XEventsTypes.SnippetCallbackExecuted} emit in
+     * the callback.
+     *
+     * @internal
+     *
+     */
+    protected get eventListeners(): XEventListeners {
+      const { callbacks } = this.snippetConfig;
+      return callbacks
```
CachedaCodes, line 45: If I'm understanding the `map` function correctly I think that you don't need to check for `callbacks` to be empty or not. If it's empty `map` will return an empty object and the function will return it in the correct type (might we invent, will test)

CachedaCodes (changes requested):

packages/x-components/src/components/snippet-callbacks.vue
javieri-empathy: You shoudn't even need the `NoElement` at all. The `NoElement` should be used when you want to render an slot first child, or when you want to render nothing (and you don't want to write an empty render function). But in this case you are rendering a `GlobalXBus` so it has no place here.

packages/x-components/src/components/__tests__/snippet-callbacks.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    wrapper.vm.$x.emit('UserAcceptedAQuery', 'lego');
+
+    expect(acceptedAQueryCallback).toHaveBeenCalledTimes(1);
+    expect(acceptedAQueryCallback).toHaveBeenCalledWith('lego', {
+      location: undefined,
+      moduleName: null
+    });
+
+    expect(clickedColumnPickerCallback).not.toHaveBeenCalled();
```
CachedaCodes, line 48: After this emit the `UserClickedColumnPicker` and do `toHaveBeenCalledTimes` assertions for both events and `toHaveBeenCalledWith` for the column one to check that all events passed as callbacks would work as expected.

CachedaCodes (changes requested):

packages/x-components/src/components/snippet-callbacks.vue
guillei10: The `callbacks` property could be definitely empty but it could also come as `undefined`. That is why I added the condition. At first I added a, empty object as a default value in the deconstruction but I think it is better to avoid the execution of `map` instead of forcing it with a default.

packages/x-components/src/components/__tests__/snippet-callbacks.spec.ts
guillei10: I don't know if I follow you. In this case we assert the function not no be called because the event is not emitted. This way we are testing that when an event is being emitted only the correct callback is triggered, and not the rest of callbacks.

pmareke (approved):

packages/x-components/src/components/__tests__/snippet-callbacks.spec.ts
CachedaCodes: Here you test that the callback works for one event and doesn't affect the others. I'm suggesting that you emit the other event too to check that the callbacks work for any* event and the callbacks are called only when the event they're linked to is called. Maybe it's too much so adding more is your call.

This is what I had in mind:

```suggestion
    expect(clickedColumnPickerCallback).not.toHaveBeenCalled();
    
    wrapper.vm.$x.emit('UserClickedColumnPicker', 1);

    expect(acceptedAQueryCallback).toHaveBeenCalledTimes(1);
    expect(clickedColumnPickerCallback).toHaveBeenCalledTimes(1);
```

CachedaCodes (changes requested):

packages/x-components/src/components/__tests__/snippet-callbacks.spec.ts
guillei10: Done!

CachedaCodes (approved):

pmareke (approved):

PR #251, opened 2021-11-29
feat(utils): now `origin` only requires the `feature`

tiborux:
This PR removes the `location` to be not required. Only the feature will be required.

EX-5077

### DIFF
--- a/packages/x-components/src/utils/__tests__/origin.spec.ts
+++ b/packages/x-components/src/utils/__tests__/origin.spec.ts
@@ -1,16 +1,10 @@
 import { createOrigin } from '../origin';
 
 describe(`testing ${createOrigin.name} utility method`, () => {
-  it('returns null when the feature or the location is undefined', () => {
+  it('returns null when the feature is undefined', () => {
     expect(
       createOrigin({
         feature: undefined,
-        location: undefined
-      })
-    ).toBeNull();
-    expect(
-      createOrigin({
-        feature: 'history_query',
         location: undefined
       })
     ).toBeNull();
@@ -22,12 +16,21 @@ describe(`testing ${createOrigin.name} utility method`, () => {
     ).toBeNull();
   });
 
-  it('returns `feature:location` when the feature and the location are provided', () => {
+  it('returns `feature:location` when the feature is provided', () => {
     expect(
       createOrigin({
         feature: 'url',
         location: 'results'
       })
     ).toEqual('url:results');
+  });
+
+  it('returns `feature:none` when the feature is provided but the location is not', () => {
+    expect(
+      createOrigin({
+        feature: 'url',
+        location: undefined
+      })
+    ).toEqual('url:none');
   });
 });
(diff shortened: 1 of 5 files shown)


packages/x-components/src/utils/__tests__/origin.spec.ts
```diff
@@ -22,12 +16,21 @@ describe(`testing ${createOrigin.name} utility method`, () => {
     ).toBeNull();
   });
 
-  it('returns `feature:location` when the feature and the location are provided', () => {
+  it('returns `feature:location` when the feature is provided', () => {
```
pmareke, line 19: I don't think this title is ok, as you are saying `the feature is provided` but the  in this case the location is relevant as it's part of the final origin.

packages/x-components/src/x-modules/identifier-results/store/__tests__/actions.spec.ts
```diff
@@ -159,9 +159,6 @@ describe('testing identifier results module actions', () => {
       await store.dispatch('saveOrigin', { location: 'predictive_layer' });
       expect(store.state.origin).toBeNull();
 
-      await store.dispatch('saveOrigin', { feature: 'search_box' });
```
pmareke, line 162: Why did you remove it? Please check the result of this case.

packages/x-components/src/x-modules/search/store/__tests__/actions.spec.ts
```diff
@@ -337,9 +337,6 @@ describe('testing search module actions', () => {
       await store.dispatch('saveOrigin', { location: 'predictive_layer' });
       expect(store.state.origin).toBeNull();
 
-      await store.dispatch('saveOrigin', { feature: 'search_box' });
```
pmareke, line 340: Same here.

pmareke (changes requested):

packages/x-components/src/x-modules/identifier-results/store/__tests__/actions.spec.ts
tiborux: I removed it because right now It makes no sense because you can create an origin with only the feature and this test is to check when it's null.

packages/x-components/src/x-modules/search/store/__tests__/actions.spec.ts
tiborux: I removed it because right now It makes no sense because you can create an origin with only the feature and this test is to check when it's null.

CachedaCodes (approved):

packages/x-components/src/x-modules/identifier-results/store/__tests__/actions.spec.ts
pmareke: But maybe you can create another test but not remove this validation right?

Now you have three different scenarios when you dispatch the action:

- Location
- Feature
- Location + Feature

But you're only testing two of them right?

pmareke (approved):

CachedaCodes (approved):

PR #252, opened 2021-11-29
docs: Remove docusaurus headers

javieri-empathy:
EX-5081

Removes all docusaurus related code from the build.


herrardo (commented):
Just one comment:

In some of their static docs, they are including a header with title and subtitle...

https://raw.githubusercontent.com/empathyco/docs-empathy-documentation/main/functional/about-empathy-platform.md?token=ABDSUWP3AJVDYRLYPP4KBM3BVXNUI

So, should we remove the whole header?

tiborux:
waiting for @rgutialo to check if we need to remove all the headers.

pmareke (approved):

packages/x-components/build/docgen/templates/docgen-component.template.js
```diff
@@ -0,0 +1,24 @@
+function createComponentTemplate(renderedUsage, doc) {
+  const { displayName, description, docsBlocks } = doc;
+  const title = kebabToPascalCase(displayName);
+  return `
+  ---
+  title: ${title}
+  ---
+  # ${title}
+
+  ${description || ''}
```
herrardo, line 10: So I understand that you are overwritting the default template for the component and with this line we avoid the `>` quotation formatting

herrardo: I think we are also fixing the third point of [this task](https://searchbroker.atlassian.net/browse/EX-5063)

herrardo (approved):

packages/x-components/build/docgen/templates/docgen-component.template.js
javieri-empathy: Accidentally, but yes, we should be solving those issues.

pmareke (approved):

PR #254, opened 2021-11-29
feat(search): change `saveOriginWire` to be used when a user interaction event is emitted

tiborux:
This Pr changes wiring in the modules that are setting the saveOrigin using an emitter, for those cases, we have to use a user action, for this change will be: `UserPickedARelatedTag`

EX-5078


pmareke (approved):
I can only find another case, the` identifier results`, and it's already using an User event, so it looks good to me.

tajespasarela (approved):

PR #258, opened 2021-12-03
docs: change h1 to be h2

tiborux:
Changed h1 like #Example to be h2 like ## Example and nest all the titles that should be inside of ## Example adding more # depending on the level.

EX-5064

### DIFF
--- a/packages/x-components/src/components/animations/animate-width.vue
+++ b/packages/x-components/src/components/animations/animate-width.vue
@@ -34,7 +34,7 @@
 </style>
 
 <docs lang="mdx">
-# Examples
+## Examples
 
 The AnimateWidth component is intended to be used as a prop in animatable components but also works
 as a transition to animate the width of an element.
--- a/packages/x-components/src/components/column-picker/base-column-picker-dropdown.vue
+++ b/packages/x-components/src/components/column-picker/base-column-picker-dropdown.vue
@@ -73,17 +73,17 @@
 </script>
 
 <docs lang="mdx">
-# Example
+## Example
 
 Column picker dropdown component renders a dropdown component which options are the different
 columns you can set for a grid.
 
-## Usage
+### Usage
 
 Notice that the slots provided match with the `BaseDropdown` component. The `item` slot is required
 unlike the `toggle`, which renders the same `item` slot defined by default.
 
-### Default usage
+#### Default usage
 
 ```vue
 <template>
--- a/packages/x-components/src/x-modules/scroll/components/window-scroll.vue
+++ b/packages/x-components/src/x-modules/scroll/components/window-scroll.vue
@@ -93,16 +93,16 @@
 </script>
 
 <docs lang="mdx">
-# Example
+## Example
 
 The `WindowScroll` component manages the scroll state of the `body` or `html` elements. It does the
 necessary calculations for knowing the direction of scroll, if the scroll has reached its starting
 position, if it is about to reach its ending position or if it has already reached it end. Whenever
 this state changes, it emits the appropiate X Event to the rest of the application
 
-## Custom usage
+### Custom usage
 
-### Overriding the properties and using document scroll events.
+#### Overriding the properties and using document scroll events.
 
 ```vue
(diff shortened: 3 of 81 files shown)


pmareke (approved):
Nice job!

pmareke (changes requested):
Fix conflicts please.

packages/x-components/src/components/column-picker/base-column-picker-dropdown.vue
```diff
@@ -73,17 +73,17 @@
...
 columns you can set for a grid.
 
-## Usage
+### Usage
 
 Notice that the slots provided match with the `BaseDropdown` component. The `item` slot is required
 unlike the `toggle`, which renders the same `item` slot defined by default.
 
-### Default usage
+#### Default usage
```
CachedaCodes, line 86: I see the two following examples (`Customizing toggle button` & `Using it without v-model / value`) as being under the same category as the `Default usage` example, so I would say to use 4 `#` on them too.

packages/x-components/src/x-modules/scroll/components/window-scroll.vue
```diff
@@ -148,9 +148,9 @@ this state changes, it emits the appropiate X Event to the rest of the applicati
 </script>
 ```
 
-## Customized usage
+### Customized usage
```
CachedaCodes, line 151: Is this one needed? I feel like it could be removed and just have the first `Custom usage` header.

packages/x-components/src/x-modules/scroll/components/window-scroll.vue
```diff
@@ -93,16 +93,16 @@
 </script>
 
 <docs lang="mdx">
```
CachedaCodes, line 95: As we're already here, remove the `:` in the `Vue Events` header at line 208.

packages/x-components/src/components/animations/animate-width.vue
```diff
@@ -34,7 +34,7 @@
 </style>
```
CachedaCodes, line 34: Leaving it here because I don't know where to put it. `redirection.vue` has a header `### Play with the component` that should be `## Play with the component`.

packages/x-components/src/x-modules/search/components/sort-dropdown.vue
```diff
@@ -76,7 +76,7 @@
 </script>
 
 <docs lang="mdx">
-# Sort Dropdown
+## Sort Dropdown
```
CachedaCodes, line 79: Any reason for having a header with the component's name here? When the docs are generated this would result in a page with two headers with the same text (the main one and this one) shouldn't this be moved under examples or is there a specific reason?

packages/x-components/src/x-modules/search/components/sort-list.vue
```diff
@@ -92,7 +92,7 @@
 </script>
 
 <docs lang="mdx">
-# Sort List
+## Sort List
```
CachedaCodes, line 95: Same as with the `sort-dropdown`

CachedaCodes (changes requested):

herrardo (commented):
@CachedaCodes has found that we have the same kind of title with three different names:
```
 # Example
 # Examples
 # See it in action
```
 This is because some time ago, we started correcting and setting the final and good documentation in some of the components, but we didn't do it for all of them.
 
All the components inside `search-box` and `query-suggestions` modules have the final documentation tha follow the[ documentation guidelines](https://searchbroker.atlassian.net/wiki/spaces/ED/pages/3231612987/Guidelines+for+UI+reference) provided by eDocs team.
 
**In this PR we don't have to correct this**, BUT, it would be interesting that in following pull requests if we need to touch any component, we check if it is following the guidelines, and if not, do it. I am setting this as a new part of our DoD.

cc @mlcorton

pmareke (dismissed):

pmareke (approved):

herrardo (approved):

PR #259, opened 2021-12-06
feat(tagging): track the 'UserClickedResultAddToCart'event in tagging.

pmareke:
EX-3529

### DIFF
--- a/packages/x-components/src/x-modules/tagging/wiring.ts
+++ b/packages/x-components/src/x-modules/tagging/wiring.ts
@@ -1,4 +1,5 @@
-import { namespacedDebounce } from '../../wiring';
+import { Result } from '@empathyco/x-types';
+import { mapWire, namespacedDebounce } from '../../wiring';
 import {
   namespacedWireCommit,
   namespacedWireDispatch
@@ -83,6 +84,16 @@ export const trackWire = moduleDebounce(
 );
 
 /**
+ * Tracks the tagging of the add to cart.
+ *
+ * @public
+ */
+export const trackAddToCartWire = mapWire(
+  wireDispatch('track'),
+  ({ tagging }: Result) => tagging.add2cart!
+);
+
+/**
  * Wiring configuration for the {@link TaggingXModule | tagging module}.
  *
  * @internal
@@ -102,5 +113,8 @@ export const taggingWiring = createWiring({
   },
   SearchTaggingChanged: {
     trackWire
+  },
+  UserClickedResultAddToCart: {
+    trackAddToCartWire
   }
 });


pmareke:
Hi @guillei10 as you're on PTO I fixes some conflicts and update the wire as we did with the query tracking.

Please take a look whenever you can!

packages/x-components/src/x-modules/tagging/wiring.ts
```diff
@@ -93,6 +93,16 @@ export const trackResultClickedWire = filter(
   ({ eventPayload: { tagging } }) => !!tagging?.click
 );
 
+/**
+ * Tracks the tagging of the add to cart.
+ *
+ * @public
+ */
+export const trackAddToCartWire = filter(
```
CachedaCodes, line 101: This one and `trackResultClickedWire` use the same code to filter and map, being the `tagging` property the only difference. We should extract that logic.

CachedaCodes (changes requested):

packages/x-components/src/x-modules/tagging/wiring.ts
javieri-empathy: I agree, these events are going to grow, so we can add a factory for this wires in this same file. Here is my proposal:

```ts
function createTrackResultWire(property: keyof Result['tagging']): Wire<Result> {
  return filter(
    mapWire(wireDispatch('track'), ({ tagging }) => tagging[property]!),
    ({ eventPayload: { tagging } }) => !!tagging?.[property]
  );
}
```

packages/x-components/src/x-modules/tagging/wiring.ts
```diff
@@ -93,6 +93,16 @@ export const trackResultClickedWire = filter(
   ({ eventPayload: { tagging } }) => !!tagging?.click
 );
 
+/**
+ * Tracks the tagging of the add to cart.
```
javieri-empathy, line 97: `track` and `tag` are synonyms in this context. My proposal

```suggestion
 * Performs a track of a result added to the cart.
```

javieri-empathy (changes requested):

pmareke:
@javieri-empathy @CachedaCodes I merged the branch with the tests in order to make the build pass.

packages/x-components/src/x-modules/tagging/wiring.ts
```diff
@@ -88,10 +89,29 @@ export const trackQueryWire = moduleDebounce(
  *
  * @public
  */
-export const trackResultClickedWire = filter(
-  mapWire(wireDispatch('track'), ({ tagging }: Result) => tagging.click!),
-  ({ eventPayload: { tagging } }) => !!tagging?.click
-);
+export const trackResultClickedWire = createTrackResultWire('query');
```
javieri-empathy, line 92: Why query? Shouldn't it be click?

javieri-empathy (changes requested):
Change result click wire

javieri-empathy (approved):

CachedaCodes (approved):

PR #260, opened 2021-12-06
Feature/ex 5123 add cancel force support to wire operators

javieri-empathy:
Performs the needed changes to the `debounce` and `throttle` operators, so we are able in a future pull request to track the pending query.
- Refactor `throttle` and `debounce` operators with simpler logic.
- Adds support for forcing and canceling `throttle` and `debounce` operators.
- Updates to RxJS 7.

**⚠️ This pull request contains BREAKING CHANGES ⚠️**

```
feat(wiring): Add support for forcing and canceling `debounce` and `throttle` operators. Update to RxJS 7.

BREAKING CHANGE: `debounce` and `throttle` operators now receive an options object instead of a `raceEvent`. The `options.cancelOn` parameter is equivalent to the previous `raceEvent` one.
BREAKING CHANGE: RxJS has been updated to v7.
EX-5123
```


packages/x-components/package.json
```diff
@@ -67,9 +67,9 @@
     "@empathyco/x-types": "^10.0.0-alpha.13",
     "@types/resize-observer-browser": "~0.1.5",
     "magic-string": "~0.25.7",
-    "nanoid": "^3.1.30",
+    "nanoid": "~3.1.30",
```
javieri-empathy, line 70: Non empathy dependencies should only match patch version.

packages/x-components/src/components/global-x-bus.vue
```diff
@@ -34,11 +34,12 @@
       const subscription = reduce(
         this.$listeners,
         (subscription, eventName, callback) => {
-          return subscription.add(
+          subscription.add(
```
javieri-empathy, line 37: This was the only breaking change of updating to [RxJS 7](https://rxjs.dev/deprecations/breaking-changes)

packages/x-components/src/wiring/__tests__/debounce-operator.spec.ts
```diff
@@ -0,0 +1,195 @@
+import { createResultStub } from '../../__stubs__/index';
```
javieri-empathy, line 1: Because `debounce` and `throttle` operators tests were too long, I decided to extract them into its own files, and make a generic utility `createWire`

packages/x-components/src/wiring/__tests__/debounce-operator.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+    jest.advanceTimersByTime(499);
+    expect(callback).not.toHaveBeenCalled();
+
+    emitWireEvent('3', '4');
+    jest.advanceTimersByTime(499);
+    expect(callback).not.toHaveBeenCalled();
+
+    jest.advanceTimersByTime(1);
+    expect(callback).toHaveBeenCalledTimes(1);
+    expect(callback).toHaveBeenCalledWith('4');
```
javieri-empathy, line 25: I'm not using `toHaveBeenNthCalledWith` on purpose. As I'm testing only one call at a time, I'm doing so immediately after it happens, and each parameter has a different value, I can simplify and not specify the `nth` of each call.

packages/x-components/package.json
diegopf: then we should do the same for:

- `"vue": "^2.6.0"`
- `"vuex": "^3.0.0"`
- `"@vue/vue2-jest": "^27.0.0-alpha.3",`
- `"vue-template-compiler": "^2.6.10"`

shouldn't we?

javieri-empathy: Yes and no.

- Vue and Vuex should be "peer dependencies". We wan't them to be more flexible, and to use the user project version in any case, that's why we specified them with the minor instead of the patch one. We would probably have to re-check Vuex minimum version, as now the installed one is 3.6, and I'm not sure if all the features we are using are available since the 3.0.0.
- Vue template compiler should match Vue version, I would probably change it to `^2.6.0`. Now they end up being the same version, but if we accidentaly install Vue 2.6.1 for example, I think components will fail at compile time or at least throw some errors or warnings.
- Vue2-jest should be changed. I'll change it also in this PR, good catch.

packages/x-components/src/wiring/namespaced-wiring.types.ts
```diff
@@ -22,29 +21,29 @@ import { AnyWire, Wire, WireMetadata } from './wiring.types';
  *
  * @public
  */
-export type NamespacedTimeRetrieving<ModuleName extends XModuleName> = (
+export type NamespacedTimeSelector<ModuleName extends XModuleName> = (
```
diegopf, line 24: Great new name!

packages/x-components/src/wiring/wires-operators.utils.ts
```diff
@@ -0,0 +1,69 @@
...
+export function mergeEvents(events: MaybeArray<XEvent>, on: XBus['on']): Observable<void> {
+  const subject = new Subject<void>();
+  const eventsList = Array.isArray(events) ? events : [events];
+
+  /* Can't use RxJS `merge` function, as it immediately emits previously emitted values to new
+   subscriptions due to the `ReplaySubject` of the bus. With this Subject we are still receiving
+   those values immediately on subscription, but as there are no subscribers before the `return`
+   happens we are fine. */
+  eventsList.forEach(raceEvent => on(raceEvent).subscribe(() => subject.next()));
+  return subject;
```
diegopf, line 27: I was thinking on a pure rxjs solution but couldn't find one yet. However I think it should be possible...  🤔

diegopf (changes requested):

packages/x-components/src/wiring/__tests__/debounce-operator.spec.ts
```diff
@@ -0,0 +1,195 @@
...
+      const { wire, callback, emitWireEvent, registerWire, bus } = createWire();
+      registerWire(
+        debounce(wire, 500, { forceOn: 'UserClickedAResult', cancelOn: 'UserClearedQuery' })
+      );
+      emitWireEvent('1', '2');
+      expect(callback).not.toHaveBeenCalled();
+
+      bus.emit('UserClearedQuery', '');
+      bus.emit('UserClickedAResult', createResultStub('Random result'));
+      expect(callback).toHaveBeenCalledTimes(0);
```
tiborux, line 189: Is there differences between .not.toHaveBeenCalled(); and .toHaveBeenCalledTimes(0);?

packages/x-components/src/wiring/__tests__/wires-operators.spec.ts
```diff
@@ -129,188 +126,18 @@ describe('testing wires operators', () => {
     });
   });
 
-  describe('testing timing operators', () => {
-    beforeAll(jest.useFakeTimers);
-    afterAll(jest.useRealTimers);
+  describe('testing operator ' + mapWire.name, () => {
+    test(mapWire.name + ' emits the valued transformed by the function passed by parameter', () => {
```
tiborux, line 130: Why we are using test instead of it?

tiborux (commented):
Good job!

packages/x-components/src/wiring/__tests__/throttle-operator.spec.ts
```diff
@@ -0,0 +1,233 @@
+import { createResultStub } from '../../__stubs__/results-stubs.factory';
+import { throttle } from '../wires.operators';
+import { createWire } from './operators-testing-utils';
+
+describe(`testing ${throttle.name} operator`, () => {
```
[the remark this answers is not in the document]
javieri-empathy: We some times do this to avoid hardcoding the name of the function in the middle of a string. This way if this function is renamed at some time (which I would probably do to `throttleWire`), the test description is also updated, while if we hardcode the `throttle` name I can guarantee that we will forget 🤣.

packages/x-components/src/wiring/__tests__/debounce-operator.spec.ts
javieri-empathy: The programmer's lack of care 🤣. I think `.not.toHaveBeenCalled()` is more readable

packages/x-components/src/wiring/__tests__/wires-operators.spec.ts
javieri-empathy: We should use `test` when we can't write a valid sentence with the description, but as [Jest says](https://jestjs.io/docs/api#testname-fn-timeout), they are both exactly the same, it is just for readability.

packages/x-components/package.json
javieri-empathy: I've just checked [vue2-jest](https://github.com/vuejs/vue-jest/commit/6edfa9cd93be05ca4c9c41849db668dbf2ae92d5), and they are treating alphas as release candidates 🤣,  so we should be safer than I thought.

tiborux (dismissed):

diegopf (approved):

tiborux (approved):

PR #262, opened 2021-12-07
test(e2e): make url tests intercept tracking and wait for each request

javieri-empathy:
EX-5127

URL e2e tests where failing some times due to 2 reasons:

- Tracking was not being intercepted
- Search responses were mixed, so some times the `the search request contains the origin` step was doing the assertion over the previous request. This was specially noticeable with a 6x cpu slowdown.

### DIFF
--- a/packages/x-components/tests/e2e/cucumber/url/url.feature
+++ b/packages/x-components/tests/e2e/cucumber/url/url.feature
@@ -20,7 +20,7 @@ Feature: Url component
     Then  the search request contains the origin "url:external"
     When  sort option "<sort>" is selected from the sort "dropdown"
     Then  the search request contains the origin "url:external"
-    Then  navigate back
+    When  navigating back
     Then  the search request contains the origin "<origin>"
     Examples:
       | sort      | origin          |
@@ -29,8 +29,8 @@ Feature: Url component
   Scenario Outline: 3. Navigate back from pdp to serp sets the url origin as "<origin>"
     Given a URL with query parameter "lego"
     Then  the search request contains the origin "url:external"
-    Then  click result in position 0
-    Then  navigate back
+    When  click result in position 0
+    And   navigating back
     Then  the search request contains the origin "<origin>"
     Examples:
       | origin              |
--- a/packages/x-components/tests/e2e/cucumber/url/url.feature
+++ b/packages/x-components/tests/e2e/cucumber/url/url.feature
@@ -6,10 +6,10 @@ Feature: Url component
     And   a next queries API
     And   a suggestions API
     And   a related tags API
+    And   a tracking API
 
   Scenario Outline: 1. Navigating to a URL from the outside sets the url origin as "<origin>"
     Given a URL with query parameter "lego"
-    And   waiting for search request intercept with new origin
     Then  the search request contains the origin "<origin>"
     Examples:
       | origin        |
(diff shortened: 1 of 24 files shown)


packages/x-components/tests/e2e/cucumber/url/url.feature
```diff
@@ -6,19 +6,20 @@ Feature: Url component
...
     Then  the search request contains the origin "<origin>"
     Examples:
-      | origin        |
-      | url:external  |
+      | origin       |
+      | url:external |
 
   Scenario Outline: 2. Navigate back and forth in serp sets the url origin as "<origin>"
     Given a URL with query parameter "lego"
+    Then  the search request contains the origin "url:external"
```
pmareke, line 20: Should we use a variable for the first origin too?

packages/x-components/tests/e2e/cucumber/url/url.feature
```diff
@@ -6,19 +6,20 @@ Feature: Url component
...
     When  sort option "<sort>" is selected from the sort "dropdown"
-    And   waiting for search request intercept with new origin
+    Then  the search request contains the origin "url:external"
     Then  navigate back
```
pmareke, line 23: I think this should be `When` instead of `Then` as it's a user interaction.

packages/x-components/tests/e2e/cucumber/url/url.feature
```diff
@@ -27,8 +28,8 @@ Feature: Url component
 
   Scenario Outline: 3. Navigate back from pdp to serp sets the url origin as "<origin>"
     Given a URL with query parameter "lego"
+    Then  the search request contains the origin "url:external"
     Then  click result in position 0
```
pmareke, line 32: I think this should be `When` instead of `Then` as it's a user interaction.

packages/x-components/tests/e2e/cucumber/url/url.feature
```diff
@@ -27,8 +28,8 @@ Feature: Url component
...
-    And   waiting for search request intercept with new origin
     Then  navigate back
```
pmareke, line 33: I think this should be `When` instead of `Then` as it's a user interaction.

pmareke (changes requested):

packages/x-components/tests/e2e/cucumber/url/url.feature
javieri-empathy: I wouldn't. I think it is a lot more confusing to see things like `firstOrigin`, `secondOrigin`, rather than the hardcoded value. I don't even like to have tables if tests only have 1 example 🤣.

packages/x-components/tests/e2e/cucumber/url/url.feature
```diff
@@ -6,30 +6,31 @@ Feature: Url component
...
       | sort      | origin          |
       | price asc | url:url_history |
 
   Scenario Outline: 3. Navigate back from pdp to serp sets the url origin as "<origin>"
     Given a URL with query parameter "lego"
-    Then  click result in position 0
-    And   waiting for search request intercept with new origin
-    Then  navigate back
+    Then  the search request contains the origin "url:external"
+    When  click result in position 0
```
mavmaf, line 32: Change step name here or in the .spec.ts, they are not matching 
`click result in position 0` vs `clicking result in position {int}`

mavmaf (changes requested):

packages/x-components/tests/e2e/cucumber/url/url.feature
javieri-empathy: Fuck, done.

mavmaf (approved):

pmareke (approved):

PR #263, opened 2021-12-08
Feature/ex 4331 make  api work with computed props

javieri-empathy:
EX-4331

Changes XComponents initialisation to be on the `beforeCreated` hook instead of the `created` one. This should allow us to run `immediate` watchers that access the `$x` API, without having to rely on the `created` hook.


packages/x-components/src/components/__tests__/x-component-utils.spec.ts
```diff
@@ -1,38 +1,30 @@
...
+  }
   it('sets and gets as XComponent name the passed name parameter to the passed component', () => {
-    setXComponentXModuleName(testXComponentWrapper, 'searchBox');
-    expect(getXComponentXModuleName(testXComponentWrapper)).toBe('searchBox');
+    expect(getXComponentXModuleName(renderComponent('searchBox').vm)).toBe('searchBox');
   });
 
   it('returns null if the passed component is not an XComponent', () => {
-    expect(getXComponentXModuleName(testXComponentWrapper)).toBeNull();
+    expect(getXComponentXModuleName(renderComponent().vm)).toBeNull();
```
javieri-empathy, line 20: Now the component stores the module name in its options object. There is no need to do that in the `beforeCreate` hook like previously.

packages/x-components/src/plugins/__tests__/x-plugin-mixin.spec.ts
```diff
@@ -95,106 +77,149 @@ describe('testing $x component API global mixin', () => {
...
-    };
-
-    mount(
-      {
-        mixins: [xComponentMixin(searchBoxXModule)],
-        created() {
-          this.$x.on('UserIsTypingAQuery', true).subscribe(listener);
+      const emitterComponent: ComponentOptions<any> & ThisType<Vue> = {
+        mounted() {
+          this.$x.emit('UserIsTypingAQuery', 'Sexy Lego', { target: testTarget });
```
guillei10, line 107: Who is the mentally ill developer that decided `Sexy Lego` was a proper query to test with? 🤣

packages/x-components/src/components/x-component.mixin.ts
```diff
@@ -1,7 +1,6 @@
 import Vue, { ComponentOptions } from 'vue';
 import { XPlugin } from '../plugins/x-plugin';
 import { AnyXModule } from '../x-modules/x-modules.types';
-import { setXComponentXModuleName } from './x-component.utils';
 
 /**
  * Initializes a component as an X-Component:
```
guillei10, line 6: Does this function still do all the things written in the doc?

packages/x-components/src/plugins/__tests__/x-plugin-mixin.spec.ts
tiborux: jijiijij

packages/x-components/src/plugins/__tests__/x-plugin-mixin.spec.ts
```diff
@@ -1,32 +1,13 @@
-import { mount, shallowMount, Wrapper } from '@vue/test-utils';
+import { mount, Wrapper } from '@vue/test-utils';
```
tiborux, line 1: Can we move the render function outside of the describe and the interfaces too? Like  we have it in the other specs

tiborux (commented):
Gj man

packages/x-components/src/plugins/__tests__/x-plugin-mixin.spec.ts
javieri-empathy: Are you talking about `renderOriginComponent`? I wouldn't, that testing factory only covers tests for the origin, which are encapsulated inside the `describe('testing origin',` block. In other tests we use a single render function for all tests, but this plugin mixin test needs different components for each one.

packages/x-components/src/components/x-component.mixin.ts
javieri-empathy: Yes, it is still registering the module, and yes, setting the `xModule` property in the options object is the new way of flagging a component as an x-component.

packages/x-components/src/plugins/__tests__/x-plugin-mixin.spec.ts
javieri-empathy: I see you complain about Sexy Lego, but you said nothing about Sexy playmobil, are you a playmobil fanboy?

packages/x-components/src/plugins/__tests__/x-plugin-mixin.spec.ts
tiborux: yes, you are right, I didn't think on that. Thank you, you make me see the light

tiborux (approved):

guillei10 (approved):

PR #264, opened 2021-12-09
feat(components): fix 'BaseRating' component styles

guillei10:
## Description

This is a very simple PR. It changes the styles of the BaseRating component in order to fit images not only horizontally but vertically also. This happens when using icons that somehow are bigger than the `line-height` and `font-size` being applied by the design tokens.

### DIFF
--- a/packages/x-components/src/components/base-rating.vue
+++ b/packages/x-components/src/components/base-rating.vue
@@ -85,14 +85,14 @@
     display: inline-block;
 
     &--empty {
-      overflow-x: hidden;
+      overflow: hidden;
       display: flex;
     }
 
     &--filled {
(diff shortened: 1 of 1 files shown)


javieri-empathy (approved):

packages/x-components/src/components/base-rating.vue
```diff
@@ -85,14 +85,14 @@
     display: inline-block;
```
javieri-empathy, line 85: I think we should also remove this `inline-block`, and replace the `span` that we render by a `div`, which defaults to `block`. This way we should also avoid problems related to the line-height.

diegopf (approved):

PR #265, opened 2021-12-10
fix(extraParams): prevent callbacks option from the snippet config to be taken as extra params EX-5104

diegopf:
## Motivation and context

In a previous feature we included the option to configure callbacks on the snippet config. Currently every option on the snippet config is set as extra params, leading to the problem that this callbacks object is sent on every request to the backend. In order to avoid this I have added a field to the SnippetConfigExtraParams to control which options from the snippet config we want to prevent to be taken as extra params. 

- [x] Open issue. If applicable, link: [EX-5103](https://searchbroker.atlassian.net/browse/EX-5103)

## Type of change

- [x] `Main`
- [ ] Other. Specify:

## How has this been tested?

I added some callbacks configuration to the snippet config on the project's demo app. I also added the SnippetConfigExtraParams component to reproduce the bug. Afterwards I have check that with my change the callbacks are not sent anymore. Besides I added a unit test.

## Checklist:

### DIFF
--- a/packages/x-components/src/x-modules/extra-params/components/snippet-config-extra-params.vue
+++ b/packages/x-components/src/x-modules/extra-params/components/snippet-config-extra-params.vue
@@ -41,6 +41,13 @@
     protected extraParams: Dictionary<unknown> = {};
 
     /**
+     * Collection of properties from the snippet config not allowed to be sent as extra params.
+     *
+     * @internal
+     */
+    protected notAllowedExtraParams: Array<keyof SnippetConfig> = ['callbacks'];
+
+    /**
      * Updates the extraParams object when the snippet config changes.
      *
      * @param snippetConfig - The new snippet config.
@@ -59,6 +66,9 @@
       ...snippetExtraParams
     }: SnippetConfig): void {
       forEach(snippetExtraParams, (name, value) => {
+        if (this.notAllowedExtraParams.includes(name)) {
+          return;
+        }
         this.$set(this.extraParams, name, value);
       });
     }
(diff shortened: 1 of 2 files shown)


guillei10 (commented):
Looks good to me. Let's see if the second review finds something wrong, if not you have my approve ✅

packages/x-components/src/x-modules/extra-params/components/snippet-config-extra-params.vue
```diff
@@ -59,6 +66,9 @@
       ...snippetExtraParams
     }: SnippetConfig): void {
       forEach(snippetExtraParams, (name, value) => {
+        if (this.notAllowedExtraParams.includes(name)) {
```
tiborux, line 69: I was trying to find a "clean" solution because I dont like to make an empty return but i dont like the other solution either:

`  if (!this.notAllowedExtraParams.includes(name)) {
          this.$set(this.extraParams, name, value);
    }`
    
 I dont know if you have more suggestions hahah if not, i will give my approve

tiborux (changes requested):

packages/x-components/src/x-modules/extra-params/components/snippet-config-extra-params.vue
guillei10: I think either way is correct. The one he choose is more _backendish_

tiborux: Yes, I prefer to use always positive statements, so lets go ahead with this solution

tiborux (approved):

guillei10 (approved):

PR #266, opened 2021-12-10
feat(scroll): remove animations on pending element to scroll

tiborux:
EX-5025


diegopf:
While reviewing this I've noticed that the this feature is not working on Firefox unless the some margin is set as prop. See screenshot.
image

I think the fix should be pretty easy and is only adjust the default value for the prop margin to ["0px 0px 0px 0px"  ](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver/rootMargin#value) instead of an empty string. IMO we could do it within this PR but we could also create another bug ticket.  What do you think?

diegopf (commented):
see https://github.com/empathyco/x/pull/266#issuecomment-992534521

packages/x-components/tests/unit/main-scroll.spec.ts
```diff
@@ -166,6 +169,44 @@ describe('testing MainScroll component', () => {
...
+        const template = `
+        <MainScroll v-bind="{ threshold, margin, useWindow }">
+          <div data-test="main-scroll">
+            <FadeAndSlide>
+              <MainScrollItem
+                v-for="item in items"
+                class="item"
+                tag="article"
+                :item="item"
+                :data-test="item.id">
```
javieri-empathy, line 182: Add a `:key="item.id"` so the transition-group works properly.

packages/x-components/tests/unit/main-scroll.spec.ts
```diff
@@ -166,6 +169,44 @@ describe('testing MainScroll component', () => {
...
+        cy.log('Item 5 should be the first one.');
+        getItem(5).should($item5 => {
+          expect($item5.get(0).getBoundingClientRect().top).to.be.eq(0);
+        });
+        cy.log("Item 4 shouldn't be visible");
+        getItem(4).should($item4 => {
+          const item4Bounds = $item4.get(0).getBoundingClientRect();
+          const item4Bottom = item4Bounds.top + item4Bounds.height;
+          expect(item4Bottom).to.be.eq(0);
+        });
```
javieri-empathy, line 207: There is an issue with this part of the test, and it is that it still works without the new CSS you added disabling the transitions. The reason appears to be that Cypress executes it immediately, before the transition is run.

I think you can fix it by just replacing the `getItem(5)` code with this one which will wait until it is visible (opacity: 1) before checking its position.

```ts
getItem(5)
            .should('be.visible')
            .should($item5 => {
              expect($item5.get(0).getBoundingClientRect().top).to.be.eq(0);
            });
```

If you do this, then the `x-main-scroll--no-transition` class checks are unnecessary, as you are testing implementation details, how the scroll is restored when there are transitions instead of that the scroll is properly restored like you are doing below. I would be fine with testing the class if there was no other way to test it, but I'd rather rely on testing that the scroll is restored properly like you are already doing. Moreover these class checks are failing for me and I'm not able to make them work, while I see them passing in the CI, so I have no confidence at all 🤣. 

image

packages/x-components/tests/unit/main-scroll.spec.ts
```diff
@@ -166,6 +169,44 @@ describe('testing MainScroll component', () => {
         });
       });
 
+      it('restores the scroll with transitions enabled', () => {
+        const template = `
+        <MainScroll v-bind="{ threshold, margin, useWindow }">
+          <div data-test="main-scroll">
+            <FadeAndSlide>
```
javieri-empathy, line 176: I would replace the `FadeAndSlide` with the `StaggeredFadeAndSlide`, which is the one that gives us more problems due to the delay.

packages/x-components/tests/unit/main-scroll.spec.ts
```diff
@@ -166,6 +169,44 @@ describe('testing MainScroll component', () => {
         });
       });
 
+      it('restores the scroll with transitions enabled', () => {
+        const template = `
```
javieri-empathy, line 173: If you are going to use a variable only once, usually it is better to inline it in the function that uses it.

javieri-empathy (changes requested):

diegopf (dismissed):

packages/x-components/src/x-modules/scroll/components/main-scroll.vue
```diff
@@ -55,7 +63,7 @@
      *
      * @public
      */
-    @Prop({ default: '' })
+    @Prop({ default: undefined })
     public margin!: string;
```
javieri-empathy, line 67: 👮 Types police officer here.
If you are going to leave the property optional, then you should flag it with `?` instead of `!` to ensure that the code that uses it knows that it can be `undefined`. Also, props are optional by default, so you don't need to specify the `default: undefined`.

```suggestion
    @Prop()
    public margin?: string;
```

Or, if you prefer @diegopf suggestion and avoid working with possible `undefined` values, this should work. Here you can still use the `!` operator because you are always guaranteeing to work with a `string`.

```suggestion
    @Prop({ default: '0px' })
    public margin!: string;
```

javieri-empathy (changes requested):

diegopf (approved):

javieri-empathy (approved):

PR #267, opened 2021-12-13
feat(tagging): track pending query

javieri-empathy:
EX-4963

Forces the query tracking to happen immediately in these situations:

- When a search result is clicked.
- When a promoted is clicked.
- When a banner is clicked.
- When a redirection is triggered.
- When a the user triggers the infinite scroll.

This is needed to ensure that we properly track each search request upon any user interaction with the current search response. Otherwise we might lose certain events, as the user might leave the SERP (like on click events), or load a new page (with the infinite scroll).

### DIFF
--- a/packages/x-components/src/adapter/mocked-adapter.ts
+++ b/packages/x-components/src/adapter/mocked-adapter.ts
@@ -16,6 +16,7 @@ import {
   TopRecommendationsResponse,
   TrackingRequest
 } from '@empathyco/x-adapter';
+import { noOp } from '../utils/index';
 import { configureAdapterWithToysrus } from './util';
 
 declare global {
@@ -71,7 +72,10 @@ class E2ETestsAdapter extends EmpathyAdapter {
   }
 
   track(request: TrackingRequest): Promise<void> {
-    return mockFetch(request, 'track');
+    return fetch(request.url, {
+      method: 'POST',
+      body: JSON.stringify(request.params)
+    }).then(noOp);
   }
 
   searchById(request: SearchByIdRequest): Promise<SearchByIdResponse> {
--- a/packages/x-components/src/x-modules/search/components/banner.vue
+++ b/packages/x-components/src/x-modules/search/components/banner.vue
@@ -32,8 +32,9 @@
     public banner!: BannerModel;
 
     /**
-     * Emits the promoted click event.
+     * Emits the banner click event.
      *
+     * @internal
      */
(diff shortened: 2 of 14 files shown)


packages/x-components/src/x-modules/tagging/wiring.ts
```diff
@@ -81,7 +81,16 @@ export const setSessionDuration = wireCommit('setSessionDuration');
...
+  {
+    cancelOn: 'UserClearedQuery',
+    forceOn: [
+      'UserClickedAResult',
+      'UserClickedAPromoted',
+      'UserClickedABanner',
+      'UserClickedARedirection',
+      'UserReachedResultsListEnd'
+    ]
+  }
```
javieri-empathy, line 93: This is the actual change, the rest are test related changes. I think I've added any event that might override the query tagging while still consider the search response as accepted.

packages/x-components/src/x-modules/search/components/banner.vue
```diff
@@ -30,6 +30,14 @@
      */
     @Prop({ required: true })
     public banner!: BannerModel;
+
+    /**
+     * Emits the promoted click event.
```
tiborux, line 35:
```suggestion
     * Emits the banner click event.
```

tiborux (changes requested):
GJ, only a little change in the doc

packages/x-components/src/adapter/mocked-adapter.ts
```diff
@@ -71,7 +72,10 @@ class E2ETestsAdapter extends EmpathyAdapter {
   }
 
   track(request: TrackingRequest): Promise<void> {
-    return mockFetch(request, 'track');
+    return fetch(request.url, {
+      method: 'POST',
+      body: JSON.stringify(request.params)
+    }).then(noOp);
```
tajespasarela, line 78: why the `.then(noOp)`?

tajespasarela (changes requested):

packages/x-components/src/adapter/mocked-adapter.ts
javieri-empathy: `SearchAdapter` specifies that `track` method should return `Promise<void>`, but `fetch` returns a `Promise<Response>`. With the `noOp` we are simply ignoring that `Response` and "transforming" it into `void`

tajespasarela (approved):

tiborux (approved):

PR #270, opened 2021-12-14
test(e2e): fix tagging tests

javieri-empathy:
EX-5161

This should fix tagging e2e tests which could fail on slower machines. Previously throttling the tests it could happen that the tracking calls were already made before cypress had time to wait for them, so it couldn't see them.

This also happened in the infinite scroll query tagging test, where some times it was expecting the first query tagging call to be the second one, or it could even fail scrolling down and loading the next page.

### DIFF
--- a/packages/x-components/tests/e2e/cucumber/tagging/tagging.spec.ts
+++ b/packages/x-components/tests/e2e/cucumber/tagging/tagging.spec.ts
@@ -37,10 +37,7 @@ Then('query tagging request has been triggered', () => {
 });
 
 Then('second page query tagging request is triggered', () => {
-  cy.wait('@queryTagging', { timeout: 10000 })
-    .its('request.body')
-    .then(JSON.parse)
-    .should('have.property', 'page', 2);
+  cy.wait('@queryTagging').its('request.body').then(JSON.parse).should('have.property', 'page', 2);
 });
 
 Then('results page number {int} is loaded', (page: number) => {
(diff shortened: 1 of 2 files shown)


packages/x-components/tests/e2e/cucumber/tagging/tagging.spec.ts
```diff
@@ -21,17 +21,28 @@ When('first redirection is clicked', () => {
 });
 
 When('scrolls down to next page', () => {
-  cy.getByDataTest('result-link').last().scrollIntoView();
+  cy.getByDataTest('result-link').last().scrollIntoView({ easing: 'linear', duration: 2000 });
 });
 
 Then('result click tagging request is triggered', () => {
-  cy.wait('@clickTagging');
+  cy.get('@clickTagging').should('exist');
```
javieri-empathy, line 28: Click tagging should happen immediately, therefore the `get` instead of `wait`

packages/x-components/tests/e2e/cucumber/tagging/tagging.spec.ts
```diff
@@ -21,17 +21,28 @@ When('first redirection is clicked', () => {
...
 });
 
-Then('query tagging request is triggered', () => {
+Then('query tagging request should be triggered', () => {
```
javieri-empathy, line 31: This assertion is talking about the future,  therefore the `wait`

packages/x-components/tests/e2e/cucumber/tagging/tagging.spec.ts
```diff
@@ -21,17 +21,28 @@ When('first redirection is clicked', () => {
...
   cy.wait('@queryTagging');
 });
 
+Then('query tagging request has been triggered', () => {
```
javieri-empathy, line 35: This assertion is talking about the past, when the tagging request should happen immediately, therefore the `get`

tiborux (approved):

tajespasarela (approved):

PR #273, opened 2021-12-15
feat(tagging): Include 'location' parameter to track result events.

tajespasarela:
EX-5163

This PR adds the `location` parameter injected by the `LocationProvider` component, to the result tagging event (click).

## Type of change

- [x] `Main`
- [ ] Other. Specify:

## How has this been tested?

- E2E test
- Manual test

### DIFF
--- a/packages/x-components/cypress.json
+++ b/packages/x-components/cypress.json
@@ -8,7 +8,7 @@
   "pluginsFile": "tests/plugins/index.js",
   "supportFile": "tests/support/index.ts",
   "integrationFolder": "tests/e2e",
-  "testFiles": ["cucumber/**/*.features"],
+  "testFiles": ["cucumber/**/*.feature"],
   "fixturesFolder": "tests/e2e/fixtures",
   "screenshotsFolder": "tests/e2e/screenshots",
   "component": {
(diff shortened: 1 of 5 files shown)


packages/x-components/cypress.json
```diff
@@ -8,7 +8,7 @@
   "pluginsFile": "tests/plugins/index.js",
   "supportFile": "tests/support/index.ts",
   "integrationFolder": "tests/e2e",
-  "testFiles": ["cucumber/**/*.features"],
+  "testFiles": ["cucumber/**/*.feature"],
```
diegopf, line 11: is this change intended? I understood that we should stick to *.features because is faster on headless mode

diegopf (commented):
Looks good to me. However I think we should clarify https://github.com/empathyco/x/pull/273/files#r769548556

packages/x-components/cypress.json
tiborux: it should be a mistake

tajespasarela: Ups, I forgot to change back. 
Fixed now

tiborux (approved):

diegopf (approved):

PR #275, opened 2021-12-17
feat(desing-system): Add new color tokens.

tajespasarela:
EX-5171

Add new token colors: Accent, Enable, Disable

# Pull request template
Add this new colors to tokens and to Desing System Utilities

## Motivation and context
These new colors are necessary for BSH setup.
Also refactored a little bit how the utilities classes are generated and the examples in the Design Syste view.

## Type of change

### DIFF
--- a/packages/x-components/src/design-system/utilities/background-color.scss
+++ b/packages/x-components/src/design-system/utilities/background-color.scss
@@ -1,23 +1,7 @@
-.x-background {
-  &--lead {
-    background-color: var(--x-color-base-lead) !important;
-  }
-  &--auxiliary {
-    background-color: var(--x-color-base-auxiliary) !important;
-  }
-  &--neutral-10 {
-    background-color: var(--x-color-base-neutral-10) !important;
-  }
-  &--neutral-35 {
-    background-color: var(--x-color-base-neutral-35) !important;
-  }
-  &--neutral-70 {
-    background-color: var(--x-color-base-neutral-70) !important;
-  }
-  &--neutral-95 {
-    background-color: var(--x-color-base-neutral-95) !important;
-  }
-  &--neutral-100 {
-    background-color: var(--x-color-base-neutral-100) !important;
+@import '../base/colors.tokens';
+
+@each $color-name, $color-value in $colors {
+  .x-background--#{$color-name} {
+    background-color: var(--x-color-base-#{$color-name}) !important;
   }
 }
--- a/packages/x-components/src/design-system/utilities/background-color.scss
+++ b/packages/x-components/src/design-system/utilities/background-color.scss
@@ -1,4 +1,4 @@
-@import '../base/colors.tokens';
+@import '../variables';
 
 @each $color-name, $color-value in $colors {
   .x-background--#{$color-name} {
--- a/packages/x-components/src/design-system/utilities/box-shadow.scss
+++ b/packages/x-components/src/design-system/utilities/box-shadow.scss
@@ -4,7 +4,7 @@
   box-shadow: none !important;
 }
 .x-shadow {
-  @for $i from 0 through 10 {
+  @for $i from 1 through 10 {
     $i: zerofill($i);
     &--#{$i} {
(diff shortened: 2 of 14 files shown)


packages/x-components/src/design-system/utilities/background-color.scss
```diff
@@ -1,23 +1,7 @@
...
-  }
-  &--neutral-70 {
-    background-color: var(--x-color-base-neutral-70) !important;
-  }
-  &--neutral-95 {
-    background-color: var(--x-color-base-neutral-95) !important;
-  }
-  &--neutral-100 {
-    background-color: var(--x-color-base-neutral-100) !important;
+@import '../base/colors.tokens';
```
javieri-empathy, line 1: While I love the idea of automating this (finally), importing token files from other files will generate the variables each time it is imported. Extracting the size and color maps to files which only contain pure SCSS variables should be enough.

image
image
image

packages/x-components/src/design-system/utilities/box-shadow.scss
```diff
@@ -4,7 +4,7 @@
   box-shadow: none !important;
 }
 .x-shadow {
-  @for $i from 1 through 10 {
+  @for $i from 0 through 10 {
```
javieri-empathy, line 7: Why did you add a `shadow-0`? you have a `shadow-none` just above

packages/x-components/src/views/design-system/design-system-utilities.vue
```diff
@@ -504,7 +345,41 @@
...
+      'enable',
+      'disable'
+    ];
+
+    protected maxSize = 20;
+
+    protected get sizes(): string[] {
+      return Array.from({ length: this.maxSize + 1 }).map((_, index) =>
+        String(index).padStart(2, '0')
+      );
```
javieri-empathy, line 367: Just FYI, you can pass the `map` as the second `Array.from` parameter:

```suggestion
      return Array.from({ length: this.maxSize + 1 }, (_, index) => String(index).padStart(2, '0'));
```

packages/x-components/src/views/design-system/design-system-utilities.vue
```diff
@@ -504,7 +345,41 @@
   import { Component } from 'vue-property-decorator';
 
   @Component
-  export default class DesignSystemUtilities extends Vue {}
+  export default class DesignSystemUtilities extends Vue {
```
javieri-empathy, line 348: On behalf of my CPU, thanks

javieri-empathy (changes requested):

packages/x-components/src/views/design-system/design-system-utilities.vue
tajespasarela: wonderful! thanks

packages/x-components/src/views/design-system/design-system-utilities.vue
tajespasarela: little step by little step

packages/x-components/src/design-system/utilities/box-shadow.scss
tajespasarela: Didn't realize, changed

javieri-empathy (approved):

packages/x-components/src/design-system/utilities/background-color.scss
javieri-empathy: Even if I don't know how I managed to make `@use` work, it still doesn't help us. If we have our variables or mixins defined in a file which renders real CSS content, `rollup-plugin-styles` will duplicate these styles as many times as they are imported.

pmareke (approved):
I'm not sure if I'm understanding right but why the title accent color is different?

image

Should we add in the Typography also the enable and disabled?

image

pmareke (changes requested):
Change accent color in typography and add missing enable and disable.

pmareke (approved):

PR #276, opened 2021-12-20
fix(search): rows and start parameters calculations moved from getter to action to avoid multiple requests

LuisMartinez15:
EX-5047

## Motivation and context
There is a bug with the request pagination, infinite scroll and search input causing multiple requests to be made. The `request` getter depends on `isAppendResults` triggering the getter, thus emitting a new `SearchRequestChanged`.

## Type of change

## How has this been tested?

Added unit tests.

## Checklist:

### DIFF
--- a/packages/x-components/src/x-modules/search/store/__tests__/actions.spec.ts
+++ b/packages/x-components/src/x-modules/search/store/__tests__/actions.spec.ts
@@ -57,14 +57,30 @@ describe('testing search module actions', () => {
   });
 
   describe('fetchAndSaveSearchResponse', () => {
-    it('should include the origin in the request', async () => {
+    it('should include the origin, start and rows properties in the request', async () => {
       resetSearchStateWith(store, { query: 'lego', origin: 'search_box:external' });
+      const { page, ...restRequest } = store.getters.request!;
       await store.dispatch('fetchAndSaveSearchResponse', store.getters.request);
 
       expect(adapter.search).toHaveBeenCalledTimes(1);
       expect(adapter.search).toHaveBeenCalledWith({
-        ...store.getters.request,
-        origin: 'search_box:external'
+        ...restRequest,
+        origin: 'search_box:external',
+        start: 0,
+        rows: 24
+      });
+    });
+
+    it('should calculate correctly the start and rows properties', async () => {
+      resetSearchStateWith(store, { query: 'lego', config: { pageSize: 48 } });
+      const { page, ...restRequest } = store.getters.request!;
+      await store.dispatch('fetchAndSaveSearchResponse', store.getters.request);
+
+      expect(adapter.search).toHaveBeenCalledTimes(1);
+      expect(adapter.search).toHaveBeenCalledWith({
+        ...restRequest,
+        start: 0,
+        rows: 24
       });
     });
 
--- a/packages/x-components/src/x-modules/search/store/__tests__/actions.spec.ts
+++ b/packages/x-components/src/x-modules/search/store/__tests__/actions.spec.ts
@@ -72,15 +72,20 @@ describe('testing search module actions', () => {
     });
 
     it('should calculate correctly the start and rows properties', async () => {
-      resetSearchStateWith(store, { query: 'lego', config: { pageSize: 48 } });
+      resetSearchStateWith(store, {
+        config: { pageSize: 48 },
+        page: 2,
+        query: 'lego',
+        results: getResultsStub(48)
+      });
       const { page, ...restRequest } = store.getters.request!;
       await store.dispatch('fetchAndSaveSearchResponse', store.getters.request);
 
       expect(adapter.search).toHaveBeenCalledTimes(1);
       expect(adapter.search).toHaveBeenCalledWith({
         ...restRequest,
-        start: 0,
-        rows: 24
+        start: 48,
+        rows: 48
       });
     });
 
--- a/packages/x-components/src/x-modules/search/store/__tests__/getters.spec.ts
+++ b/packages/x-components/src/x-modules/search/store/__tests__/getters.spec.ts
@@ -1,11 +1,9 @@
 import Vue from 'vue';
 import Vuex, { Store } from 'vuex';
-import { SearchRequest } from '@empathyco/x-adapter';
 import { map } from '../../../../utils';
+import { PageableSearchRequest } from '../../types';
 import { searchXStoreModule } from '../module';
 import { SearchState } from '../types';
-import { getSelectedRelatedTagsStub } from '../../../../__stubs__/related-tags-stubs.factory';
-import { getSimpleFilterStub } from '../../../../__stubs__/filters-stubs.factory';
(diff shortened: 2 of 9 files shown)


packages/x-components/src/x-modules/search/store/__tests__/actions.spec.ts
```diff
@@ -57,14 +57,30 @@ describe('testing search module actions', () => {
...
+    it('should calculate correctly the start and rows properties', async () => {
+      resetSearchStateWith(store, { query: 'lego', config: { pageSize: 48 } });
+      const { page, ...restRequest } = store.getters.request!;
+      await store.dispatch('fetchAndSaveSearchResponse', store.getters.request);
+
+      expect(adapter.search).toHaveBeenCalledTimes(1);
+      expect(adapter.search).toHaveBeenCalledWith({
+        ...restRequest,
+        start: 0,
+        rows: 24
```
CachedaCodes, line 83: The test is failing. Receives `48` instead of `24`.

packages/x-components/src/x-modules/search/types.ts
```diff
@@ -0,0 +1,11 @@
+import { SearchRequest } from '@empathyco/x-adapter';
+
+/**
+ * A pageable search request.
+ *
+ * @public
+ */
+export interface PageableSearchRequest extends SearchRequest {
```
CachedaCodes, line 8: In the declaration of `SearchRequest` it extends from `PageableRequest`, so creating the type `PageableSearchRequest` to make it actually pageable is kind of weird. I would say we can change the type `PageableRequest` to include the `page` field directly and get rid of this new type.

packages/x-components/src/x-modules/search/store/actions/fetch-and-save-search-response.action.ts
```diff
@@ -51,6 +48,34 @@ const { fetchAndSave, cancelPrevious } = createFetchAndSaveActions<
...
+    results
+  } = state;
+  const start = page === 1 ? 0 : results.length;
+
+  return {
+    ...restRequest,
+    // eslint-disable-next-line @typescript-eslint/no-extra-parens
+    ...(origin && { origin }),
+    start,
+    rows: pageSize * page! - start
```
CachedaCodes, line 75: `page` is an optional field in `PageableSearchRequest`, but we have to tell TS that it will have value when it might not be true. 
If we'll always have `page`, it shouldn't be optional in the type declaration. If not, then we have to handle unexpected values for it. In this line and in line 68 the logic breaks if `page` is note defined

CachedaCodes (changes requested):

packages/x-components/src/x-modules/search/store/__tests__/actions.spec.ts
```diff
@@ -57,14 +57,30 @@ describe('testing search module actions', () => {
...
-        origin: 'search_box:external'
+        ...restRequest,
+        origin: 'search_box:external',
+        start: 0,
+        rows: 24
+      });
+    });
+
+    it('should calculate correctly the start and rows properties', async () => {
+      resetSearchStateWith(store, { query: 'lego', config: { pageSize: 48 } });
```
pmareke, line 75: As Guille mentioned please change here to 24.

packages/x-components/src/x-modules/search/store/__tests__/actions.spec.ts
```diff
@@ -57,14 +57,30 @@ describe('testing search module actions', () => {
...
-        ...store.getters.request,
-        origin: 'search_box:external'
+        ...restRequest,
+        origin: 'search_box:external',
+        start: 0,
+        rows: 24
+      });
+    });
+
+    it('should calculate correctly the start and rows properties', async () => {
```
pmareke, line 74: I'll add another test for the case of start bigger than 0.

packages/x-components/src/x-modules/search/store/__tests__/getters.spec.ts
```diff
@@ -18,52 +16,24 @@ describe('testing search module getters', () => {
...
 
-      expect(store.getters[gettersKeys.request]).toEqual<SearchRequest>({
+      expect(store.getters[gettersKeys.request]).toEqual<PageableSearchRequest>({
         query: 'doraemon',
         relatedTags: [],
         filters: {},
         sort: '',
-        rows: 24,
-        start: 0,
+        page: 1,
```
pmareke, line 32: should we add another test for pages bigger than one?

packages/x-components/src/x-modules/search/store/actions/fetch-and-save-search-response.action.ts
pmareke: I think page is not optional as it will be as least `1` (default value in the store)

packages/x-components/src/x-modules/search/types.ts
pmareke: It's for me too. When I saw that `SearchRequest` extends `PageableRequest` was really weird.

pmareke (changes requested):

packages/x-components/src/x-modules/search/store/actions/fetch-and-save-search-response.action.ts
CachedaCodes: If it's not possible for `page` to reach this function without value then it's all good 👍

packages/x-components/src/x-modules/search/store/__tests__/getters.spec.ts
LuisMartinez15: Done

packages/x-components/src/x-modules/search/store/__tests__/actions.spec.ts
LuisMartinez15: Fixed. I put `48`, which was the initial intention here.

packages/x-components/src/x-modules/search/store/__tests__/actions.spec.ts
LuisMartinez15: As we are already testing `start: 0` in the previous test, I changed the `start` value here to something > 0.

packages/x-components/src/x-modules/search/types.ts
LuisMartinez15: I've renamed it to `InternalSearchRequest`, as this type is just some sort of "partial/incomplete" (not a `Typescript` `partial`) `SearchRequest` containing a `page` property, which is just used to ease the calculations of the `SearchRequest` properties `start` and `rows`.

packages/x-components/src/x-modules/search/store/actions/fetch-and-save-search-response.action.ts
LuisMartinez15: Done

pmareke (approved):

CachedaCodes (approved):

PR #277, opened 2021-12-21
fix(components): Fix infinite scroll not working in Android Chrome.

tajespasarela:
EX-5181

We don't know why, but a rootMargin over ~1700000 in the IntersectionObserver doesn't work in Android Chrome


pmareke (commented):
I can't try it on an Android device but at least the iPhone devices keep working.

Good job.

pmareke (approved):

tiborux (approved):

PR #280, opened 2021-12-22
feat(tagging): add new SearchTaggingEmitted event for tracking the query

diegopf:
EX-5170

 

## Motivation and context

Provide a new event _SearchTaggingEmitted_ that can be listened on the snippet config callbacks in order to give the customer the chance to track the query on its own. From now on _SearchTaggingChanged_ only stores the query tagging info in the store with a debounce. Modifying this value on the store will emit _SearchTaggingEmitted_ and also will dispatch the query tracking

- [x] Open issue. If applicable, link: [EX-5170](https://searchbroker.atlassian.net/browse/EX-5170)

## Type of change

- [x] `Main`
- [ ] Other. Specify:

## How has this been tested?

Import SnippetCallbacks component and add the callback configuration: 
` 
  callbacks: {
    SearchTaggingEmitted: payload => console.log(payload)
  }
`

Check that the number of track requests is the same as the calls to the callback.

## Checklist:

### DIFF
--- a/packages/x-components/src/x-modules/tagging/events.types.ts
+++ b/packages/x-components/src/x-modules/tagging/events.types.ts
@@ -26,7 +26,7 @@ export interface TaggingXEvents {
    * Query tagging info has changed.
    * * Payload: The new query tagging info.
    */
-  SearchTaggingEmitted: TaggingInfo;
+  SearchTaggingReceived: TaggingInfo;
   /**
    * Session duration has been provided.
    * * Payload: The new session duration value.
--- a/packages/x-components/src/x-modules/tagging/events.types.ts
+++ b/packages/x-components/src/x-modules/tagging/events.types.ts
@@ -1,10 +1,11 @@
+import { TaggingInfo } from '@empathyco/x-types';
+
 /**
  * Dictionary of the events of Tagging XModule, where each key is the event name, and the value is
  * the event payload type or `void` if it has no payload.
  *
  * @public
  */
-import { TaggingInfo } from '@empathyco/x-types';
 
 export interface TaggingXEvents {
   /**
--- a/packages/x-components/src/x-modules/tagging/events.types.ts
+++ b/packages/x-components/src/x-modules/tagging/events.types.ts
@@ -27,7 +27,6 @@ export interface TaggingXEvents {
    * * Payload: The new session duration value.
    */
   SessionDurationProvided: number;
-
   /**
    * Query tagging info has changed.
    * * Payload: The new query tagging info.
--- a/packages/x-components/src/x-modules/tagging/events.types.ts
+++ b/packages/x-components/src/x-modules/tagging/events.types.ts
@@ -6,7 +6,6 @@ import { TaggingInfo } from '@empathyco/x-types';
  *
  * @public
  */
-
 export interface TaggingXEvents {
   /**
    * Consent has changed.
@@ -24,13 +23,13 @@ export interface TaggingXEvents {
    */
   QueryTaggingDebounceProvided: number;
   /**
-   * Session duration has been provided.
-   * * Payload: The new session duration value.
-   */
-  SessionDurationProvided: number;
-  /**
    * Query tagging info has changed.
    * * Payload: The new query tagging info.
    */
   SearchTaggingEmitted: TaggingInfo;
+  /**
+   * Session duration has been provided.
+   * * Payload: The new session duration value.
+   */
+  SessionDurationProvided: number;
 }
--- a/packages/x-components/src/x-modules/tagging/events.types.ts
+++ b/packages/x-components/src/x-modules/tagging/events.types.ts
@@ -4,6 +4,8 @@
  *
  * @public
  */
+import { TaggingInfo } from '@empathyco/x-types';
+
 export interface TaggingXEvents {
   /**
    * Consent has changed.
@@ -25,4 +27,10 @@ export interface TaggingXEvents {
    * * Payload: The new session duration value.
    */
   SessionDurationProvided: number;
+
+  /**
+   * Query tagging info has changed.
+   * * Payload: The new query tagging info.
+   */
+  SearchTaggingEmitted: TaggingInfo;
 }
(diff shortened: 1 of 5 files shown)


packages/x-components/src/x-modules/tagging/events.types.ts
```diff
@@ -25,4 +28,9 @@ export interface TaggingXEvents {
    * * Payload: The new session duration value.
    */
   SessionDurationProvided: number;
+  /**
```
pmareke, line 31: Put the events form A to Z please.

```suggestion
  /**
   * Query tagging info has changed.
   * * Payload: The new query tagging info.
   */
  SearchTaggingEmitted: TaggingInfo;
  /**
   * Session duration has been provided.
   * * Payload: The new session duration value.
   */
  SessionDurationProvided: number;
```

packages/x-components/src/x-modules/tagging/events.types.ts
```diff
@@ -1,9 +1,12 @@
+import { TaggingInfo } from '@empathyco/x-types';
+
 /**
  * Dictionary of the events of Tagging XModule, where each key is the event name, and the value is
  * the event payload type or `void` if it has no payload.
  *
  * @public
  */
+
```
pmareke, line 9: Remove blank line please.

```suggestion
```

packages/x-components/src/x-modules/tagging/store/emitters.ts
```diff
@@ -7,5 +7,6 @@ import { taggingXStoreModule } from './module';
  * @internal
  */
 export const taggingEmitters = createStoreEmitters(taggingXStoreModule, {
-  ConsentChanged: state => state.consent!
+  ConsentChanged: state => state.consent!,
+  SearchTaggingEmitted: state => state.queryTaggingInfo!
```
pmareke, line 11: I know is not related to the PR but we need to avoid import from barrels (line 1)

```javascript
import { createStoreEmitters } from '../../../store/utils/store-emitters.utils';
```

packages/x-components/src/x-modules/tagging/wiring.ts
```diff
@@ -145,12 +152,15 @@ export const taggingWiring = createWiring({
...
   SearchTaggingChanged: {
-    trackQueryWire
+    setQueryTaggingInfo
   },
   UserClickedAResult: {
     trackResultClickedWire
   },
   UserClickedResultAddToCart: {
     trackAddToCartWire
+  },
```
pmareke, line 162: Put the event from A to Z please.

```
  ConsentProvided: {
    setConsent
  },
  ConsentChanged: {
    clearSessionWire
  },
  QueryTaggingDebounceProvided: {
    setQueryTaggingDebounce
  },
  SearchTaggingChanged: {
    setQueryTaggingInfo
  },
  SearchTaggingEmitted: {
    trackQueryWire
  },
  SessionDurationProvided: {
    setSessionDuration
  },
  UserClickedAResult: {
    trackResultClickedWire
  },
  UserClickedResultAddToCart: {
    trackAddToCartWire
  }
```

packages/x-components/src/x-modules/tagging/events.types.ts
```diff
@@ -25,4 +28,9 @@ export interface TaggingXEvents {
    * * Payload: The new session duration value.
    */
   SessionDurationProvided: number;
+  /**
+   * Query tagging info has changed.
+   * * Payload: The new query tagging info.
+   */
+  SearchTaggingEmitted: TaggingInfo;
```
pmareke, line 35: I'm not sure about the name.

I mean this event will be emitted when the `TaggingInfo` change (at the same time this change is done when the search tagging changes)

image

I would rather use something like `TaggingInfoChanged`, which for me express better what changes.

What do you think?

pmareke (changes requested):
Nice job @diegopf , really well done!

javieri-empathy (changes requested):

packages/x-components/src/x-modules/tagging/wiring.ts
```diff
@@ -145,12 +152,15 @@ export const taggingWiring = createWiring({
...
+  SearchTaggingEmitted: {
+    trackQueryWire
```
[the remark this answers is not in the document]
diegopf: I was really interested about this so I made some research about it:

According to the spec, during the process of navigate to a new resource the active document is aborted ([step 15](https://html.spec.whatwg.org/multipage/browsing-the-web.html#navigate)) and aborting a document should cancel any instances of fetch ([step 2](https://html.spec.whatwg.org/multipage/browsing-the-web.html#abort-a-document)) 

The [Fetch API](https://developer.mozilla.org/en-US/docs/Web/API/fetch) provides the option `keepalive` to prevent a request from being cancelled. [Chrome 80](https://developers.google.com/web/updates/2019/12/chrome-80-deps-rems) recommends this approach or using the Beacon API in case you want to make a request when the document is about to be unloaded, since they disabled the option to do a synchronous XHR.

Google analytics docs describes a way to [specify the transport mechanism](https://developers.google.com/analytics/devguides/collection/analyticsjs/sending-hits#specifying_different_transport_mechanisms) where you can set the beacon API. They even mentioned that the other supported methods face the issue: 

> hits are often not sent if the page is being unloaded

I run a local (dumb) experiment to see if it was possible to reach a server with a request juste before the unload of the document, see results:

  | Chrome | Safari | Firefox | Opera | Vivaldi
-- | -- | -- | -- | -- | --
Fetch | ✅ | ❌ | ❌ | ✅ | ✅
Fetch + keep alive | ✅ | ❌ | ❌ | ✅ | ✅
Beacon | ✅ | ✅ | ✅ | ✅ | ✅

Given that the main idea of this callback event is to allow the consumer of x-components to use its own tracking system, I think we are safe here.

packages/x-components/src/x-modules/tagging/events.types.ts
diegopf: I was not sure about SearchTaggingEmitted so I just changed it to SearchTaggingReceived which seems to be quite better to me. Anyway, we can have a call tomorrow about it.

pmareke (approved):

packages/x-components/src/x-modules/tagging/wiring.ts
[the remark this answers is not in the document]
javieri-empathy: Thanks once again for this research. I knew about the possibility of being cancelled, but TBH I hadn't read it from the spec 🤣. Knowing that GA allows to use beacon is also a great discovery, so I agree, we should be safe.

javieri-empathy (approved):

PR #282, opened 2021-12-23
Feature/ex 5189 reset scroll on more events

javieri-empathy:
EX-5189

- Exposes a prop to change the events that should reset the scroll position. For example, when searching, changing filters, related tags. 
- Changing an extra param has also been added as a trigger for resetting the scroll position
- Adds e2e tests for this feature.

### DIFF
--- a/packages/x-components/src/components/scroll/base-scroll.vue
+++ b/packages/x-components/src/components/scroll/base-scroll.vue
@@ -28,10 +28,10 @@
 <docs lang="mdx">
 ## Example
 
-The BaseScroll is a component that manage the states of scroll of a specified element. The component
-does the necessary calculations for knowing the direction of scroll, if the scroll has reached to
-start or to end, and is about to reaching to end. The components emits the next events depending of
-movement that realize the user:
+The `BaseScroll` is a component that manages the state of scroll of a specified element. The
+component does the necessary calculations for knowing the direction of scroll, if the scroll has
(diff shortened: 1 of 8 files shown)


packages/x-components/src/components/scroll/base-scroll.vue
```diff
@@ -117,18 +109,44 @@ Set to false the reset scroll on query change feature which is true by default.
...
-        console.log('scroll:at-end');
+      }
+    }
+  };
+</script>
+```
+
+### Reset scroll
+
+You can configure which events make the scroll position be reset using the `resetOn` prop.
```
LuisMartinez15, line 120:
```suggestion
You can configure which events reset the scroll position using the `resetOn` prop.
```

LuisMartinez15 (dismissed):

packages/x-components/src/components/scroll/scroll.mixin.ts
```diff
@@ -1,6 +1,7 @@
 import Vue from 'vue';
 import { Component, Prop, Watch } from 'vue-property-decorator';
 import { throttle } from '../../utils/throttle';
+import { XEvent } from '../../wiring/index';
```
pmareke, line 4:
```suggestion
import { XEvent } from '../../wiring/events.types';
```

pmareke (changes requested):

LuisMartinez15 (approved):

pmareke (approved):

PR #286, opened 2021-12-27
fix(components): `SlidingPanel` can now disable the resetting scroll …

javieri-empathy:
EX-5197

This is just a simple fix to allow disabling the sliding panel reset on content change behavior. This feature was added to auto-reset the scroll position to the start when the related tags, which are one of the common components that are painted inside a `SlidingPanel`,  change from one query to another.

However this is not always the desired behavior. Another common case is to paint results inside the `SlidingPanel`. Results are far more complex components than the related tags and its content might change for different reasons. For example its image is loaded, we have an integration with a third party sevice that makes it render some custom component, the results have some kind of special interaction that makes it render new nodes (i.e. a dropdown for the size or color selector)... In this cases we do not want to reset the scroll position.

The simplest solution I could come up with is to add this prop to disable the behavior. I didn't want to tie this component to certain events, as the surfacing API is far worse for the end user than just a boolean prop.

`SlidingPanel` tests have also been migrated to Cypress component testing, as they were relying on mocking browser APIs and styles.


packages/x-components/src/components/sliding-panel.vue
```diff
@@ -39,70 +39,111 @@
...
     mounted(): void {
-      this.scrollObserver.observe(this.$refs.scrollContainer, {
-        attributes: false,
-        childList: true,
-        subtree: true,
-        characterData: false
+      const resizeObserver = new ResizeObserver(this.debouncedUpdateScrollPosition);
+      resizeObserver.observe(this.$el);
+
+      const contentChangedObserver = new MutationObserver(this.restoreAndUpdateScroll);
```
javieri-empathy, line 122: Browser APIs like `ResizeObserver` or `MutationObserver` shouldn't be used before the `mounted` hook.

packages/x-components/src/components/sliding-panel.vue
```diff
@@ -39,70 +39,111 @@
...
+     * @public
      */
     @Prop({ default: true })
-    protected showButtons!: boolean;
+    public showButtons!: boolean;
 
     /**
-     * MutationObserver that watches for changes inside the wrapping div to update
-     * the scroll position.
+     *
```
diegopf, line 68: Missing documentation for the prop.

packages/x-components/src/components/sliding-panel.vue
```diff
@@ -39,70 +39,111 @@
...
+              subtree: true,
+              characterData: false
+            });
+          } else {
+            contentChangedObserver.disconnect();
+          }
+        },
+        { immediate: true }
+      );
+      this.$on('hook:beforeDestroy', () => {
```
diegopf, line 139: I'm new to Vue so I'm not sure but should't we call $off when the component is destroyed to prevent memory leaks here?

diegopf (changes requested):

packages/x-components/src/components/sliding-panel.vue
javieri-empathy: That's a good question. AFAIK Vue cleans this listeners itself. In the doc I found a quite similar example on what I'm doing here: https://vuejs.org/v2/guide/components-edge-cases.html#Programmatic-Event-Listeners. If you prefer we can change it by `$once`, but I think it will have the same effect, as the component can only be destroyed once.

To give you more confidence, I wrote a little test to check it:

```ts
it('does things', async () => {
  const destroyed = jest.fn();
  const wrapper = mount(
    {
      template: `<Test v-if="shouldRender"/>`,
      components: {
        Test: Vue.extend({
          template: `<div />`,
          mounted() {
            this.$on('hook:beforeDestroy', destroyed);
          }
        })
      },
      props: ['shouldRender']
    },
    {
      propsData: {
        shouldRender: false
      }
    }
  );
  await wrapper.setProps({ shouldRender: true });
  expect(destroyed).not.toHaveBeenCalled();
  await wrapper.setProps({ shouldRender: false });
  await wrapper.setProps({ shouldRender: true });
  expect(destroyed).toHaveBeenCalledTimes(1);
  await wrapper.setProps({ shouldRender: false });
  await wrapper.setProps({ shouldRender: true });
  expect(destroyed).toHaveBeenCalledTimes(2);
});
```

diegopf: Thanks for the explanation! I was looking for something like that in the docs but did not find it. IMO it would be better change it to $once in order to avoid that someone has the same doubt in the future. Besides, in the docs link you shared they are using $once, I think it is just more clear.

diegopf (approved):

packages/x-components/src/components/sliding-panel.vue
```diff
@@ -39,70 +39,113 @@
...
-   * Horizontal slide panel component. The content that is provided to this component would be
-   * navigable horizontally using two navigational buttons.
+   * This component allows for any other component or element inside it to be horizontally
+   * navigable. It also implements customizable buttons as well as other minor customizations to its
+   * general behavior.
    *
    * @public
    */
   @Component
+  /* eslint-disable @typescript-eslint/unbound-method */
```
pmareke, line 49: Should we move this rule to the lines with the problem (120 and 123)?

packages/x-components/src/components/sliding-panel.vue
```diff
@@ -39,70 +39,113 @@
...
   export default class SlidingPanel extends Vue {
     /**
      * Scroll factor that will dictate how much the scroll moves when pressing a navigation button.
+     *
+     * @public
      */
     @Prop({ default: 0.7 })
-    protected scrollFactor!: number;
+    public scrollFactor!: number;
```
pmareke, line 57: Why are you making it public (also `showButtons`)?

pmareke (changes requested):
That's impressive capo, nice PR.

packages/x-components/tests/unit/sliding-panel.spec.ts
```diff
@@ -0,0 +1,304 @@
...
+        duration: 500
+      });
+    },
+    setScrollWidth(widthPx) {
+      cy.getByDataTest('sliding-panel-scroll').then($scroll => {
+        $scroll.css('width', `${widthPx}px`);
+      });
+    },
+    setItemsCount(itemsCountCount) {
+      cy.getByDataTest('items-count').type(itemsCountCount.toString());
```
pmareke, line 105: I don't like the `.toString()`, unpopular opinion, I'd rather use 
```javascript
.type(`${itemsCountCount}`)
```

packages/x-components/tests/unit/sliding-panel.spec.ts
```diff
@@ -0,0 +1,304 @@
...
+      return cy.getByDataTest('sliding-panel');
+    },
+    getLeftButton() {
+      return cy.getByDataTest('sliding-panel-left-button');
+    },
+    getRightButton() {
+      return cy.getByDataTest('sliding-panel-right-button');
+    },
+    getScrollPosition() {
+      return cy.getByDataTest('sliding-panel-scroll').invoke('scrollLeft');
```
pmareke, line 91: Why a función which names implies getting the position of the scroll is a call to a method which mades a scroll and also returns void?
```javascript
protected scrollLeft(): void {
  this.scrollTo(-this.$refs.scrollContainer.clientWidth);
}
```

Where this position comes?

pmareke (changes requested):

pmareke (approved):

diegopf (approved):

PR #288, opened 2022-01-03
fix(design-system): Fix Sliding Panel styles.

tajespasarela:
BREAKING-CHANGE: Remove tokens `--x-color-background-sliding-panel-middle` and `--x-color-background-sliding-panel-end`.

Fix an issue with the z-index in the sliding panel:
image

Also changed the gradient styles to use the CSS `mask` instead of linear gradient. This way we can control the gradient color with the same token as the sliding panel background.

### DIFF
--- a/packages/x-components/src/design-system/components/sliding-panel/default.scss
+++ b/packages/x-components/src/design-system/components/sliding-panel/default.scss
@@ -8,15 +8,13 @@
   &__button.x-button {
     // size
     --x-size-padding-button: var(--x-size-padding-sliding-panel-button);
+
+    // other
+    z-index: 1;
   }
 
   &.x-sliding-panel--buttons-overflow {
     .x-sliding-panel {
-      &__button {
-        // others
-        pointer-events: all;
-      }
-
       &__button-left {
         // layout
         transform: translateX(
@@ -43,9 +41,6 @@
       .x-sliding-panel__button-left {
         // visibility
         opacity: 1;
-
-        // others
-        pointer-events: all;
       }
     }
 
@@ -53,9 +48,6 @@
       .x-sliding-panel__button-right {
         // visibility
         opacity: 1;
-
-        // others
-        pointer-events: all;
       }
     }
   }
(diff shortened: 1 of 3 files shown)


packages/x-components/src/design-system/components/sliding-panel/default.scss
```diff
@@ -78,25 +84,15 @@
 
     &::before {
       // color
-      background: linear-gradient(
-        to right,
-        var(--x-color-background-sliding-panel) 43%,
-        var(--x-color-background-sliding-panel-middle) 67%,
-        var(--x-color-background-sliding-panel-end)
-      );
+      mask: linear-gradient(to right, black 43%, rgba(0, 0, 0, 0.8) 67%, rgba(0, 0, 0, 0));
```
javieri-empathy, line 87: When I was talking with you about this I was thinking in applying the mask to the `__scroll` element, instead of the `before` or `after` elements so we don't have to care about the background anymore or those elements.

javieri-empathy (approved):

packages/x-components/src/design-system/components/sliding-panel/default.scss
```diff
@@ -1,4 +1,7 @@
 .x-sliding-panel {
```
javieri-empathy, line 1: There are a few pointer-events: all that you can now remove, both in `sliding-panel.vue` and `sliding-panel/default.scss`

javieri-empathy (changes requested):

packages/x-components/src/design-system/components/sliding-panel/default.scss
tajespasarela: We must keep them, the buttons only change the opacity, and we have to avoid the click when opacity is 0

javieri-empathy: Yep, my bad, didn't realise about this.

javieri-empathy (approved):

CachedaCodes (approved):

PR #290, opened 2022-01-04
fix(design-system): Fix grid styles to stretch cells.

tajespasarela:
EX-5239

The idea is to stretch the `grid__item`s inside the `x-grid`, to allow set all the items with the same height.


CachedaCodes (changes requested):
e2e scroll and tagging tests are failing because now results are smaller and more fit in the screen.

CachedaCodes:
Unrelated to the PR and more of a question than anything but why there isn't an example of `x-grid` in the design system layout?

PR #296, opened 2022-01-06
fix(identifier-results): Fix default plugin config.

tajespasarela:
[EX-5247]

The Identifier Result doesn't work when X-components are served.

The problem is the `baseInstallXOptions` were not used, because it was always overridden by the config coming from URL (empty object {} if not present), then the Identifier Result Regexp was not right.

- Refactored the Identifier Results in the layout.
- Refactored the Identifier Results E2E tests.

[EX-5247]: https://searchbroker.atlassian.net/browse/EX-5247?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/x-components/tests/e2e/cucumber/identifier-results/identifier-results.feature
+++ b/packages/x-components/tests/e2e/cucumber/identifier-results/identifier-results.feature
@@ -3,32 +3,20 @@ Feature: Identifier results component
   Background:
     Given following config: identifier detection Regexp "^[0-9]{2,}$"
 
-  Scenario Outline: 1. ID search with results is made
+  Scenario: 1. ID search shows results and clears them when a non id query is made
     Given an ID results API with a known response
     And   start button is clicked
-    When  a "<query>" with results is typed
+    When  a "01" with results is typed
     Then  identifier results are displayed
-
-    Examples:
-      | query |
-      | 01    |
-
-  Scenario Outline: 2. ID search with no results is made
-    Given an ID results API with no results
-    And   start button is clicked
-    When  a "<query>" with results is typed
+    When  a "lego" with results is typed
     Then  no identifier results are displayed
 
-    Examples:
-      | query |
-      | 23    |
 
-  Scenario Outline: 3. No ID search is made
+  Scenario: 2. ID search shows results and clears them when a id query with no results is made
     Given an ID results API with a known response
     And   start button is clicked
-    When  a "<query>" with results is typed
+    When  a "01" with results is typed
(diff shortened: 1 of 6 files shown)


diegopf (approved):

packages/x-components/tests/e2e/cucumber/identifier-results/identifier-results.feature
```diff
@@ -22,7 +16,7 @@ Feature: Identifier results component
   Scenario Outline: 2. ID search with no results is made
```
javieri-empathy, line 16: The `2` and `3` scenarios will still pass if a request is made and it returns results, as the `Then no identifier results are displayed` is executed immediately after typing, and at that moment there will never be any painted results. I suggest combining it with the happy path with results, and assert that the identifier results are reset in both cases.

```feature
  Scenario: 1. ID search shows results and clears them when a non id query is made
    Given an ID results API with a known response
    And   start button is clicked
    When  a "01" with results is typed
    Then  identifier results are displayed
    When  a "lego" with results is typed
    Then  no identifier results are displayed


  Scenario: 2. ID search shows results and clears them when a id query with no results is made
    Given an ID results API with a known response
    And   start button is clicked
    When  a "01" with results is typed
    Then  identifier results are displayed
    Given an ID results API with no results
    When  a "23" with results is typed
    Then  no identifier results are displayed
```

javieri-empathy (changes requested):

packages/x-components/tests/e2e/cucumber/identifier-results/identifier-results.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+Given(
+  'following config: identifier detection Regexp {string}',
+  (identifierDetectionRegexp: string) => {
+    const config = {
+      identifierResults: {
+        config: {
+          identifierDetectionRegexp
+        }
+      }
+    } as XPluginOptions['xModules'];
```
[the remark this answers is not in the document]
tajespasarela: done

packages/x-components/tests/e2e/cucumber/identifier-results/identifier-results.feature
tajespasarela: changed

javieri-empathy (approved):

diegopf (approved):

PR #301, opened 2022-01-13
fix(design-system): Fix some style issues and add two missing utilities.

tajespasarela:
EX-5270

This is just 3 things:

Little fix with the style in the base result rating component
Refactor x-row and x-list gap to use size variables
Add position and line-clap utilities

### DIFF
--- a/packages/x-components/src/components/result/base-result-rating.vue
+++ b/packages/x-components/src/components/result/base-result-rating.vue
@@ -76,7 +76,6 @@
 
 <style scoped lang="scss">
   .x-result-rating {
-    display: inline-block;
     color: inherit;
     text-decoration: none;
   }
--- a/packages/x-components/src/design-system/utilities/line-clamp.scss
+++ b/packages/x-components/src/design-system/utilities/line-clamp.scss
@@ -1,0 +1,8 @@
+@for $i from 2 through 6 {
+  .x-line-clamp--#{$i} {
+    overflow: hidden;
+    display: -webkit-box;
+    -webkit-box-orient: vertical;
+    -webkit-line-clamp: #{$i};
+  }
+}
--- a/packages/x-components/src/design-system/utilities/position.scss
+++ b/packages/x-components/src/design-system/utilities/position.scss
@@ -1,0 +1,15 @@
+.x-static {
+  position: static;
+}
+.x-fixed {
+  position: fixed;
+}
+.x-absolute {
+  position: absolute;
+}
+.x-relative {
+  position: relative;
+}
+.x-sticky {
+  position: sticky;
+}
--- a/packages/x-components/src/views/design-system/design-system-utilities.vue
+++ b/packages/x-components/src/views/design-system/design-system-utilities.vue
@@ -474,6 +474,48 @@
         <div class="x-background--neutral-100 x-margin--left-04">Margin Left 04</div>
       </div>
     </article>
+
+    <article class="positions x-list x-list--wrap x-list--gap-06 x-list--align-start">
+      <h2 class="x-title2">Position</h2>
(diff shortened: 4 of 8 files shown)


packages/x-components/src/design-system/utilities/line-clamp.scss
```diff
@@ -0,0 +1,8 @@
+@for $i from 2 through 6 {
+  .x-line-clamp--#{$i} {
+    overflow: hidden;
+    display: -webkit-box;
+    -webkit-box-orient: vertical;
+    -webkit-line-clamp: #{$i};
```
javieri-empathy, line 6: As this is an utility class, should you add `!important` to some of the properties? I'm specially worried with the `display: -webkit-box`, we tend to override the display property.

packages/x-components/src/design-system/utilities/position.scss
```diff
@@ -0,0 +1,15 @@
+.x-static {
+  position: static;
+}
+.x-fixed {
+  position: fixed;
+}
+.x-absolute {
+  position: absolute;
+}
```
javieri-empathy, line 9: We can change the position property with a class, but what about the top, left, bottom, and right properties? Tailwind adds [CSS classes](https://tailwindcss.com/docs/top-right-bottom-left) for them. Should we also ad  `!important`?

packages/x-components/src/views/design-system/design-system-utilities.vue
```diff
@@ -530,4 +572,14 @@
...
     text-shadow: #888888 0px 0px;
   }
+
+  .positions {
+    .x-absolute,
+    .x-static,
+    .x-fixed,
+    .x-relative,
+    .x-sticky {
+      position: static !important;
```
javieri-empathy, line 582: 🤣

packages/x-components/src/components/result/base-result-rating.vue
```diff
@@ -76,7 +76,6 @@
 
 <style scoped lang="scss">
   .x-result-rating {
```
javieri-empathy, line 78: Not related with this PR, but we have to be careful with the new result rating. In the demo view it is placed as a child another `a` element, which is [invalid HTML](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element).

javieri-empathy (changes requested):

javieri-empathy (approved):

packages/x-components/src/views/design-system/design-system-utilities.vue
```diff
@@ -474,6 +474,48 @@
...
+
+    <article class="positions x-list x-list--wrap x-list--gap-06 x-list--align-start">
+      <h2 class="x-title2">Line Clamp</h2>
+      <div
+        v-for="lines in 5"
+        :key="lines"
+        class="x-background--neutral-95"
+        :class="' x-line-clamp--' + ++lines"
+      >
+        Lorem ipsum dolor sit amet, consectetur adipiscing elit. Maecenas auctor ligula et tristique
```
CachedaCodes, line 495: Other utilities demos have titles or text with the variable values of the token they're representing. Should we do the same here? Like starting the paragraph with `Line clamp of {{ lines }} lines` and then continue with the lorem ipsum. It makes the demo more obvious.

CachedaCodes (changes requested):

packages/x-components/src/views/design-system/design-system-utilities.vue
tajespasarela: added title for that

javieri-empathy (approved):

CachedaCodes (approved):

PR #302, opened 2022-01-13
feat(components): add `buttonClass` prop in `SlidingPanel`

javieri-empathy:
BREAKING CHANGE: `SlidingPanel` buttons do not have the round variant set by default.
EX-5233

Exposes a prop for adding CSS classes to the buttons inside the `SlidingPanel`. The main use case for this is adding variant classes of our design system.

### DIFF
--- a/packages/x-components/src/components/sliding-panel.vue
+++ b/packages/x-components/src/components/sliding-panel.vue
@@ -3,7 +3,8 @@
     <button
       v-if="showButtons"
       @click="scrollLeft"
-      class="x-sliding-panel__button x-sliding-panel__button-left x-button x-button--round"
+      class="x-sliding-panel__button x-sliding-panel__button-left x-button"
+      :class="buttonClass"
       data-test="sliding-panel-left-button"
     >
       <!-- @slot Left button content -->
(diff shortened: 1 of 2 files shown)


packages/x-components/src/components/sliding-panel.vue
```diff
@@ -72,6 +74,14 @@
     @Prop({ default: true })
     public resetOnContentChange!: boolean;
 
+    /**
+     * CSS classes to add to the buttons.
+     *
+     * @public
+     */
+    @Prop()
+    public buttonClass!: string;
```
LuisMartinez15, line 83: Shouldn't this be `?` instead of `!` as it it does not have a `default` value and is not `required`, hence optional?

LuisMartinez15, line 83: Shouldn't this be ? instead of ! as it it does not have a default value and is not required, hence optional?

```suggestion
    public buttonClass?: string;
```

LuisMartinez15 (changes requested):

packages/x-components/src/components/sliding-panel.vue
javieri-empathy: image

tiborux (approved):

LuisMartinez15 (approved):

PR #305, opened 2022-01-14
fix(url): default extra params are now properly restored

javieri-empathy:
EX-5203

Makes the `URLHandler` component use the initial extra params when parsing the URL params. This fixes the issue where selecting an extra param and navigating back (so the default value is selected again), wasn't restoring its value properly.

### DIFF
--- a/packages/x-components/tests/e2e/cucumber/url/url.spec.ts
+++ b/packages/x-components/tests/e2e/cucumber/url/url.spec.ts
@@ -1,4 +1,4 @@
-import { Given, Then, When } from 'cypress-cucumber-preprocessor/steps';
+import { And, Given, Then, When } from 'cypress-cucumber-preprocessor/steps';
 
 // Scenario 1
 Given('a URL with query parameter {string}', (query: string) => {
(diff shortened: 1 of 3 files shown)


packages/x-components/tests/e2e/cucumber/url/url.spec.ts
```diff
@@ -21,3 +24,17 @@ When('navigating back', () => {
 When('clicking result in position {int}', (index: number) => {
   cy.getByDataTest('result-link').eq(index).click();
 });
+
+// Scenario 4
+Then("url doesn't contain parameter {string} with value {string}", (key: string, value: string) => {
```
LuisMartinez15, line 29: Shouldn't this be an `And` instead of a `Then`?

```suggestion
And("url doesn't contain parameter {string} with value {string}", (key: string, value: string) => {
```

packages/x-components/tests/e2e/cucumber/url/url.spec.ts
```diff
@@ -21,3 +24,17 @@ When('navigating back', () => {
...
+  cy.location('search').should('not.contain', `${key}=${value}`);
+});
+
+Then('url contains parameter {string} with value {string}', (key: string, value: string) => {
```
LuisMartinez15, line 33: Same here.

```suggestion
And('url contains parameter {string} with value {string}', (key: string, value: string) => {
```

CachedaCodes (approved):

PR #306, opened 2022-01-17
feat(EX-3539): track PDP add to cart

diegopf:
## Motivation and context

Be able to track that a product was added to the cart from a PDP.

- [x] Open issue. If applicable, link: [EX-3538](https://searchbroker.atlassian.net/browse/EX-3538)

## Type of change

- [X] `Main`
- [ ] Other. Specify:

## How has this been tested?

First of all the tagging component was added to the fake PDP we have so tagging works there.

Scenario `url is configured for tracking`:
1. Click on a result.
2. From the PDP of that result call X.addProductToCart().

Scenario `product id is configured for tracking`:
1. Configure the tagging component on App.vue to use the product id for tracking `<Tagging clicked-result-storage-key="id" />`
2. Click on a result.
3. From the PDP of that result call X.addProductToCart(productId).

## Checklist:


packages/x-components/src/adapter/util.ts
```diff
@@ -15,7 +15,7 @@ export function configureAdapterWithToysrus(
 ): EmpathyAdapterBuilder {
   adapterBuilder
     .addMapper((_, result: Result) => {
-      result.url = `./product_page.html?productId=${result.id}`;
+      result.url = `${location.origin}/product_page.html?productId=${result.id}&test=t e s t`;
```
diegopf, line 18: Currently we are only able to track the PDP add to cart if we are using complete urls. We should think how to handle the case of relative urls.

packages/x-components/src/x-modules/tagging/service/pdp-add-to-cart.service.ts
```diff
@@ -0,0 +1,94 @@
...
+    if (result) {
+      this.sessionStorageService.setItem(clickedResultStorageKeyId, result);
+      this.localStorageService.removeItem(clickedResultStorageKeyId);
+    }
+  }
+
+  trackAddToCart(id?: string | null): void {
+    let clickedResultStorageKeyId: string;
+
+    if (!id) {
```
guillei10, line 73: Considering the productId is coming directly from the snippet or the API (which is fed by the customer), maybe we should make a safer validation? 
Like for example, if we receive an empty string, it will be `!'' === true` but should be treated as if the id is not provided.

packages/x-components/src/App.vue
```diff
@@ -1,5 +1,6 @@
 <template>
   <div class="app">
+    <Tagging :consent="false" />
```
guillei10, line 3: Just curious, why are we importing the `Tagging` in the `App.vue`?

guillei10 (changes requested):
Looks really good.

packages/x-components/src/x-modules/tagging/service/__tests__/pdpd-add-to-cart.service.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+  });
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+    localStorageService.clear();
+    sessionStorageService.clear();
+  });
+
+  it('stores the result with the id as key and ttl', () => {
+    store.commit('x/tagging/setClickedResultStorageKey', 'id');
```
CachedaCodes, line 62: Can you extract these two lines to a function? It's repeated a couple of times in the tests. Maybe making TTL optional too.

packages/x-components/src/x-modules/tagging/service/__tests__/pdpd-add-to-cart.service.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+    store.commit('x/tagging/setClickedResultStorageTTL', 30000);
```
CachedaCodes, line 63: Extract the `30000` to a constant and use that here and in the assertion.

packages/x-components/src/x-modules/tagging/service/__tests__/pdpd-add-to-cart.service.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+
+    service.storeResultClicked(result);
+    expect(localSetItemSpy).toHaveBeenCalledWith(`add-to-cart-${result.id}`, result, 30000);
+  });
+
+  it('stores the result using the url as id and ttl', () => {
+    store.commit('x/tagging/setClickedResultStorageKey', 'url');
+    store.commit('x/tagging/setClickedResultStorageTTL', 30);
```
CachedaCodes, line 71: Same here. Extract the `30` and reuse the constant.

packages/x-components/src/x-modules/tagging/service/__tests__/pdpd-add-to-cart.service.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+    sessionStorageService,
+    store
+  };
+}
+
+describe('testing pdp add to cart', () => {
+  const { service, localStorageService, sessionStorageService, store } =
+    preparePDPAddToCartService();
+
+  const result = createResultStub('Product 001');
```
CachedaCodes, line 39: Build the stub with the url here with:

```suggestion
  const result = createResultStub('Product 001', {url: url});
```

packages/x-components/src/x-modules/tagging/service/__tests__/pdpd-add-to-cart.service.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+    service.storeResultClicked(result);
+    expect(localSetItemSpy).toHaveBeenCalledWith(`add-to-cart-${result.id}`, result, 30000);
+  });
+
+  it('stores the result using the url as id and ttl', () => {
+    store.commit('x/tagging/setClickedResultStorageKey', 'url');
+    store.commit('x/tagging/setClickedResultStorageTTL', 30);
+
+    const encodedURL = `http://localhost:8080/?param=test&param2=&param3=t%20e%20s%20t`;
+    result.url = url;
```
CachedaCodes, line 74: You don't need this if you build the stub with the URL already.

packages/x-components/src/x-modules/tagging/service/__tests__/pdpd-add-to-cart.service.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  const localGetItemSpy = jest.spyOn(localStorageService, 'getItem');
+  const localSetItemSpy = jest.spyOn(localStorageService, 'setItem');
+  const localRemoveItemSpy = jest.spyOn(localStorageService, 'removeItem');
+
+  const sessionGetItemSpy = jest.spyOn(sessionStorageService, 'getItem');
+  const sessionSetItemSpy = jest.spyOn(sessionStorageService, 'setItem');
+  const storeDispatchSpy = jest.spyOn(store, 'dispatch');
+
+  const url = 'http://localhost:8080/?param=test&param2=&param3=t+e+s+t';
```
CachedaCodes, line 49: Create a constant here for the `encondedURL` so it's not polluting the tests.

packages/x-components/src/x-modules/tagging/service/__tests__/pdpd-add-to-cart.service.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+    store.commit('x/tagging/setClickedResultStorageKey', 'url');
+    store.commit('x/tagging/setClickedResultStorageTTL', 30);
+
+    const encodedURL = `http://localhost:8080/?param=test&param2=&param3=t%20e%20s%20t`;
+    result.url = url;
+
+    service.storeResultClicked(result);
+    expect(localSetItemSpy).toHaveBeenCalledWith(`add-to-cart-${encodedURL}`, result, 30);
+
+    result.url = 'http://localhost:8080/?param=test&param2=&param3=t e s t';
```
CachedaCodes, line 79: I don't get why we testing again with a different URL (that would be the same after encoding). Any reason I'm missing?

packages/x-components/src/x-modules/tagging/service/__tests__/pdpd-add-to-cart.service.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  return {
+    service,
+    localStorageService,
+    sessionStorageService,
+    store
+  };
+}
+
+describe('testing pdp add to cart', () => {
```
CachedaCodes, line 35: Would it make sense to test that the TTL is working and the storage is cleaned after the time passes?

CachedaCodes (changes requested):

packages/x-components/src/App.vue
diegopf: yes, that's exactly right @CachedaCodes. First was added for local testing but now it is needed also for the E2E test to work properly.

packages/x-components/src/x-installer/api/api.types.ts
```diff
@@ -10,6 +10,12 @@ import { WireMetadata } from '../../wiring/wiring.types';
  * @public
  */
 export interface XAPI {
+  /**
+   * To track that a product was added to the cart from PDP.
+   *
+   * @param productId - The id of the product added to cart.
+   */
+  addProductToCart(productId?: string): void;
```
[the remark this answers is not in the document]
diegopf: If the tracking is done with the url then you don't need to send the product id because we will use the current url instead.

packages/x-components/src/x-modules/tagging/service/pdp-add-to-cart.service.ts
diegopf: I'm not sure if I follow you, if the customer provides and empty string there, that is a falsy value and it will be treated as no id was provided, which I think is the correct way, isn't it? 😅

packages/x-components/src/x-modules/tagging/components/tagging.vue
```diff
@@ -67,6 +86,25 @@
 
     // eslint-disable-next-line @typescript-eslint/no-empty-function
     render(): void {}
+
+    /**
+     * To emit the event that PDP is loaded  just when the snippet config includes a product id.
```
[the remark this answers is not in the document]
diegopf: I just followed the same approach as we are doing for the [url-handler](https://github.com/empathyco/x/blob/main/packages/x-components/src/x-modules/url/components/url-handler.vue#L104)

packages/x-components/src/x-modules/tagging/service/__tests__/pdpd-add-to-cart.service.spec.ts
diegopf: I want to test here is that does not matter that the url is already encoded or not for whitespaces. IMO makes sense to test both scenarios just to make sure everything is fine.

packages/x-components/src/x-modules/tagging/service/__tests__/pdpd-add-to-cart.service.spec.ts
diegopf: IMO I don't see the reason for that, since that's responsibility of the storage service and we already have tests for that, [see browser storage service tests](https://github.com/empathyco/x/blob/main/packages/storage-service/src/__tests__/browser-storage-service.spec.ts#L45)

CachedaCodes: Okay true. What about the case we discussed in `packages/x-components/src/x-installer/api/api.types.ts` about no `productId` when calling `addToCart`?

CachedaCodes (changes requested):

packages/x-components/src/adapter/util.ts
tajespasarela: can compare only the path of the URL?

packages/x-components/src/x-modules/tagging/service/pdp-add-to-cart.service.ts
```diff
@@ -0,0 +1,112 @@
...
+    this.localStorageService.setItem(storageKey, result, clickedResultStorageTTLMs as number);
+  }
+
+  /**
+   * Checks if the local storage contains a result information for the given id and moves
+   * the result info from the local storage to the session storage.
+   *
+   * @param id - The id of the result to move to the session storage.
+   */
+  moveToSessionStorage(id: string): void {
```
tajespasarela, line 65: add `public` modifier and also in the doc

packages/x-components/src/x-modules/tagging/service/pdp-add-to-cart.service.ts
```diff
@@ -0,0 +1,112 @@
+import { Store } from 'vuex';
```
tajespasarela, line 1: The methods are too big, and do too much things. Refactor them extracting some parts to make it more redable.

tajespasarela (changes requested):

packages/x-components/src/x-modules/tagging/service/pdp-add-to-cart.service.ts
```diff
@@ -0,0 +1,112 @@
...
+    let clickedResultStorageKeyId: string;
+
+    if (id === 'url') {
+      const url = window.location.href.replace(/\s|\+/g, '%20');
+      clickedResultStorageKeyId = `${DefaultPDPAddToCartService.RESULT_CLICKED_ID_KEY}-${url}`;
+    } else {
+      clickedResultStorageKeyId = `${DefaultPDPAddToCartService.RESULT_CLICKED_ID_KEY}-${id}`;
+    }
+
+    const result = this.localStorageService.getItem(clickedResultStorageKeyId);
```
tajespasarela, line 75: the `localStorageService.removeItem` already returns the item or null. So you can do the get and the remove in one single step.

tajespasarela (changes requested):

packages/x-components/src/x-modules/tagging/service/pdp-add-to-cart.service.ts
diegopf: I'm afraid ESLint does not allow it but I've added it to the docs.

guillei10 (commented):
It looks really good to me. Good job. :)

tajespasarela (changes requested):

packages/x-components/src/views/pdp.vue
```diff
@@ -0,0 +1,28 @@
...
+  import { Tagging } from '../x-modules/tagging';
+
+  @Component({
+    components: {
+      Tagging
+    }
+  })
+  export default class Pdp extends Vue {
+    addProductToCart(): void {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-call
```
tajespasarela, line 22: shouldn't be necessary this disable

packages/x-components/src/App.vue
```diff
@@ -1,6 +1,5 @@
 <template>
   <div class="app">
-    <Tagging :consent="false" />
```
tajespasarela: leave this here, it will be used in the results page too

packages/x-components/src/x-modules/tagging/service/pdp-add-to-cart.service.ts
```diff
@@ -99,29 +103,26 @@ export class DefaultPDPAddToCartService implements PDPAddToCartService {
    *
    * @internal
    */
-  protected getStorageId(id: string): string {
+  protected getStorageId(id?: string): string | null {
     if (this.clickedResultStorageKey === 'url') {
-      const url = id.replace(/\s|\+/g, '%20');
+      const url = id ? id.replace(/\s|\+/g, '%20') : window.location.href.replace(/\s|\+/g, '%20');
```
tajespasarela, line 108: if the `url` is configured to be used, then we use it directly:
`const url = window.location.href.replace(/\s|\+/g, '%20');`

tajespasarela (changes requested):

tajespasarela (approved):

tajespasarela (approved):

guillei10 (approved):

PR #311, opened 2022-01-27
feat(design-system): add curated tags

tiborux:
Added curated tags tokens.

EX-5089

### DIFF
--- a/packages/x-components/src/design-system/components/tag/default.scss
+++ b/packages/x-components/src/design-system/components/tag/default.scss
@@ -125,6 +125,19 @@
     --x-size-line-height-filter-default: var(--x-size-line-height-tag-default);
   }
 
+  &.x-tag.x-tag--is-curated {
+    // color
+    background-color: var(--x-color-background-curated-tag-default);
+    border-color: var(--x-color-border-curated-tag-default);
+    color: var(--x-color-text-curated-tag-default);
+
+    // typography
+    line-height: var(--x-size-line-height-curated-tag-default);
+    font-family: var(--x-font-family-curated-tag-default);
+    font-size: var(--x-size-font-curated-tag-default);
+    font-weight: var(--x-number-font-weight-curated-tag-default);
+  }
+
   // selected (the double .x-tag is to override specificity of the scoped component style)
   &.x-tag.x-tag--is-selected {
     // color
@@ -137,6 +150,19 @@
     font-family: var(--x-font-family-tag-default-selected);
     font-size: var(--x-size-font-tag-default-selected);
     font-weight: var(--x-number-font-weight-tag-default-selected);
+
+    &.x-tag--is-curated {
+      // color
+      background-color: var(--x-color-background-curated-tag-default-selected);
+      border-color: var(--x-color-border-curated-tag-default-selected);
+      color: var(--x-color-text-curated-tag-default-selected);
+
+      // typography
+      line-height: var(--x-size-line-height-tag-default-selected);
+      font-family: var(--x-font-family-tag-default-selected);
+      font-size: var(--x-size-font-tag-default-selected);
+      font-weight: var(--x-number-font-weight-tag-default-selected);
+    }
 
     &.x-filter {
       // color
--- a/packages/x-components/src/design-system/components/tag/default.scss
+++ b/packages/x-components/src/design-system/components/tag/default.scss
@@ -127,15 +127,15 @@
 
   &.x-tag.x-tag--is-curated {
     // color
-    background-color: var(--x-color-background-curated-tag-default);
-    border-color: var(--x-color-border-curated-tag-default);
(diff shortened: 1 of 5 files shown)


packages/x-components/src/views/design-system/design-system-tag.vue
```diff
@@ -143,15 +143,38 @@
         <span>text tag extra large name</span>
       </div>
     </article>
+    <article class="x-list x-list--wrap x-list--gap-06 x-list--align-start">
```
tiborux, line 146: Im not sure about this example, I wouldn't add it or mix tags and curated with different styles

diegopf: IMO having them alone is enough.

packages/x-components/src/design-system/components/tag/default.tokens.scss
```diff
@@ -35,4 +42,14 @@
...
+  // typography curated
+  --x-font-family-curated-tag-default: var(--x-font-family-text);
+  --x-size-font-curated-tag-default: var(--x-size-font-text);
+  --x-size-line-height-curated-tag-default: var(--x-size-line-height-text);
+  --x-number-font-weight-curated-tag-default: var(--x-number-font-weight-text);
+  --x-font-family-curated-tag-default-selected: var(--x-font-family-tag-default);
+  --x-size-font-curated-tag-default-selected: var(--x-size-font-tag-default);
+  --x-size-line-height-curated-tag-default-selected: var(--x-size-line-height-tag-default);
+  --x-number-font-weight-curated-tag-default-selected: var(--x-number-font-weight-tag-default);
+
```
diegopf, line 54: remove the empty line

diegopf (changes requested):

packages/x-components/src/design-system/components/tag/default.scss
```diff
@@ -138,6 +151,19 @@
...
+      // color
+      background-color: var(--x-color-background-curated-tag-default-selected);
+      border-color: var(--x-color-border-curated-tag-default-selected);
+      color: var(--x-color-text-curated-tag-default-selected);
+
+      // typography
+      line-height: var(--x-size-line-height-tag-default-selected);
+      font-family: var(--x-font-family-tag-default-selected);
+      font-size: var(--x-size-font-tag-default-selected);
+      font-weight: var(--x-number-font-weight-tag-default-selected);
```
javieri-empathy, line 164: Shouldn't this use the `curated` variant tokens?

packages/x-components/src/views/design-system/design-system-tag.vue
```diff
@@ -143,15 +143,38 @@
...
+      <h2 class="x-title2">Curated</h2>
+      <div class="x-tag x-tag--is-curated">
+        <CuratedCheck />
+        <span>text tag</span>
+      </div>
+      <div class="x-tag x-tag--is-curated">
```
javieri-empathy, line 152: The curated tags work with pill and card variants, but do not work with ghost variant

packages/x-components/src/design-system/components/tag/default.tokens.scss
```diff
@@ -6,6 +6,13 @@
   --x-color-background-tag-default-selected: var(--x-color-base-neutral-95);
   --x-color-border-tag-default-selected: var(--x-color-border-tag-default);
   --x-color-text-tag-default-selected: var(--x-color-text-tag-default);
+  // color curated
+  --x-color-background-curated-tag-default: var(--x-color-base-neutral-100);
```
javieri-empathy, line 10: Is the naming fine? `x-color-background-curated-tag-default` means that we are treating `curated-tag` as a component (and therefore it should have their own styles in its own folder) while I think `curated` is a modifier.

However treating `curated` as a modifier and naming it `x-color-background-tag-default-curated` brings the problem of combining it with the other `selected` modifier.

javieri-empathy (changes requested):

packages/x-components/src/design-system/components/tag/default.tokens.scss
tiborux: the name should be:

`--x-color-background-tag-default-curated-selected`
`--x-color-background-tag-default-curated`

diegopf (dismissed):

packages/x-components/src/views/design-system/design-system-tag.vue
javieri-empathy: Now the ghost variant removes the background in curated RTs, but when selected it does not apply the font weight, which causes them to look identical when selected and when un-selected. You can see so in the second RT of this picture which has the ghost variant applied

image

javieri-empathy (changes requested):

javieri-empathy (approved):

diegopf (approved):

PR #312, opened 2022-01-27
fix: merge extra params with snippet extra params

tiborux:
With this approach instead of using multiple ExtraParams component, only one SnippetExtraParams component will be used. So to test it, just launch the server and see how the extra params are loaded and emitted correctly, you can try to add more values at home.vue in:

` protected initialExtraParams = { store: 'Portugal' };`

to check that you can add more params and they will be combined with snippet one.

EX-5205

### DIFF
--- a/packages/x-components/src/x-modules/extra-params/components/snippet-config-extra-params.vue
+++ b/packages/x-components/src/x-modules/extra-params/components/snippet-config-extra-params.vue
@@ -4,7 +4,7 @@
 
 <script lang="ts">
   import Vue from 'vue';
-  import { Component, Watch, Inject } from 'vue-property-decorator';
+  import { Component, Watch, Inject, Prop } from 'vue-property-decorator';
   import { xComponentMixin } from '../../../components';
   import { Dictionary, forEach } from '../../../utils';
   import { SnippetConfig } from '../../../x-installer';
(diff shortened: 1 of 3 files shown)


diegopf (dismissed):

packages/x-components/src/x-modules/extra-params/components/snippet-config-extra-params.vue
```diff
@@ -65,7 +73,7 @@
       currency,
       ...snippetExtraParams
     }: SnippetConfig): void {
-      forEach(snippetExtraParams, (name, value) => {
+      forEach({ ...snippetExtraParams, ...this.values }, (name, value) => {
```
javieri-empathy, line 76: Where is the test for this new feature? 🤣

packages/x-components/src/x-modules/extra-params/components/snippet-config-extra-params.vue
```diff
@@ -65,7 +73,7 @@
...
         if (this.notAllowedExtraParams.includes(name)) {
```
javieri-empathy, line 77: Just noticed that we are using both destructuring and an array named `notAllowedExtraParams` for discarding parameters, I think we should pick one of those options. We can do it in another PR if you prefer so. 
Also, `scope` parameter is discarded, and its solely purpose is to be sent to the API to group requests depending on the device. We can do it in another PR if you wish so.

javieri-empathy (changes requested):

packages/x-components/src/x-modules/extra-params/components/snippet-config-extra-params.vue
tiborux: the tests are the parents

diegopf (approved):

diegopf:
I see the build is failing but looks like it is due to the tagging flaky tests we currently have 😓

javieri-empathy (approved):

PR #314, opened 2022-01-31
feat(components): Make `BaseResultAddToCart` to use the outter listeners.

tajespasarela:
EX-5323

The point of this PR is to allow to add listeners to the `BaseResultAddToCart`. Without this we can't add a listener to the button in the Customer project:

```html
    <BaseAddToCart @click="doStuff">Add To Cart</BaseAddToCart>
```

### DIFF
--- a/packages/x-components/src/components/result/__tests__/base-result-add-to-cart.spec.ts
+++ b/packages/x-components/src/components/result/__tests__/base-result-add-to-cart.spec.ts
@@ -1,47 +1,86 @@
-import { mount } from '@vue/test-utils';
+import { Result } from '@empathyco/x-types';
+import { mount, Wrapper } from '@vue/test-utils';
 import { getResultsStub } from '../../../__stubs__/results-stubs.factory';
 import { getDataTestSelector, installNewXPlugin } from '../../../__tests__/utils';
+import { XBus } from '../../../plugins/index';
+import { Dictionary } from '../../../utils/index';
 import BaseResultAddToCart from '../base-result-add-to-cart.vue';
 
 describe('testing BaseResultAddToCart component', () => {
   const result = getResultsStub()[0];
-  const [, localVue] = installNewXPlugin();
 
-  it('emits UserClickedResultAddToCart when the user click on the component', () => {
-    const resultAddToCartWrapper = mount(BaseResultAddToCart, {
-      localVue,
-      propsData: { result }
-    });
+  function renderAddToCart({
+    result = { name: 'TEST' } as Result,
+    template = '<BaseResultAddToCart :result="result"/>',
+    methods = {}
+  }: RenderAddToCartOptions): RenderAddToCartApi {
+    const [, localVue] = installNewXPlugin();
+    const wrapper = mount(
+      { template, data: () => ({ result }), methods },
+      {
+        components: { BaseResultAddToCart },
+        localVue
+      }
+    );
+    return {
+      clickAddToCart(): Promise<void> {
+        return wrapper
+          .find(getDataTestSelector('result-add-to-cart'))
+          .trigger('click') as Promise<void>;
+      },
+      on: wrapper.vm.$x.on,
+      addToCartWrapper: wrapper.find(getDataTestSelector('result-add-to-cart'))
+    };
+  }
+
+  it('emits UserClickedResultAddToCart when the user click on the component', async () => {
+    const { clickAddToCart, on } = renderAddToCart({ result });
     const listener = jest.fn();
-    resultAddToCartWrapper.vm.$x.on('UserClickedResultAddToCart').subscribe(listener);
-    resultAddToCartWrapper.trigger('click');
+    on('UserClickedResultAddToCart').subscribe(listener);
+    await clickAddToCart();
     expect(listener).toHaveBeenCalledWith(result);
   });
 
   it('renders the content overriding default slot', () => {
-    const wrapperComponent = {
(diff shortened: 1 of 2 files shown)


packages/x-components/src/components/result/__tests__/base-result-add-to-cart.spec.ts
```diff
@@ -1,47 +1,86 @@
...
+      }
+    });
+    await clickAddToCart();
+
+    expect(listener).toHaveBeenCalled();
   });
 });
+
+interface RenderAddToCartOptions {
+  result: Result;
```
tiborux, line 78: Document both interfaces, in other prs we documented the properties.

packages/x-components/src/components/result/__tests__/base-result-add-to-cart.spec.ts
```diff
@@ -1,47 +1,86 @@
...
-      propsData: { result }
-    });
+  function renderAddToCart({
+    result = { name: 'TEST' } as Result,
+    template = '<BaseResultAddToCart :result="result"/>',
+    methods = {}
+  }: RenderAddToCartOptions): RenderAddToCartApi {
+    const [, localVue] = installNewXPlugin();
+    const wrapper = mount(
+      { template, data: () => ({ result }), methods },
```
tiborux, line 19: Instead of using the data, should we pass the result as a prop by using propsData and props?

tiborux (changes requested):
gj

packages/x-components/src/components/result/__tests__/base-result-add-to-cart.spec.ts
```diff
@@ -1,47 +1,86 @@
...
+    const { clickAddToCart } = renderAddToCart({
+      template: '<BaseResultAddToCart @click="miClick" :result="result"/>',
+      result,
+      methods: {
+        miClick: listener
+      }
+    });
+    await clickAddToCart();
+
+    expect(listener).toHaveBeenCalled();
```
javieri-empathy, line 73: What is the expected payload for this click callback? A result? A `MouseEvent` (which it is right now)?

```suggestion
    expect(listener).toHaveBeenCalledTimes(1);
    expect(listener).toHaveBeenCalledWith(expect.any(MouseEvent));
```

packages/x-components/src/components/result/__tests__/base-result-add-to-cart.spec.ts
```diff
@@ -1,47 +1,86 @@
...
     ).toBeDefined();
-    expect(customResultAddToCartWrapper.text()).toEqual('Add to cart');
+    expect(addToCartWrapper.text()).toEqual('Add to cart');
+  });
+
+  it('uses the listeners passed', async () => {
+    const listener = jest.fn();
+    const { clickAddToCart } = renderAddToCart({
+      template: '<BaseResultAddToCart @click="miClick" :result="result"/>',
+      result,
```
javieri-empathy, line 66: You are doing nothing with the result in this test and the above one so you can remove it and use the default one.

packages/x-components/src/components/result/__tests__/base-result-add-to-cart.spec.ts
```diff
@@ -1,47 +1,86 @@
...
   const result = getResultsStub()[0];
-  const [, localVue] = installNewXPlugin();
 
-  it('emits UserClickedResultAddToCart when the user click on the component', () => {
-    const resultAddToCartWrapper = mount(BaseResultAddToCart, {
-      localVue,
-      propsData: { result }
-    });
+  function renderAddToCart({
+    result = { name: 'TEST' } as Result,
```
javieri-empathy, line 13: You are always rendering the same result. Why not set its default value here and then return it in the API object? That way you can also remove the global result, but you still have the rendered result available for when you need to access it.

javieri-empathy (changes requested):

packages/x-components/src/components/result/__tests__/base-result-add-to-cart.spec.ts
```diff
@@ -1,47 +1,86 @@
...
+    };
+  }
+
+  it('emits UserClickedResultAddToCart when the user click on the component', async () => {
+    const { clickAddToCart, on } = renderAddToCart({ result });
     const listener = jest.fn();
-    resultAddToCartWrapper.vm.$x.on('UserClickedResultAddToCart').subscribe(listener);
-    resultAddToCartWrapper.trigger('click');
+    on('UserClickedResultAddToCart').subscribe(listener);
+    await clickAddToCart();
```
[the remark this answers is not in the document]
tajespasarela: 👍🏻

packages/x-components/src/components/result/__tests__/base-result-add-to-cart.spec.ts
tajespasarela: changed 👍🏻

packages/x-components/src/components/result/__tests__/base-result-add-to-cart.spec.ts
tajespasarela: 👍🏻

packages/x-components/src/components/result/__tests__/base-result-add-to-cart.spec.ts
tajespasarela: done

tiborux (approved):

javieri-empathy (approved):

PR #315, opened 2022-01-31
feat(components): Add `payload` and `metadata` to the `SnippetCallbackExecuted` event payload.

tajespasarela:
EX-5324

Until now, the event  `SnippetCallbackExecuted`, emitted after a callback execution, only passed the event name and the callback return. But sometimes it's necessary also to know the original payload and metadata.

So this PR is to fix that.

### DIFF
--- a/packages/x-components/src/wiring/events.types.ts
+++ b/packages/x-components/src/wiring/events.types.ts
@@ -1,4 +1,5 @@
 import { Result, Suggestion } from '@empathyco/x-types';
+import { ExtractPayload } from '../store/index';
 import { ArrowKey, PropsWithType } from '../utils';
 import { DeviceXEvents } from '../x-modules/device';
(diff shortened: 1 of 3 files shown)


packages/x-components/src/wiring/events.types.ts
```diff
@@ -174,7 +176,12 @@ export interface XEventsTypes
    * A callback from the snippet has been executed.
    * * Payload: An object containing the event that executed the callback and the callback result.
```
tiborux, line 177: update the doc

```suggestion
   * * Payload: An object containing the event that executed the callback, the callback result the payload and the metadata.
```

tiborux (changes requested):

javieri-empathy (approved):

tiborux (approved):

PR #317, opened 2022-02-04
feat(design-system): change curated tag toke values to reuse default …

javieri-empathy:
This just changes the value of the curated tag tokens to use the default tag ones, as they are identical. This should ease customization in customer projects.

### DIFF
--- a/packages/x-components/src/design-system/components/tag/default.tokens.scss
+++ b/packages/x-components/src/design-system/components/tag/default.tokens.scss
@@ -7,12 +7,12 @@
   --x-color-border-tag-default-selected: var(--x-color-border-tag-default);
   --x-color-text-tag-default-selected: var(--x-color-text-tag-default);
   // color curated
-  --x-color-background-tag-default-curated: var(--x-color-base-neutral-100);
-  --x-color-border-tag-default-curated: var(--x-color-text-tag-default);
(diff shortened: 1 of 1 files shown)


packages/x-components/src/design-system/components/tag/default.tokens.scss
```diff
@@ -7,12 +7,12 @@
   --x-color-border-tag-default-selected: var(--x-color-border-tag-default);
   --x-color-text-tag-default-selected: var(--x-color-text-tag-default);
   // color curated
-  --x-color-background-tag-default-curated: var(--x-color-base-neutral-100);
-  --x-color-border-tag-default-curated: var(--x-color-text-tag-default);
-  --x-color-text-tag-default-curated: var(--x-color-text-default);
-  --x-color-background-tag-default-curated-selected: var(--x-color-base-neutral-95);
-  --x-color-border-tag-default-curated-selected: var(--x-color-border-tag-default);
-  --x-color-text-tag-default-curated-selected: var(--x-color-text-tag-default);
+  --x-color-background-tag-default-curated: var(--x-color-background-tag-default);
```
tajespasarela, line 10: please do de same with the typography tokens

tajespasarela (changes requested):

tajespasarela (approved):

PR #320, opened 2022-02-07
feat(components): Created abstract factory for animations.

tajespasarela:
BREAKING-CHANGE: Removed `CollapseFromTop` Animation. Use `animateScale()` instead.
BREAKING-CHANGE: Removed `TranslateFromLeft` and `TranslateFromRight` components. Use `animateTranslate('left')` and `animateTranslate('right') instead.

EX-5355


packages/x-components/src/components/animations/animate-scale/animate-scale.style.scss
```diff
@@ -0,0 +1,100 @@
+.x-animate-scale {
```
javieri-empathy, line 1: I'm worried about the generated size of this file, 22kb of css, which end up being 3kb gzipped just for a simple scale animation.

In theory you should only need 18 frames for the animation to look smooth, 60Hz of average screens multiplied by 0.3, or 36 frames if you still want to support super smooth screens. I tested with 36 and it reduces the css to 10kb and 1kb ggzipped, and still looks smooth.

packages/x-components/src/components/animations/animation.abstract-factory.ts
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Abstract Factory to create animations Factory. The returned animation factory uses the
+ * `animationName` parameter to create a Transition Component with that name.
+ *
+ * @param animationName - The name to use in the Transition Component of animation.
+ * @returns The animation factory configured.
+ *
+ * @internal
+ */
+export function animationAbstractFactory(
```
javieri-empathy, line 12: I know some times might be good to add in to the name of what you are doing the pattern, but wouldn't it be more natural to just call it  something more explicit and easier like `createDirectionalAnimationFactory`. We even have this in our naming guide:

>
> ## Methods and functions
> 
> - Methods and functions must start with a verb.

javieri-empathy (changes requested):

packages/x-components/src/components/animations/animation.abstract-factory.ts
```diff
@@ -0,0 +1,33 @@
...
+  animationName: string
+): (animationOrigin?: AnimationOrigin) => VueConstructor<Vue> {
+  return (animationOrigin: AnimationOrigin = 'top'): VueConstructor<Vue> => {
+    return Vue.extend({
+      data() {
+        return { animationOrigin };
```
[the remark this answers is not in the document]
LuisMartinez15: The only use case I can think of is if a user changes from a `ltr` language to a `rtl` one, with an animation component already loaded, and without the page triggering a reload. However, the chance for this to happen is minimal.

packages/x-components/src/components/animations/animate-translate/animate-translate.style.scss
```diff
@@ -0,0 +1,24 @@
+.x-animate-translate {
+  $p: &;
+
+  &--enter-active,
+  &--leave-active {
+    transition: transform var(--x-duration-animation, 0.3s) ease-out;
+  }
+  $p: &;
```
LuisMartinez15, line 8: Isn't this already done in `line 2`? You can remove this one:

```suggestion

```

packages/x-components/src/components/animations/animate-scale/animate-scale.style.scss
LuisMartinez15: I agree. We are creating a lot of unnecessary frames here.

packages/x-components/src/components/animations/__tests__/animations-abstract-factory.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+        },
+        props: {
+          open: { default: false }
+        }
+      },
+      {}
+    );
+
+    return {
+      toggleOpen(): void | Promise<void> {
```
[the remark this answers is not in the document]
LuisMartinez15: Agree.

packages/x-components/src/views/home/Home.vue
```diff
@@ -519,17 +489,9 @@
   import SlicedFilters from '../../x-modules/facets/components/lists/sliced-filters.vue';
   import SortedFilters from '../../x-modules/facets/components/lists/sorted-filters.vue';
   // eslint-disable-next-line max-len
-  import ClearHistoryQueries from '../../x-modules/history-queries/components/clear-history-queries.vue';
-  import HistoryQueries from '../../x-modules/history-queries/components/history-queries.vue';
-  // eslint-disable-next-line max-len
-  import IdentifierResult from '../../x-modules/identifier-results/components/identifier-result.vue';
-  // eslint-disable-next-line max-len
```
LuisMartinez15, line 526: You can remove this line now too.

packages/x-components/src/components/animations/__tests__/animations-abstract-factory.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+      },
+      getContentWrapper(): Wrapper<Vue> {
+        return wrapper.find(getDataTestSelector('animation-content'));
+      },
+      wrapper
+    };
+  }
+
+  // eslint-disable-next-line max-len
+  it('renders the animation component created by abstract factory.', async () => {
```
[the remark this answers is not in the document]
LuisMartinez15: I agree. Instead of checking if the wrapper exists, I think we should be checking that a `transition` node is being rendered with a prop `name` containing ``x-${animationName}--${animationOrigin} x-${animationName}-``

LuisMartinez15 (changes requested):
Good job!

packages/x-components/src/components/animations/create-directional-animation-factory.ts
```diff
@@ -0,0 +1,32 @@
...
+ * @returns The animation factory configured.
+ *
+ * @internal
+ */
+export function createDirectionalAnimationFactory(
+  animationName: string
+): (animationOrigin?: AnimationOrigin) => VueConstructor {
+  return (animationOrigin = 'top') => {
+    return Vue.extend({
+      name: `Transition${capitalize(animationName)}${capitalize(animationOrigin)}`,
```
javieri-empathy, line 18: If you want to leave the name in kebab case it is fine, now its kinda a monster mixing kebab, pascal, and camelcase🤣 but fortunately Vue devtools are able to transform it to pascal case properly.

```suggestion
      name: `${animationName}-${animationOrigin}`,
```

image

javieri-empathy (changes requested):

packages/x-components/src/components/animations/create-directional-animation-factory.ts
LuisMartinez15: You're right. Fortunately, Vue devtools already transforms from `kebab-case` to `PascalCase`.

Note: if you want to keep the `Transition` prefix, don't commit Javi's suggestion directly and do the change manually 😅 .

LuisMartinez15 (changes requested):

packages/x-components/src/components/animations/create-directional-animation-factory.ts
tajespasarela: done

javieri-empathy (approved):

PR #325, opened 2022-02-09
feat(search-adapter): detect adblockers and use fetch instead of beacon for tracking when present

CachedaCodes:
[EX-5372](https://searchbroker.atlassian.net/browse/EX-5371)

## Motivation and context
Tracking requests are being blocked by ad blocked (specifically uBlock) because they're of type `ping`. We're using `sendBeacon` for these requests and due to its nature, it's not trivial to know when the request is being blocked.
We opted for detecting when an adblock is present by sending a regular request with `fetch` to an URL that will be blocked. If an ad blocked is present then instead of sending the beacon we use a regular fetch.

## Type of change

- [X] `Main`
- [ ] Other. Specify:

## How has this been tested?
This change is aimed to solve issues in clients, so the test done was:
* Take a client's project (`x-massimoduti` for example)
* Add the local `x-adapter` as a dependency
* Override the adapter configuration so it uses the `x-adapter` `BeaconTrackingRequestor`
````javascript
configureAdapter
...
.configureContainer((container) => {
  container.rebind(DEPENDENCIES.Requestors.track).to(BeaconTrackingRequestor)
})
````
* Check that the request is sent even with uBlock on and the Beacon failing

But it can be tested any other way that triggers the `BeaconTrackingRequestor`.

### DIFF
--- a/packages/search-adapter/src/empathy/requestors/beacon-tracking.requestor.ts
+++ b/packages/search-adapter/src/empathy/requestors/beacon-tracking.requestor.ts
@@ -2,6 +2,11 @@ import { injectable } from 'inversify';
 import { TrackingRequest } from '../../types';
 import { Requestor } from '../empathy-adapter.types';
 
+let isAdblockPresent = false;
+detectAdblock()
+  .then(() => (isAdblockPresent = false))
+  .catch(() => (isAdblockPresent = true));
+
 /**
  * TODO https://searchbroker.atlassian.net/browse/EX-2163
  *
@@ -9,20 +14,35 @@ import { Requestor } from '../empathy-adapter.types';
  */
 @injectable()
 export class BeaconTrackingRequestor implements Requestor<TrackingRequest, void> {
-
   request(trackingRequest: TrackingRequest): Promise<void> {
     const url = this.buildUrl(trackingRequest);
 
-    if (navigator.sendBeacon(url)) {
+    if (!isAdblockPresent && navigator.sendBeacon(url)) {
       return Promise.resolve();
     } else {
+      if (isAdblockPresent) {
+        fetch(url);
+      }
       return Promise.reject('Beacon not queued');
     }
   }
 
   private buildUrl(trackingRequest: TrackingRequest): string {
     const url = new URL(trackingRequest.url);
-    Object.entries(trackingRequest.params).forEach(([key, value]) => url.searchParams.append(key, value));
+    Object.entries(trackingRequest.params).forEach(([key, value]) =>
+      url.searchParams.append(key, value)
+    );
     return url.href;
   }
+}
+
+function detectAdblock(): Promise<Response> {
+  const flaggedURL = 'https://google.com/pagead/js/adsbygoogle.js';
+
+  return fetch(
+    new Request(flaggedURL, {
+      method: 'HEAD',
+      mode: 'no-cors'
+    })
+  );
 }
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/search-adapter/src/empathy/requestors/beacon-tracking.requestor.ts [hunk 1]
```diff
@@ -2,27 +2,51 @@ import { injectable } from 'inversify';
 import { TrackingRequest } from '../../types';
 import { Requestor } from '../empathy-adapter.types';
 
+let isAdblockPresent = false;
+detectAdblock()
+  .then(() => (isAdblockPresent = false))
+  .catch(() => (isAdblockPresent = true));
+
 /**
  * TODO https://searchbroker.atlassian.net/browse/EX-2163
```
CachedaCodes, line 11: This links nowhere, what was the TODO about?

packages/search-adapter/src/empathy/requestors/beacon-tracking.requestor.ts
```diff
@@ -9,20 +9,47 @@ import { Requestor } from '../empathy-adapter.types';
  */
 @injectable()
 export class BeaconTrackingRequestor implements Requestor<TrackingRequest, void> {
+  constructor() {
+    this.detectAdblock()
+      .then(() => (this.isAdblockPresent = false))
+      .catch(() => (this.isAdblockPresent = true));
+  }
+
+  private isAdblockPresent = false;
```
tiborux, line 18: move this at the beginning of the clase

packages/search-adapter/src/empathy/requestors/beacon-tracking.requestor.ts
```diff
@@ -9,20 +9,47 @@ import { Requestor } from '../empathy-adapter.types';
...
 
   request(trackingRequest: TrackingRequest): Promise<void> {
     const url = this.buildUrl(trackingRequest);
 
-    if (navigator.sendBeacon(url)) {
+    if (!this.isAdblockPresent && navigator.sendBeacon(url)) {
       return Promise.resolve();
     } else {
+      if (this.isAdblockPresent) {
+        fetch(url);
```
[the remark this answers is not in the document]
tiborux: return the promise here, to know if the fetch failed or not

packages/search-adapter/src/empathy/requestors/beacon-tracking.requestor.ts
javieri-empathy: This was about documenting the entire adapter :rip:

packages/search-adapter/src/empathy/requestors/beacon-tracking.requestor.ts
```diff
@@ -9,20 +9,47 @@ import { Requestor } from '../empathy-adapter.types';
  */
 @injectable()
 export class BeaconTrackingRequestor implements Requestor<TrackingRequest, void> {
+  constructor() {
+    this.detectAdblock()
+      .then(() => (this.isAdblockPresent = false))
+      .catch(() => (this.isAdblockPresent = true));
```
javieri-empathy, line 15: Use fat arrow functions, you don't want to return the boolean, you just want to assign it:
```suggestion
      .then(() => {
        this.isAdblockPresent = false;
      })
      .catch(() => {
        this.isAdblockPresent = true;
      });
```

javieri-empathy (changes requested):
Tested in:
- Chrome with ublock
- Safari desktop with adguard & adblock
- Safari iOS with adguard
- Firefox with ublock

And track requests are working

javieri-empathy (approved):

PR #326, opened 2022-02-09
ci: add `concurrency` option to `build` `workflow`

LuisMartinez15:
EX-5357

Avoid overlapping multiple executions of the `build` `workflow`.

## Testing

We've tested this option in a dummy repository and seems to be working as expected.


javieri-empathy (dismissed):

javieri-empathy:
What would happen in a pull request which is build 2 times using the `build` workflow? Are these builds run sequentially or in parallel?

diegopf (approved):

PR #327, opened 2022-02-14
chore: create empty `x-adapter` (`x-adapter-next`) and `x-adapter-platform` packages

LuisMartinez15:
EX-5310

## Motivation and context
Create new empty `x-adapter` (`@empathyco/x-adapter-next` until we can get rid of the current adapter) and `x-adapter-platform` packages.

### DIFF
--- a/packages/x-adapter-platform/README.md
+++ b/packages/x-adapter-platform/README.md
@@ -1,0 +1,8 @@
+# x-adapter-platform
+
+`x-adapter-platform` is a library to ease the communication with `empathy.co platform search API's`.
+
+### Contributing
+
+To start contributing to the project, please take a look to our
+[Contributing Guide](../../.github/CONTRIBUTING.md).
--- a/packages/x-adapter-platform/package.json
+++ b/packages/x-adapter-platform/package.json
@@ -1,0 +1,25 @@
+{
+  "name": "@empathyco/x-adapter-platform",
+  "version": "0.1.0",
+  "description": "A search client for empathy.co's platform search API",
+  "author": "Empathy Systems Corporation S.L.",
+  "license": "Apache-2.0",
+  "homepage": "https://github.com/empathyco/x/tree/main/packages/x-adapter#readme",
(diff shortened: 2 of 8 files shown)


tajespasarela:
should we configure the eslint and prettier with our plugin package?

LuisMartinez15:
> should we configure the `eslint` and `prettier` with our `plugin` package?

I was thinking about including those here, but didn't know if we wanted to include anything beside the package _skeleton_. I'll gladly add them if you guys think we should do it here.

tiborux:
> > should we configure the `eslint` and `prettier` with our `plugin` package?
> 
> I was thinking about including those here, but didn't know if we wanted to include anything beside the package _skeleton_. I'll gladly add them if you guys think we should do it here.
> 
> Edit: I've been checking our current configuration for `eslint` & `prettier`, and I've seen that both are installed at the `monorepo` level, instead of per `package`. In addition, the `eslint's config`, when using `typescript`, expects a `tsconfig.eslint.json` file to be configured (and this one depends on the `tsconfig.json`). So, should I configure everything, including adding `typescript` and its dependencies or should I leave it like it is right now and leave this configuration for when we start coding? @reviewers

Maybe its a good idea to have the package ready to code. I mean, if we add this configurations we will be able to start coding in the next task that it is going to be added in the new sprint, anyways, we can create a new task just for this.

tiborux (approved):

tiborux (approved):

LuisMartinez15:
As requested, I've added a bare bone config for `eslint` to work inside both packages. In order to verify it is working, you can create a dummy `.ts` file under `src/` folder with content that `eslint` should warn you about. 

Note: I didn't create a specific `.eslintrc.js` file for each package as the global one in the `monorepo` should be enough for now. In case we need to customise rules, extend from another set of rules from our own plugin, or even add new ones, we can create a custom file inside the desired package.

CachedaCodes (approved):

PR #331, opened 2022-02-15
feat(scroll): add scroll data state to the alias API

tiborux:
Go to any component and inside a function write the following code:

`this.$x.scroll['main-scroll'].WHATEVER` check the different values that it is offering. You can print these values to check that everything is working as expected when you use the scroll.

Or you can add it in the scroll.vue component, inside the emitScroll function:

` protected emitScroll(position: number): void {
     console.log(this.$x.scroll['main-scroll'].hasReachedStart);
      this.$x.emit('UserScrolled', position, this.createEventMetadata());
 }`

EX-5249


LuisMartinez15 (approved):
Good job!

LGTM

CachedaCodes (approved):

PR #332, opened 2022-02-15
feat(design-system): add curated variant to the suggestion

CachedaCodes:
[EX-5290](https://searchbroker.atlassian.net/browse/EX-5290)

## Motivation and context
The curated variant had to be added to the next queries in the design system. To do this be opted to make this general to all suggestions. And in a similar fashion to what was done in [EX-5088](https://searchbroker.atlassian.net/browse/EX-5088).

## Type of change

- [X] `Main`
- [ ] Other. Specify:

## How has this been tested?
New sub-sections were added to the Suggestions section in the Design System layout. Any change in the tokens for the curated suggestions will be reflected there and only affect curated suggestions.

### DIFF
--- a/packages/x-components/src/components/suggestions/base-suggestion.vue
+++ b/packages/x-components/src/components/suggestions/base-suggestion.vue
@@ -68,6 +68,14 @@
     protected suggestionSelectedEvents!: Partial<XEventsTypes>;
 
     /**
+     * Indicates if the curated suggestion should be highlighted.
+     *
+     * @public
+     */
+    @Prop({ default: false, type: Boolean })
+    protected highlightCurated!: boolean;
+
+    /**
      * The event handler that will be triggered when clicking on a suggestion.
      *
      * @remarks
@@ -112,6 +120,17 @@
     }
 
     /**
+     * Checks if the suggestion is curated and if it should be highlighted.
+     *
+     * @returns True if the suggestion is curated and should be highlighted.
+     *
+     * @internal
+     */
+    protected get shouldHighlightCurated(): boolean {
+      return this.highlightCurated && (this.suggestion.isCurated ?? false);
+    }
+
+    /**
      * Generates css classes dynamically.
      *
      * @remarks
@@ -122,7 +141,8 @@
      */
     protected get dynamicCSSClasses(): VueCSSClasses {
       return {
-        'x-suggestion--matching': this.hasMatchingQuery
+        'x-suggestion--matching': this.hasMatchingQuery,
+        'x-suggestion--is-curated': this.shouldHighlightCurated,
       };
     }
 
--- a/packages/x-components/src/components/suggestions/base-suggestion.vue
+++ b/packages/x-components/src/components/suggestions/base-suggestion.vue
@@ -127,7 +127,7 @@
      * @internal
      */
     protected get shouldHighlightCurated(): boolean {
-      return this.highlightCurated && (this.suggestion.isCurated ?? false);
+      return this.highlightCurated && !!this.suggestion.isCurated;
     }
 
     /**
--- a/packages/x-components/src/design-system/components/suggestion/default.scss
+++ b/packages/x-components/src/design-system/components/suggestion/default.scss
@@ -74,4 +74,42 @@
   > .x-identifier-result {
     flex: none;
   }
+
+  &.x-suggestion--is-curated {
+    // color
+    background-color: var(--x-color-background-suggestion-default-curated);
+    color: var(--x-color-text-suggestion-default-curated);
+    border-color: var(--x-color-border-suggestion-default-curated);
+
+    // typography
+    font-family: var(--x-font-family-suggestion-default-curated);
+    font-size: var(--x-size-font-suggestion-default-curated);
+    line-height: var(--x-size-line-height-suggestion-default-curated);
+    font-weight: var(--x-number-font-weight-suggestion-default-curated);
+
+    .x-suggestion__matching-part,
+    .x-identifier-result__matching-part {
+      // typography
+      font-family: var(--x-font-family-suggestion-matching-part-default-curated);
+      font-size: var(--x-size-font-suggestion-matching-part-default-curated);
+      line-height: var(--x-size-line-height-suggestion-matching-part-default-curated);
+      font-weight: var(--x-number-font-weight-suggestion-matching-part-default-curated);
+
+      // color
+      color: var(--x-color-text-suggestion-matching-part-default-curated);
+    }
+  }
+
+  &.x-suggestion--is-curated.x-suggestion--matching {
+    // typography
+    --x-font-family-suggestion-default-curated: var(--x-font-family-suggestion-default-matching-curated);
+    --x-size-font-suggestion-default-curated: var(--x-size-font-suggestion-default-matching-curated);
+    --x-size-line-height-suggestion-default-curated: var(--x-size-line-height-suggestion-default-matching-curated);
+    --x-number-font-weight-suggestion-default-curated: var(
+                    --x-number-font-weight-suggestion-default-matching-curated
+    );
+
+    // color
+    --x-color-text-suggestion-default-curated: var(--x-color-text-suggestion-default-matching-curated);
+  }
 }
--- a/packages/x-components/src/design-system/components/suggestion/default.scss
+++ b/packages/x-components/src/design-system/components/suggestion/default.scss
@@ -98,18 +98,18 @@
       // color
       color: var(--x-color-text-suggestion-matching-part-default-curated);
     }
-  }
 
-  &.x-suggestion--is-curated.x-suggestion--matching {
-    // typography
-    --x-font-family-suggestion-default-curated: var(--x-font-family-suggestion-default-matching-curated);
-    --x-size-font-suggestion-default-curated: var(--x-size-font-suggestion-default-matching-curated);
-    --x-size-line-height-suggestion-default-curated: var(--x-size-line-height-suggestion-default-matching-curated);
-    --x-number-font-weight-suggestion-default-curated: var(
-                    --x-number-font-weight-suggestion-default-matching-curated
-    );
+    &.x-suggestion--matching {
+      // typography
+      --x-font-family-suggestion-default-curated: var(--x-font-family-suggestion-default-matching-curated);
+      --x-size-font-suggestion-default-curated: var(--x-size-font-suggestion-default-matching-curated);
+      --x-size-line-height-suggestion-default-curated: var(--x-size-line-height-suggestion-default-matching-curated);
+      --x-number-font-weight-suggestion-default-curated: var(
+                      --x-number-font-weight-suggestion-default-matching-curated
+      );
 
-    // color
-    --x-color-text-suggestion-default-curated: var(--x-color-text-suggestion-default-matching-curated);
+      // color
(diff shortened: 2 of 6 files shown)


CachedaCodes:
Given that this moves the curated highlighting logic to the `Suggestion.vue` component I think that the code in the `NextQuery.vue` component is no longer necessary (I'm talking about the `shouldHighlightCurated` method). This would also affect the tests of the component if it's removed, so those tests would have to be done in the Suggestions test file.

If we have to make that change I can do it in this PR or open a new task (or subtask of the parent of this PR).

packages/x-components/src/design-system/components/suggestion/default.tokens.scss
```diff
@@ -8,6 +8,12 @@
   --x-color-text-suggestion-default-matching: var(--x-color-text-suggestion-default);
   --x-color-background-suggestion-default: transparent;
   --x-color-border-suggestion-default: var(--x-color-text-suggestion-default);
+  // color curated
```
tiborux, line 11: add white space

packages/x-components/src/components/suggestions/base-suggestion.vue
```diff
@@ -111,6 +119,17 @@
...
 
+    /**
+     * Checks if the suggestion is curated and if it should be highlighted.
+     *
+     * @returns True if the suggestion is curated and should be highlighted.
+     *
+     * @internal
+     */
+    protected get shouldHighlightCurated(): boolean {
+      return this.highlightCurated && (this.suggestion.isCurated ?? false);
```
tiborux, line 130: I know that this is copied from the related tags but I think that we can change it in both to:
```suggestion
      return this.highlightCurated && !!this.suggestion.isCurated;
```

packages/x-components/src/design-system/components/suggestion/default.scss
```diff
@@ -74,4 +74,42 @@
...
+    --x-font-family-suggestion-default-curated: var(--x-font-family-suggestion-default-matching-curated);
+    --x-size-font-suggestion-default-curated: var(--x-size-font-suggestion-default-matching-curated);
+    --x-size-line-height-suggestion-default-curated: var(--x-size-line-height-suggestion-default-matching-curated);
+    --x-number-font-weight-suggestion-default-curated: var(
+                    --x-number-font-weight-suggestion-default-matching-curated
+    );
+
+    // color
+    --x-color-text-suggestion-default-curated: var(--x-color-text-suggestion-default-matching-curated);
+  }
```
tiborux, line 114:
```suggestion
  &.x-suggestion.x-suggestion--is-curated {
    // color
    background-color: var(--x-color-background-suggestion-default-curated);
    color: var(--x-color-text-suggestion-default-curated);
    border-color: var(--x-color-border-suggestion-default-curated);

    // typography
    font-family: var(--x-font-family-suggestion-default-curated);
    font-size: var(--x-size-font-suggestion-default-curated);
    line-height: var(--x-size-line-height-suggestion-default-curated);
    font-weight: var(--x-number-font-weight-suggestion-default-curated);

    .x-suggestion__matching-part,
    .x-identifier-result__matching-part {
      // typography
      font-family: var(--x-font-family-suggestion-matching-part-default-curated);
      font-size: var(--x-size-font-suggestion-matching-part-default-curated);
      line-height: var(--x-size-line-height-suggestion-matching-part-default-curated);
      font-weight: var(--x-number-font-weight-suggestion-matching-part-default-curated);

      // color
      color: var(--x-color-text-suggestion-matching-part-default-curated);
    }
      &.x-suggestion--matching {
    // typography
    --x-font-family-suggestion-default-curated: var(--x-font-family-suggestion-default-matching-curated);
    --x-size-font-suggestion-default-curated: var(--x-size-font-suggestion-default-matching-curated);
    --x-size-line-height-suggestion-default-curated: var(--x-size-line-height-suggestion-default-matching-curated);
    --x-number-font-weight-suggestion-default-curated: var(
                    --x-number-font-weight-suggestion-default-matching-curated
    );

    // color
    --x-color-text-suggestion-default-curated: var(--x-color-text-suggestion-default-matching-curated);
  }
  }
```

tiborux: Basically, add .x-suggestion in the first line to be like this:

&.x-suggestions.x-suggestion--is-curated 

And move the   &.x-suggestion--is-curated.x-suggestion--matching  inside of the previous  classes like this:

&.x-suggestions.x-suggestion--is-curated {
     &.x-suggestion--matching {
}
}

Instead of having it as a root element

packages/x-components/src/design-system/components/suggestion/default.tokens.scss
LuisMartinez15: Yep, add a blank line here.

packages/x-components/src/design-system/components/suggestion/default.tokens.scss
```diff
@@ -8,6 +8,12 @@
...
+  --x-color-text-suggestion-default-curated: var(--x-color-text-default);
+  --x-color-text-suggestion-matching-part-default-curated: var(--x-color-base-neutral-35);
+  --x-color-text-suggestion-default-matching-curated: var(--x-color-text-suggestion-default);
+  --x-color-background-suggestion-default-curated: transparent;
+  --x-color-border-suggestion-default-curated: var(--x-color-text-suggestion-default);
```
LuisMartinez15, line 16: These tokens should be referencing their default counterparts:

```suggestion
  --x-color-text-suggestion-default-curated: var(--x-color-text-suggestion-default);
  --x-color-text-suggestion-matching-part-default-curated: var(
    --x-color-text-suggestion-matching-part-default
  );
  --x-color-text-suggestion-default-matching-curated: var(
    --x-color-text-suggestion-default-matching
  );
  --x-color-background-suggestion-default-curated: var(--x-color-background-suggestion-default);
  --x-color-border-suggestion-default-curated: var(--x-color-border-suggestion-default);
```

packages/x-components/src/design-system/components/suggestion/default.tokens.scss
```diff
@@ -48,4 +54,23 @@
...
+  );
+  --x-number-font-weight-suggestion-matching-part-default-curated: var(
+                  --x-number-font-weight-suggestion-default
+  );
+  --x-font-family-suggestion-default-matching-curated: var(--x-font-family-suggestion-default);
+  --x-size-font-suggestion-default-matching-curated: var(--x-size-font-suggestion-default);
+  --x-size-line-height-suggestion-default-matching-curated: var(--x-size-line-height-suggestion-default);
+  --x-number-font-weight-suggestion-default-matching-curated: var(
+                  --x-number-font-weight-suggestion-default
+  );
```
LuisMartinez15, line 75: Same here:

```suggestion
  --x-font-family-suggestion-default-curated: var(--x-font-family-suggestion-default);
  --x-size-font-suggestion-default-curated: var(--x-size-font-suggestion-default);
  --x-size-line-height-suggestion-default-curated: var(--x-size-line-height-suggestion-default);
  --x-number-font-weight-suggestion-default-curated: var(--x-number-font-weight-suggestion-default);
  --x-font-family-suggestion-matching-part-default-curated: var(--x-font-family-suggestion-matching-part-default);
  --x-size-font-suggestion-matching-part-default-curated: var(--x-size-font-suggestion-matching-part-default);
  --x-size-line-height-suggestion-matching-part-default-curated: var(
    --x-size-line-height-suggestion-matching-part-default
  );
  --x-number-font-weight-suggestion-matching-part-default-curated: var(
    --x-number-font-weight-suggestion-matching-part-default
  );
  --x-font-family-suggestion-default-matching-curated: var(--x-font-family-suggestion-default-matching);
  --x-size-font-suggestion-default-matching-curated: var(--x-size-font-suggestion-default-matching);
  --x-size-line-height-suggestion-default-matching-curated: var(
    --x-size-line-height-suggestion-default-matching
  );
  --x-number-font-weight-suggestion-default-matching-curated: var(
    --x-number-font-weight-suggestion-default-matching
  );
```

LuisMartinez15 (changes requested):
Good job!

tiborux:
> Given that this moves the curated highlighting logic to the `Suggestion.vue` component I think that the code in the `NextQuery.vue` component is no longer necessary (I'm talking about the `shouldHighlightCurated` method). This would also affect the tests of the component if it's removed, so those tests would have to be done in the Suggestions test file.
> 
> If we have to make that change I can do it in this PR or open a new task (or subtask of the parent of this PR).

Yes, in this case we have to remove the shouldHighlightCurated function and all the references in the next-query.vue compnent

LuisMartinez15:
> Given that this moves the curated highlighting logic to the `Suggestion.vue` component I think that the code in the `NextQuery.vue` component is no longer necessary (I'm talking about the `shouldHighlightCurated` method). This would also affect the tests of the component if it's removed, so those tests would have to be done in the Suggestions test file.
> 
> If we have to make that change I can do it in this PR or open a new task (or subtask of the parent of this PR).

We can extract the `shouldHightlightCurated` logic (with the `highlightCurated` prop) to a `mixin`, as it is used in multiple components (`related-tag.vue`, `base-suggestion.vue` and `next-query.vue`).

Here you have an example implementation:

```
// is-curated.mixin.ts

import Vue from 'vue';
import Component from 'vue-class-component';
import { Prop } from 'vue-property-decorator';

/**
 * Mixin to facilitate the check of the isCurated state of an entity.
 *
 * @public
 */
@Component
export class IsCuratedMixin<EntityType extends { isCurated?: boolean }> extends Vue {
  /**
   * Indicates if the curated entity should be highlighted.
   *
   * @public
   */
  @Prop({ default: false })
  protected highlightCurated!: boolean;

  /**
   * The entity to check the curated state.
   *
   * @public
   */
  protected entity!: EntityType;

  /**
   * Checks if the entity is curated and if it should be highlighted.
   *
   * @returns True if the entity is curated and should be highlighted.
   *
   * @public
   */
  protected get shouldHighlightCurated(): boolean {
    return this.highlightCurated && !!this.entity.isCurated;
  }
}
```
This way, you can use it in those components like this:

`export default class RelatedTag extends IsCuratedMixin<RelatedTagModel> {`

`export default class BaseSuggestion extends IsCuratedMixin<Suggestion> {`

`export default class NextQuery extends IsCuratedMixin<NextQueryModel> {`

packages/x-components/src/design-system/components/suggestion/default.scss
CachedaCodes: I did it a little bit different from your suggestion because in the first line you don't need `x-suggestion` because that's already given by the `&`. Rest was spot on, thank you 👍

LuisMartinez15 (approved):
LGTM. Good job!

packages/x-components/src/design-system/components/suggestion/default.scss
```diff
@@ -74,4 +74,42 @@
   > .x-identifier-result {
     flex: none;
   }
+
+  &.x-suggestion--is-curated {
```
tiborux, line 78:
```suggestion
  &.x-suggestion.x-suggestion--is-curated {
```

Do it like we do in tags to give them more specificity

tiborux (changes requested):

tiborux (approved):

LuisMartinez15 (approved):

PR #11, opened 2021-05-12
ci: make releases create a new PR instead of committing in main.

javieri-empathy:
# Pull Request Template #

Please include a summary of the change and which issue is fixed.

## Motivation and Context ##

After seeing today the proposed release flow we decided to improve it:

1. Manually trigger the `Prepare release` action. This will create a new PR bumping the needed packages versions, and updating its change-logs
2. Review the changes in that PR. If necessary, fix changelogs, version, whatever is needed.
3. When merging the PR (I suggest using the `squash` technique) the `Publish release` action will be automatically invoked.

## Type of change ##

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## What is the destination branch of this PR? ##

The `main` branch

## How Has This Been Tested? ##

This flow has been tested at the [Lerna Repo Playground](https://github.com/javieri-empathy/lerna-repo/)

## Checklist: ##

- [ x ] My code follows the style guidelines of this project
- [ x ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [ x ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

PR #333, opened 2022-02-16
Bugfix/ex 5374 tagging e2e tests

javieri-empathy:
This is the 3rd attempt to fix the flaky tagging e2e tests. Hopefully, this is the last part of this great series.

There were multiple things wrong with these tests. The issues are much easier to see using FireFox for running the tests.

# Cancelled fetch tagging requests

As you may know, fetch does not guarantee that the request is queued when it is executed in the unload process. In our case, what was happening in these tests is that usually the tracking request that was performed while leaving the page due to a click on a banner, result... was cancelled and this raised an error of an unhandled rejected promise.

Fortunately, Cypress can also intercept `navigator.sendBeacon` requests. So now this is the API that we are using in the mocked adapter for tracking requests.

# Mocked adapter was not used in tests.

Previously we had to pass a `useMockedAdapter=true` when navigating so this was the adapter used. When navigating to certain pages like the fake result ones this parameter was not passed. So in the add to cart tests for example, the real adapter was used. Now It is enough to load the demo view with Cypress.

# Cypress is too fast

I think this is the key of the issue of the randomly failing e2e tests, at least in Firefox. When asking Cypress to click a banner, result, promoted, or redirection, it did that immediately, as soon as the component was mounted to the DOM. Some of the siblings of that component weren't even yet mounted when this happened. But that is not really a problem. 

The problem was that there was a queue of asynchronous emitter events pending to be executed at the time the entity was clicked. These events sometimes included the `SearchTaggingChanged`, which is used by the `tagging` module to copy the query tagging with a defined debounce. As this event was not processed yet, the user click event (i.e. `UserClickedARedirection`), which is a synchronous event, was executed before it, and as a result it couldn't force to copy the query tagging.

TL;DR; Sometimes `UserClickedXXX` was executed before `SearchTaggingChanged` because of asynchronicity. This is (sadly) solved by adding an arbitrary wait. I couldn't find any other condition or guard to make cypress slowdown the click.

# Chunks

There was a warning that I saw a while ago complaining about the chunks order. Now I've removed unused views from the "production" environment, which is only used by Cypress. This should help us speed up our test server a little bit.

### DIFF
--- a/packages/x-components/src/utils/object.ts
+++ b/packages/x-components/src/utils/object.ts
@@ -140,7 +140,7 @@ export function getNewAndUpdatedKeys<ObjectType extends Dictionary>(
 }
 
 /**
- * Ensures that the given condition is met in all the entries of the object.
+ * Ensures that the given condition is met in all the non-undefined entries of the object.
  *
  * @param object - The object to check if every item meets the given condition.
  * @param condition - The condition to check in each one of the entries of the object.
@@ -156,11 +156,7 @@ export function every<ObjectType extends Dictionary>(
     index: number
   ) => boolean
 ): boolean {
-  return reduce(
-    object,
-    (paramsMatch, key, value, index) => {
-      return !paramsMatch || condition(key, value, index);
-    },
-    true as boolean
-  );
+  return Object.entries(object)
+    .filter(([, value]) => value !== undefined)
+    .every(([key, value], index) => condition(key, value, index));
 }
--- a/packages/x-components/src/utils/object.ts
+++ b/packages/x-components/src/utils/object.ts
@@ -138,3 +138,29 @@ export function getNewAndUpdatedKeys<ObjectType extends Dictionary>(
 
   return Object.keys(newValue).filter(key => !(key in oldValue) || newValue[key] !== oldValue[key]);
 }
+
+/**
+ * Ensures that the given condition is met in all the entries of the object.
+ *
+ * @param object - The object to check if every item meets the given condition.
+ * @param condition - The condition to check in each one of the entries of the object.
+ *
+ * @returns True when all the entries pass the condition. False otherwise.
+ * @public
+ */
+export function every<ObjectType extends Dictionary>(
+  object: ObjectType,
+  condition: (
+    key: keyof ObjectType,
+    value: Exclude<ObjectType[keyof ObjectType], undefined>,
+    index: number
+  ) => boolean
+): boolean {
+  return reduce(
+    object,
+    (paramsMatch, key, value, index) => {
+      return !paramsMatch || condition(key, value, index);
+    },
+    true as boolean
+  );
+}
--- a/packages/x-components/tests/e2e/cucumber/filters/sliced-filters.spec.ts
+++ b/packages/x-components/tests/e2e/cucumber/filters/sliced-filters.spec.ts
@@ -3,7 +3,7 @@ import { And, Given, Then, When } from 'cypress-cucumber-preprocessor/steps';
 let totalFilters: number, hiddenFilters: number;
 
 Given('following config: max of sliced filters {int}', (slicedFiltersMax: number) => {
-  cy.visit('/?useMockedAdapter=true');
+  cy.visit('/?');
   cy.getByDataTest('sliced-filters-max').clear().type(slicedFiltersMax.toString());
 });
 
--- a/packages/x-components/tests/e2e/cucumber/filters/sliced-filters.spec.ts
+++ b/packages/x-components/tests/e2e/cucumber/filters/sliced-filters.spec.ts
@@ -3,7 +3,7 @@ import { And, Given, Then, When } from 'cypress-cucumber-preprocessor/steps';
 let totalFilters: number, hiddenFilters: number;
 
 Given('following config: max of sliced filters {int}', (slicedFiltersMax: number) => {
-  cy.visit('/?');
+  cy.visit('/');
   cy.getByDataTest('sliced-filters-max').clear().type(slicedFiltersMax.toString());
 });
 
--- a/packages/x-components/tests/e2e/cucumber/next-queries/next-queries.spec.ts
+++ b/packages/x-components/tests/e2e/cucumber/next-queries/next-queries.spec.ts
@@ -21,7 +21,7 @@ Given(
       }
     };
 
-    cy.visit('/?useMockedAdapter=true', {
+    cy.visit('/?', {
       qs: {
         xModules: JSON.stringify(config)
       }
--- a/packages/x-components/tests/e2e/cucumber/next-queries/next-queries.spec.ts
+++ b/packages/x-components/tests/e2e/cucumber/next-queries/next-queries.spec.ts
@@ -21,7 +21,7 @@ Given(
       }
     };
 
-    cy.visit('/?', {
+    cy.visit('/', {
       qs: {
         xModules: JSON.stringify(config)
       }
--- a/packages/x-components/tests/e2e/cucumber/popular-searches/popular-searches.spec.ts
+++ b/packages/x-components/tests/e2e/cucumber/popular-searches/popular-searches.spec.ts
@@ -18,7 +18,7 @@ Given(
         }
       }
     };
-    cy.visit('/?', {
+    cy.visit('/', {
       qs: {
         xModules: JSON.stringify(config)
       }
--- a/packages/x-components/tests/e2e/cucumber/popular-searches/popular-searches.spec.ts
+++ b/packages/x-components/tests/e2e/cucumber/popular-searches/popular-searches.spec.ts
@@ -18,7 +18,7 @@ Given(
         }
       }
     };
-    cy.visit('/?useMockedAdapter=true', {
+    cy.visit('/?', {
       qs: {
         xModules: JSON.stringify(config)
       }
--- a/packages/x-components/tests/e2e/cucumber/query-suggestions/query-suggestions.spec.ts
+++ b/packages/x-components/tests/e2e/cucumber/query-suggestions/query-suggestions.spec.ts
@@ -18,7 +18,7 @@ Given(
         }
       }
     };
-    cy.visit('/?', {
+    cy.visit('/', {
       qs: {
         xModules: JSON.stringify(config)
       }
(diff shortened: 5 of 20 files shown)


packages/x-components/src/utils/object.ts
```diff
@@ -138,3 +138,29 @@ export function getNewAndUpdatedKeys<ObjectType extends Dictionary>(
...
+/**
+ * Ensures that the given condition is met in all the entries of the object.
+ *
+ * @param object - The object to check if every item meets the given condition.
+ * @param condition - The condition to check in each one of the entries of the object.
+ *
+ * @returns True when all the entries pass the condition. False otherwise.
+ * @public
+ */
+export function every<ObjectType extends Dictionary>(
```
javieri-empathy, line 151: Initially I had manually stubbed the navigator `sendBeacon` method, and then used this `every` method to perform checks in the tests. Now this is not needed anymore, but I think this function can help us in the future.

packages/x-components/tests/e2e/cucumber/tagging/tagging.feature
```diff
@@ -5,9 +5,7 @@ Feature: Tagging component
     And   a suggestions API
     And   a related tags API
     And   a recommendations API with a known response
-    And   a query tagging API
```
CachedaCodes, line 8: These 3 steps you removed were not used anywhere else, so I think their logic can be removed from `mocked-responses.spec.ts` (unless they might be used in the future instead of the new `a tracking API with a known response`).

packages/x-components/tests/e2e/cucumber/tagging/tagging.spec.ts
```diff
@@ -1,22 +1,28 @@
 import { Given, Then, When } from 'cypress-cucumber-preprocessor/steps';
 
-Given('a URL with query parameter {string}', (query: string) => {
-  cy.visit(`/?useMockedAdapter=true&q=${query}`);
+Given('a tracking API with a known response', () => {
```
CachedaCodes, line 3: Would it make sense to move this to `mocked-responses.spec.ts` with the other mocked tracking

CachedaCodes (changes requested):

packages/x-components/src/utils/object.ts
LuisMartinez15: Hey bro! You're human! Welcome to the human realm! :trollface:

In a more serious note, we are fine with the addition of this utility. However, it currently has a bug. The `paramsMatch` flag "accumulator" is not working as intended, as it is being reset each iteration. We should not invert its value and we should use an `&&` operator instead of the `||` one, so the previous value is always taken into account. For example, in case in a previous iteration `paramsMatch` resulted in a `false` value, we shouldn't even care about the `condition` of the current/next ones, as at least one object key/entry has already failed the condition.

You can clearly see this behaviour happening in the second test you made, if you invert the order in the object, { a: '1', b: 2 }, you will see that the test fails, where it shouldn't, as we are only checking, in the end, that the last entry/key fulfils the condition.

As a side note, we could even optimise this method so it breaks the reduce execution whenever the condition is not met (probably not worth it for the majority of use cases, tho).

packages/x-components/tests/e2e/cucumber/filters/sliced-filters.spec.ts
```diff
@@ -3,7 +3,7 @@ import { And, Given, Then, When } from 'cypress-cucumber-preprocessor/steps';
 let totalFilters: number, hiddenFilters: number;
 
 Given('following config: max of sliced filters {int}', (slicedFiltersMax: number) => {
-  cy.visit('/?useMockedAdapter=true');
+  cy.visit('/?');
```
LuisMartinez15, line 6: I guess we can also remove the `?` here, right? Same for all the other files where you've applied this change (I'll just leave code suggestions in case you want to batch commit them instead of changing them manually):

```suggestion
  cy.visit('/');
```

packages/x-components/tests/e2e/cucumber/next-queries/next-queries.spec.ts
```diff
@@ -21,7 +21,7 @@ Given(
       }
     };
 
-    cy.visit('/?useMockedAdapter=true', {
+    cy.visit('/?', {
```
LuisMartinez15, line 24:
```suggestion
    cy.visit('/', {
```

packages/x-components/tests/e2e/cucumber/popular-searches/popular-searches.spec.ts
```diff
@@ -18,7 +18,7 @@ Given(
         }
       }
     };
-    cy.visit('/?useMockedAdapter=true', {
+    cy.visit('/?', {
```
LuisMartinez15, line 21:
```suggestion
    cy.visit('/', {
```

packages/x-components/tests/e2e/cucumber/query-suggestions/query-suggestions.spec.ts
```diff
@@ -18,7 +18,7 @@ Given(
         }
       }
     };
-    cy.visit('/?useMockedAdapter=true', {
+    cy.visit('/?', {
```
LuisMartinez15, line 21:
```suggestion
    cy.visit('/', {
```

packages/x-components/tests/e2e/cucumber/recommendations/recommendations.spec.ts
```diff
@@ -11,7 +11,7 @@ Given('following config: max items to store is {int}', (maxItemsToRequest: numbe
     }
   };
 
-  cy.visit('/?useMockedAdapter=true', {
+  cy.visit('/?', {
```
LuisMartinez15, line 14:
```suggestion
  cy.visit('/', {
```

packages/x-components/tests/e2e/cucumber/related-tags/related-tags.spec.ts
```diff
@@ -16,7 +16,7 @@ Given(
       }
     };
 
-    cy.visit('/?useMockedAdapter=true', {
+    cy.visit('/?', {
```
LuisMartinez15, line 19:
```suggestion
    cy.visit('/', {
```

packages/x-components/tests/e2e/cucumber/search-box/search-box.spec.ts
```diff
@@ -19,7 +19,7 @@ Given(
         }
       }
     };
-    cy.visit('/?useMockedAdapter=true', {
+    cy.visit('/?', {
```
LuisMartinez15, line 22:
```suggestion
    cy.visit('/', {
```

packages/x-components/tests/e2e/cucumber/tagging/tagging.spec.ts
```diff
@@ -29,15 +35,15 @@ Then('result click tagging request is triggered', () => {
 });
 
 Then('query tagging request should be triggered', () => {
-  cy.wait('@queryTagging');
+  cy.wait('@queryTagging').should('exist');
 });
 
 Then('query tagging request has been triggered', () => {
-  cy.get('@queryTagging').should('exist');
+  cy.get('@queryTagging', { timeout: 0 }).should('exist');
```
LuisMartinez15, line 42: Sadly, I'm afraid to say this is causing some tests to fail locally. And they keep failing randomly (sometimes they pass sometimes they don't 🤷🏿‍♂️ ).

image

Both tests failing in this image is due to same error in this line.

LuisMartinez15 (changes requested):

packages/x-components/tests/e2e/cucumber/filters/sliced-filters.spec.ts
javieri-empathy: Global replaces in the morning do not work well

packages/x-components/src/utils/object.ts
```diff
@@ -138,3 +138,29 @@ export function getNewAndUpdatedKeys<ObjectType extends Dictionary>(
...
+    key: keyof ObjectType,
+    value: Exclude<ObjectType[keyof ObjectType], undefined>,
+    index: number
+  ) => boolean
+): boolean {
+  return reduce(
+    object,
+    (paramsMatch, key, value, index) => {
+      return !paramsMatch || condition(key, value, index);
+    },
```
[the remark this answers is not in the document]
javieri-empathy: You are right, I tried to go clever in this one and failed miserably. I changed the approach to use `Object.entries(object).every`, and we will gain that early return as soon as one condition fails.

packages/x-components/tests/e2e/cucumber/tagging/tagging.spec.ts
LuisMartinez15: As we talked in private, this seems to be working as expected (something might have been wrong in my local environment by the time I ran the tests).

mavmaf (commented):
No issues seen after running it several times locally with different throttling and browsers.
Just one favor: it seems that at some point cypress.json file was merged by mistake. Please modify line #11 to the following:
`  "testFiles": ["cucumber/**/*.features"],`

LuisMartinez15:
> No issues seen after running it several times locally with different throttling and browsers. Just one favor: it seems that at some point cypress.json file was merged by mistake. Please modify line #11 to the following: ` "testFiles": ["cucumber/**/*.features"],`

I think you had it changed locally and checked out this pull request (resulting in a merge with its content changed to your local modifications), because I see the `cypress.json` file with the same content you're asking to apply. This file hasn't been changed in pull request 😅

mavmaf:
> > No issues seen after running it several times locally with different throttling and browsers. Just one favor: it seems that at some point cypress.json file was merged by mistake. Please modify line #11 to the following: ` "testFiles": ["cucumber/**/*.features"],`
> 
> I think you had it changed locally and checked out this pull request (resulting in a merge with its content changed to your local modifications), because I see the `cypress.json` file with the same content you're asking to apply. This file hasn't been changed in pull request 😅

Hahahaha F, what a _bocachancla_

CachedaCodes (approved):

LuisMartinez15 (approved):
LGTM

Good job!

PR #341, opened 2022-02-22
build: remove `polyfills` support

LuisMartinez15:
EX-3559

Remove unnecessary polyfills support.

## How has this been tested?

Build x-components library in this pull request and install it in the x-archetype equivalent pull request to check that the build is still working and no polyfills are added.


javieri-empathy (changes requested):

packages/x-components/src/x-installer/x-installer/x-installer.ts
```diff
@@ -103,12 +104,17 @@ export class XInstaller {
...
+   * object specifying the {@link SnippetConfig | snippet config}.
+   *
+   *
    * @param options - The {@link InstallXOptions}.
    *
    * @public
    */
   public constructor(protected readonly options: InstallXOptions) {
     this.createAPI();
+    this.autoInit();
```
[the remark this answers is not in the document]
diegopf: I like this suggestion. To me feels strange that just creating the installer also initialises the app out of the box.  However, wouldn't be better the following?:

1. Refactor `autoInit` into something like `getSnippetConfig(): SnippetConfig` .
2.  Refactor `init` to call `getSnippetConfig` first of all.
3. Just call `new XInstaller(options).init()`.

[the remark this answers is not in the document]
LuisMartinez15: Even though I like your approach @diegopf , I think we should still support the option of passing a `SnippetConfig` as a parameter to the `XInstaller.init` function, as `x-components` can be used as a `library`, hence there shouldn't be a need of injecting the `snippet config` object into the `window` object.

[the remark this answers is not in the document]
javieri-empathy: @diegopf The point is that there are 2 ways that the XInstaller can be initialised:

1. There is a defined `initX` object in `window` at the time the x-components are downloaded
  a. It is an static object of `SnippetConfig` type.
  b. It is a function that should return a `SnippetConfig` object.
2. There is no `initX` object defined. Therefore, the developer that is integrating the XComponents into his website should call `window.X.init(snippetConfig)` manually when he is ready, which will call `XInstaller.init` method.

With your suggestion I'm not sure how this last requirement fits. If we didn't need to support it I would like it. Maybe I should have suggested to rename `autoInit` to `tryAutoInit` which describes better that it might not actually initialise the project?.

In the case that @LuisMartinez15 says, using XComponents as a standard components library in your Vue project, we shouldn't use the `XInstaller`at all, that's why I'm willing to move it together with other archetype utilities into its own project, so there is no confusion with that.

packages/x-components/src/x-installer/x-installer/__tests__/x-installer.spec.ts
```diff
@@ -155,4 +155,34 @@ describe('testing `XInstaller` utility', () => {
...
+
+  describe('testing autoInit function', () => {
+    const mockedInit = jest.spyOn(XInstaller.prototype as any, 'init');
+
+    beforeEach(() => {
+      delete window.initX;
+    });
+
+    it('auto initializes XComponents if window.initX is a function', () => {
+      window.initX = jest.fn().mockReturnValue(snippetConfig);
```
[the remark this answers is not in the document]
LuisMartinez15: There is not, so thanks for the tip. I've refactored it.

packages/x-components/src/x-installer/x-installer/x-installer.ts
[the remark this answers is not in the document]
LuisMartinez15: I've refactored it based on your suggestions. Let me know if you find it okey or want to suggest further changes.

packages/x-components/src/x-installer/x-installer/types.ts
```diff
@@ -79,3 +79,19 @@ export interface ExtraPluginsOptions {
...
  */
 export type VueConstructorPartialArgument = Partial<ConstructorParameters<VueConstructor>[0]>;
+
+/**
+ * Interface for the returned type of the XInstaller.init function.
+ *
+ * @public
+ */
+export interface InitWrapper {
+  /** The Vue component used as root of the application. */
```
javieri-empathy, line 89: This is not the component, but the instance of the component

```suggestion
  /** The root Vue instance of the application. */
```

packages/x-components/src/x-installer/x-installer/types.ts
```diff
@@ -79,3 +79,19 @@ export interface ExtraPluginsOptions {
  * @public
  */
 export type VueConstructorPartialArgument = Partial<ConstructorParameters<VueConstructor>[0]>;
+
+/**
+ * Interface for the returned type of the XInstaller.init function.
```
javieri-empathy, line 84:
```suggestion
 * Interface for the returned type of the {@link XInstaller.init} function.
```

packages/x-components/src/x-installer/x-installer/types.ts
```diff
@@ -79,3 +79,19 @@ export interface ExtraPluginsOptions {
...
+ * @public
+ */
+export interface InitWrapper {
+  /** The Vue component used as root of the application. */
+  app: Vue | undefined;
+  /** The API to expose globally. */
+  api: XAPI | undefined;
+  /** The {@link XBus} used in the {@link XPlugin}. */
+  bus: XBus;
+  /** The XPlugin which will be installed. */
```
javieri-empathy, line 95:
```suggestion
  /** The installed {@link XPlugin} instance. */
```

javieri-empathy (changes requested):

packages/x-components/src/x-installer/x-installer/x-installer.ts
```diff
@@ -128,38 +133,54 @@ export class XInstaller {
   }
 
   /**
-   * Receives the {@link SnippetConfig | snippet config}, installs the plugin and initializes
-   * the Vue application.
+   * Retrieves the {@link SnippetConfig | snippet config} it is defined in the window.initX.
    *
-   * @param snippetConfig - The {@link SnippetConfig} that receives from snippet
-   * integration.
+   * @returns The snippet config if it is defined. Else undefined.
```
diegopf, line 138:
```suggestion
   * @returns The snippet config if it is defined or undefined otherwise.
```

packages/x-components/src/x-installer/x-installer/x-installer.ts
```diff
@@ -128,38 +133,54 @@ export class XInstaller {
...
+
+      return {
+        api: this.api,
+        app,
+        bus,
+        plugin
+      };
+    }
+
+    return Promise.resolve();
```
diegopf, line 183: Shouldn't be better `return Promise.reject()` in this case? I'm thinking that basically we failed to install anthing.

diegopf (changes requested):

packages/x-components/src/x-installer/x-installer/x-installer.ts
```diff
@@ -128,38 +133,54 @@ export class XInstaller {
...
-  }> {
-    const adapterConfig = this.getAdapterConfig(snippetConfig);
-    this.applyConfigToAdapter(adapterConfig);
-    const bus = this.createBus();
-    const pluginOptions = this.getPluginOptions();
-    const plugin = this.installPlugin(pluginOptions, bus);
-    const extraPlugins = await this.installExtraPlugins(snippetConfig, bus);
-    const app = this.createApp(extraPlugins, snippetConfig);
-    this.api?.setBus(bus);
+  async init(snippetConfig?: SnippetConfig): Promise<InitWrapper | void> {
```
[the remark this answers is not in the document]
LuisMartinez15: I didn't know you could overload a function call directly in its implementation 🤦🏿 
I didn't know you could call a function to set the default value of a parameter 🤦🏿 

TIL a lot. Thanks 😅

packages/x-components/src/x-installer/x-installer/x-installer.ts
javieri-empathy: If this was a normal method that was doing only 1 thing yes, but as we have to support the 2 ways of loading x-components, and we don't want to stop in every client to ask them how they prefer to load, I would just return the Promise resolve, even if looks a little weird.

packages/x-components/src/x-installer/x-installer/types.ts
```diff
@@ -79,3 +79,19 @@ export interface ExtraPluginsOptions {
...
+ * Interface for the returned type of the {@link XInstaller.init} function.
+ *
+ * @public
+ */
+export interface InitWrapper {
+  /** The root Vue instance of the application. */
+  app: Vue;
+  /** The {@link XAPI} to expose globally. */
+  api: XAPI;
+  /** The {@link XBus} used in the {@link XPlugin}. */
```
javieri-empathy, line 93: Remember that this two properties were optional before, and thanks for extracting them.

```suggestion
  app?: Vue;
  /** The {@link XAPI} to expose globally. */
  api?: XAPI;
  /** The {@link XBus} used in the {@link XPlugin}. */
```

packages/x-components/src/x-installer/x-installer/types.ts
```diff
@@ -79,3 +79,19 @@ export interface ExtraPluginsOptions {
  * @public
  */
 export type VueConstructorPartialArgument = Partial<ConstructorParameters<VueConstructor>[0]>;
+
+/**
+ * Interface for the returned type of the {@link XInstaller.init} function.
```
javieri-empathy, line 84:
```suggestion
 * Interface for the returned type of the {@link XInstaller.(init:1)} function.
```

javieri-empathy (changes requested):

diegopf (dismissed):
Given that the build will work with @javieri-empathy latest suggestions, LGTM

diegopf (approved):

javieri-empathy (approved):

PR #343, opened 2022-02-22
chore: update unused & at risk dependencies, add dependabot.yaml and clean package-locks

CachedaCodes:
[EX-5299](https://searchbroker.atlassian.net/browse/EX-5299)

## Motivation and context

Some dependencies were flagged as vulnerabilities and appeared in the dependabot alerts, `npm audit` and `npx lerna-audit`. This PR solves these vulnerabilities by updating dependencies, removing unused ones and cleaning up `package-lock.json` files that were no longer needed.

On top of the dependency changes a `dependabot.yaml` was added so the GitHub dependabot runs a check of possible vulnerabilities every week. It checks the root `package-lock` and the one in the `eslint-plugin-x` package.
`eslint-plugin-x` is the only package with its own `package-lock` due to the version of `prettier` having to be forced in it. The last version `prettier` changed the way they structured the CSS classes in the templates and that breaks the project.

## Type of change

- [X] `Main`
- [ ] Other. Specify:

## How has this been tested?

Running `npm audit` and `npx lerna-audit` should show that there are no vulnerabilities. Regarding dependabot, it should be extracting the flagged vulnerable versions of the dependencies from the same places that `npm` and `lerna`.


packages/x-components/package.json
```diff
@@ -84,12 +84,12 @@
     "@cypress/webpack-dev-server": "~1.4.0",
```
tiborux, line 84: update nanoid from 3.1.30 to 3.1.31

.github/dependabot.yaml
```diff
@@ -0,0 +1,15 @@
+version: 2
+updates:
+
+  # Maintain dependencies for the root package
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
tiborux, line 8: By default, the day is on Monday, so if we are going to check the dependencies on Friday, we should move it to Friday. Also, the time, by default its a random time, but we can set an hour.

https://docs.github.com/es/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#scheduleday

```suggestion
      interval: "weekly"
      day: "friday"
      time: "08:00"
```

LuisMartinez15: Add the `timezone: "Europe/Madrid"` property too, so it is always based in our timezone instead of defaulting to UTC.

packages/eslint-plugin-x/README.md
```diff
@@ -6,7 +6,6 @@ custom plugin, but it includes also style formatter and good practices outside E
 
 - [ESLint](https://eslint.org/) and third-party plugins
 - [Prettier](https://prettier.io/)
-- [standard-version](https://github.com/conventional-changelog/standard-version)
```
LuisMartinez15, line 9: Why are we removing the `standard-version` references from this readme? I've noticed we are not using it anymore in the `monorepo`, but it is still exported and may be used by anyone using this `package`. Should we remove this config entirely from the package?

LuisMartinez15 (changes requested):
Good job!

LuisMartinez15 (dismissed):

LuisMartinez15:
I guess removing the `standard-version-config.js` in the `eslint-plugin-x` is a `breaking-change` in that package.

LuisMartinez15 (approved):
LGTM

tiborux (approved):

PR #359, opened 2022-03-01
chore: remove standard-version from eslint-plugin-x

CachedaCodes:
BREAKING CHANGE: standard-version dependency has been removed from eslint-plugin-x package

[EX-5526](https://searchbroker.atlassian.net/browse/EX-5526)

## Motivation and context
As part of the cleaning up of dependencies of the monorepo we needed to remove the `standard-version` dependency. This would be a breaking change so it was decided that, to prevent the changelogs of all the packages to reference the breaking change of a single one, this change be done in it's own PR.

## Type of change

- [ X ] Breaking change (fix or feature that causes existing functionality to not work as expected)

- [ X ] `Main`
- [ ] Other. Specify:


packages/eslint-plugin-x/package.json
```diff
@@ -42,8 +42,7 @@
     "eslint-plugin-prettier": "^3.1.3",
     "eslint-plugin-tsdoc": "^0.2.5",
     "eslint-plugin-vue": "^6.2.2",
-    "prettier": "^2.1.2",
-    "standard-version": "^8.0.0"
```
javieri-empathy, line 46: There is still a reference to standard version in `eslint-plugin-x/lib/index.js`

guillei10: Yes I also see this. I think fixing it and we are ready to merge.

guillei10 (changes requested):

PR #364, opened 2022-03-07
[EX-5525] Review E2E tests in Firefox

mavmaf:
## Motivation and context
After running the E2E several times with different throttling, it was observed that 3 different tests were randomly failing in Firefox:

- Clear selected filters → Scenario 3. Remove single filter.
- History queries component → Scenario 6. Number and order of rendered history queries.
- Scroll → Scenario 1. Scroll is kept in the URL.

After analyzing each scenario, minor changes were needed.

## Type of change

- [x] `Main`
- [ ] Other. Specify:

## How has this been tested?

As usual, please change line 11 in `cypress.json` for the following to run the tests independently:
`  "testFiles": ["cucumber/**/*.feature"],`

## Checklist:

### DIFF
--- a/packages/x-components/tests/e2e/cucumber/history-queries.feature
+++ b/packages/x-components/tests/e2e/cucumber/history-queries.feature
@@ -17,10 +17,6 @@ Feature: History queries component
     Then  the searched query is displayed in the search-box
     And   the searched query is not displayed in history queries if <hideIfEqualsQuery> is true
 
-    Examples:
-      | hideIfEqualsQuery | debounceInMs | maxItemsToStore | maxItemsToRender | instant | historyQueryItem | list                                                                                                             |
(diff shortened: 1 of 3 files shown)


packages/x-components/tests/e2e/cucumber/history-queries.feature
```diff
@@ -81,7 +81,8 @@ Feature: History queries component
     And   a "<list>" of queries already searched
     When  search-input is focused
     And   history query number <historyQueryItem> is clicked
-    And   clear search button is pressed
+    Then  related results are displayed
+    When  clear search button is pressed
```
javieri-empathy, line 85: Is this issue related with firefox doing things too fast once again? What I saw after making it fail like it was before is that the query is properly cleared from the input, but not from other modules.

This is what Cypress saw
image

But when you clear the query, either by using the clear button or the delete key, if I remember correctly, every module but the next queries should reset its state, which is not happening in the test.
I think this might happen because of the asynchronicity of certain events, and how FireFox handles the task queue, but I would need to dig deeper into this.

javieri-empathy (approved):

guillei10 (approved):

PR #365, opened 2022-03-08
feat: move related tags concatenation logic EX-5410 

javieri-empathy:
- Standardises `query` and  `setQuery` in all modules.
- Related tag concatenation is now done in the x-components.
- BREAKING CHANGE: Removes `previous` and `selected` properties from `RelatedTag` model.
- BREAKING CHANGE Remove related tags query mapper from the `x-adapter` package.


packages/search-adapter/src/empathy/__tests__/request-mappers/empathy-queryable-request-mapper.spec.ts
```diff
@@ -1,19 +1,13 @@
-import { RelatedTag } from '@empathyco/x-types';
 import { Container } from 'inversify';
```
javieri-empathy, line 1: As adapter files have been auto-formatted by prettier they are going to be a little harder to review, but the intention was to only remove the related tags  mapping from the request mappers.

packages/x-components/src/__stubs__/related-tags-stubs.factory.ts
```diff
@@ -15,15 +15,6 @@ export function getRelatedTagsStub(amount = 12): RelatedTag[] {
   );
 }
 
-/**
- * Function to create selected related tags stub.
- *
- * @returns Array of selected related tags stub.
- */
-export function getSelectedRelatedTagsStub(): RelatedTag[] {
```
javieri-empathy, line 23: We don't need this function anymore, there is no difference in the model of a selected related tag or an unselected one.

packages/x-components/src/x-modules/search/store/index.ts
```diff
@@ -2,5 +2,6 @@ export * from './actions/fetch-and-save-search-response.action';
 export * from './actions/fetch-search-response.action';
```
javieri-empathy, line 2: There is a big mess with x-module exports. There are actions or getters that are not exported, and some times the name even collides with the wires. I'm planning to fix this in another PR.

packages/x-components/src/x-modules/search/store/types.ts
```diff
@@ -73,14 +74,16 @@ export interface SearchGetters {
   /** The adapter request object for retrieving the results, or null if there is not
    * valid data to create a request. */
   request: InternalSearchRequest | null;
+  /** The combination of the query and the selected related tags. */
+  query: string;
```
javieri-empathy, line 78: I decided to leave this `query` and the `setQuery` mutation in every module so we can link it through the documentation.

diegopf: I was finishing the review and I just saw this, you are so good that answer my questions before I asked them 🤣

diegopf (commented):
LGTM, I'm just adding a comment to find out about the EX-2163 issue and if we can remove the TODO comments

packages/search-adapter/src/types/requests.types.ts
```diff
@@ -68,15 +67,20 @@ export interface RequestOptions {
  * @public
  */
 export interface ExtraParamsRequest {
-  [key: string]: unknown
+  [key: string]: unknown;
 }
 
 /**
  * TODO https://searchbroker.atlassian.net/browse/EX-2163
```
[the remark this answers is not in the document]
javieri-empathy: That issue was just "Document the whole adapter"

guillei10 (approved):

packages/x-components/src/x-modules/history-queries/store/types.ts
```diff
@@ -57,7 +58,7 @@ export interface HistoryQueriesGetters {
  *
  * @public
  */
-export interface HistoryQueriesMutations {
+export interface HistoryQueriesMutations extends QueryMutations {
```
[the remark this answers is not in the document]
javieri-empathy: Just answering here too for the future haha, but yeah, my reason was that with API extractor we can't link to interface inherited properties, we have to link the base interface, and I don't like that in the documentation. As types do not affect runtime performance, I think we can afford some duplication just to provide better API documentantion. But I would love to hear your opinion on this.

diegopf (approved):

PR #370, opened 2022-03-15
Feature/ex 5387 create adapter necessary types

javieri-empathy:
Adds the types needed to implement the first version of the adapter.


packages/x-adapter/src/adapter.types.ts
```diff
@@ -0,0 +1,70 @@
+import { HttpClient, RequestOptions } from './http-client.types';
+import { Mapper } from './mapper.types';
+
+/**
+ * A facade containing all the different {@link EndpointAdapter} methods.
+ *
+ * @public
+ */
+export interface Adapter {}
```
javieri-empathy, line 9: I've intentionally left this interface empty, so we can add the requests and responses later one by one.

packages/x-adapter/src/mapper.types.ts
```diff
@@ -0,0 +1,30 @@
...
+  /**
+   * When the mapper is used to transform a request or a response, the endpoint that
+   * was or is going be used.
+   */
+  endpoint?: string;
+  /**
+   * If multiple mappers are taking charge of mapping the same property, the partial result of
+   * applying them to it.
+   */
+  to?: unknown;
```
javieri-empathy, line 28: I decided to move this to the context to hide it a little bit. I'm not sure whether we are going to need the `to` with the new  schema mappers. Relying on the previous mapper result is a code smell, an indirect dependency that you don't have control of, so I would try to avoid it in the core adapter implementation.

In the case we need it we can always cast it, but (for once) I would rather not complicate types with unnecessary generics just to type scenarios that should not happen.

packages/x-adapter/src/adapter.types.ts
```diff
@@ -0,0 +1,73 @@
...
+  endpoint?: string | Mapper<Request, string>;
+  /**
+   * The {@link HttpClient} to use for the requests.
+   */
+  httpClient?: HttpClient;
+  /**
+   * The {@link Mapper} in charge of adapting the request object so the backend API can
+   * understand it.
+   */
+  requestMapper: Mapper<Request, Record<string, string | boolean | number>>;
```
javieri-empathy, line 67: Request mapper should transform the given object to a record of primitive values that are serializable for the request.

packages/x-adapter/src/adapter.types.ts
```diff
@@ -0,0 +1,73 @@
...
+  /**
+   * The {@link Mapper} in charge of adapting the response object to whatever format or
+   * shape it is needed.
+   */
+  responseMapper: Mapper<any, Response>;
```
javieri-empathy, line 72: Response mapper should transform the API response (which should be unknown) to the desired shape. However if we set `unknown` as type we are going to force checking the API response with type guards or cast it in every single mapper. Because of the way we are going to build these mappers, I would leave this as `any` to ease typing, But I'd like to hear your feedback here.

packages/x-adapter/src/mapper.types.ts
```diff
@@ -0,0 +1,30 @@
...
+ * @public
+ */
+export interface MapperContext {
+  /**
+   * When the mapper is used to transform a response, the request that triggered it.
+   */
+  requestParameters?: Record<string, string | boolean | number>;
+  /**
+   * When the mapper is used to transform a request or a response, the endpoint that
+   * was or is going be used.
```
tiborux, line 21:
```suggestion
   * was or is going to be used.
```

packages/x-adapter/src/adapter.types.ts
tiborux: Should be this property not required? In the task says that requestMapper and responseMapper should not be required.

packages/x-adapter/src/adapter.types.ts
LuisMartinez15: We agree with your approach. @tiborux and myself were discussing how it is going to work and your proposal seems to be the better one (if we understood it correctly 😅 )

LuisMartinez15 (changes requested):
Good job! This is starting to take shape!

packages/x-adapter/src/adapter.types.ts
tiborux: As Luis said, we think that we understood it correctly, in this case, if we send any in the mapper, we can avoid casting the response, but with the new utility paths, we can get the "type" of this any, makes sense?

packages/x-adapter/package.json
```diff
@@ -11,12 +11,28 @@
...
+  ],
   "repository": {
     "type": "git",
     "url": "git+https://github.com/empathyco/x.git",
     "directory": "packages/x-adapter"
   },
-  "scripts": {},
+  "scripts": {
+    "prebuild": "rimraf ./dist ./types",
+    "build": "tsc --project tsconfig.json && tsc --project tsconfig.esm.json",
```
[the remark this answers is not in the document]
LuisMartinez15: M1 vibes :trollface:

LuisMartinez15 (approved):
LGTM. Good job!

tiborux (approved):

PR #372, opened 2022-03-15
build(dependencies): prevent `node-ipc` vulnerability

javieri-empathy:
There is a new "popular" vulnerability in the :js: ecosystem. Apparently, the node-ipc  package [has been hijacked by its own creator.](https://github.com/RIAEvangelist/node-ipc/issues/233#issuecomment-1068182278)
Version 9.2.2 and 11.0.0 uses a module called [peacenotwar](https://github.com/RIAEvangelist/peacenotwar) which will try to create a file in desktop protesting against the war. We are potentially affected by the 9.2.2 version, as Vue CLI uses this package's 9 version
There is another vulnerability, in theory in 10.1.3 (but I haven't seen it in the published versions)which is more dangerous, trying to override some files for people from Russia or Belarus.


LuisMartinez15 (approved):
Thanks for the quick response!

tajespasarela (approved):

PR #373, opened 2022-03-16
feat(utils): add `interpolate` function

javieri-empathy:
Adds a small utility to interpolate parameters in the `endpoint` field of each one of the `EndpointAdapter`.

### DIFF
--- a/packages/x-adapter/jest.config.js
+++ b/packages/x-adapter/jest.config.js
@@ -1,0 +1,7 @@
+module.exports = {
+  moduleFileExtensions: ['ts', 'js'],
+  transform: {
+    '^.+\\.ts?$': 'ts-jest'
+  },
+  testMatch: ['<rootDir>/**/*.spec.ts']
+};
--- a/packages/x-adapter/src/utils/interpolate.ts
+++ b/packages/x-adapter/src/utils/interpolate.ts
@@ -15,6 +15,13 @@
 const STRING_PARAMETERS = /{([^}]+)}/g;
 
 /**
+ * Shape of the optional head and tail parts of the {@link STRING_PARAMETER_CONTENT} regex.
+ * This can be anything wrapped into parentheses.
+ *
+ * @internal
+ */
+const HEAD_OR_TAIL = '(?:\\((.+)\\))?';
+/**
  * Syntax of a single string parameter. A string parameter shape is composed by
  * the name of the property that should be replaced. This property name can be preceded
  * or followed by an optional string to prepend or to append to the property value
@@ -30,7 +37,7 @@ const STRING_PARAMETERS = /{([^}]+)}/g;
  * ```
  * @internal
  */
-const STRING_PARAMETER_CONTENT = /^(?:\((.+)\))?([^(]+)(?:\((.+)\))?$/g;
+const STRING_PARAMETER_CONTENT = new RegExp(`^${HEAD_OR_TAIL}([^(]+)${HEAD_OR_TAIL}$`, 'g');
 
 /**
  * Interpolates different parameters into a string.
--- a/packages/x-adapter/src/utils/interpolate.ts
+++ b/packages/x-adapter/src/utils/interpolate.ts
@@ -1,0 +1,99 @@
+/**
+ * Syntax to detect and extract string parameters. A string parameter contains a property name
+ * with an optional header or tail to concatenate wrapped in curly braces (`{}`).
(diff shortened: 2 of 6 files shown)


packages/x-adapter/jest.config.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  moduleFileExtensions: ['ts', 'js'],
+  transform: {
+    '^.+\\.ts?$': 'ts-jest'
```
tiborux, line 4: Reading this article: https://miyauchi.dev/posts/speeding-up-jest/

We test to use different transformers to improve the perfomance:

ts-jest:

image

@swc/jest:

image

esbuild-jest:

image

I dont know if we should give a try to improve the jest performance.

packages/x-adapter/src/utils/interpolate.ts
```diff
@@ -0,0 +1,99 @@
...
+ * ```js
+ *   "env" // No optional head or tail to concatenate.
+ *   "(.)env" // The `.` character will be prepended as long as the `env` property is defined.
+ *   "env(-api)" // The `-api` string will be appended as long as the `env` property is defined.
+ *   "(api-)env(.)" // The `api-` string and the `.` character will be added as long as
+ *   // the `env` property is defined.
+ * ```
+ * @internal
+ */
+const STRING_PARAMETER_CONTENT = /^(?:\((.+)\))?([^(]+)(?:\((.+)\))?$/g;
```
tiborux, line 33: What do you thin about splitting this?

`
const optionalParameter = '(?:\\((.+)\\))?';

const STRING_PARAMETER_CONTENT = new RegExp(
  `^${optionalParameter}([^(]+)${optionalParameter}$`,
  'g'
);
`

tiborux (changes requested):

packages/x-adapter/src/utils/interpolate.ts
```diff
@@ -0,0 +1,99 @@
...
+ *    instance: 'demo'
+ *  }) // 'https://api.empathy.co/demo'
+ *
+ *  interpolate('https://search.{(api-)env(.)}empathy.co/{instance}', {
+ *    instance: 'demo'
+ *  }) // 'https://search.empathy.co/demo'
+ * ```
+ * @public
+ */
+export function interpolate(
```
LuisMartinez15, line 84: `interpolate`  sounds too generic, but it is somewhat coupled to an endpoint. Maybe adding the suffix `endpoint` or `Url` would make its intention clearer:

`interpolateEndpoint` or `interpolateUrl`

LuisMartinez15 (changes requested):
beautiful regex 🚬

packages/x-adapter/jest.config.js
javieri-empathy: It is really great to see that you are exploring alternative tools, and properly testing them, but TBH, I don't think we have any performance issue with jest. 974 tests and it takes 20s to execute in my laptop, I've seen yesterday 50s worst scenario in yours, and it went down up to 35s with tweaks. There is no much room of improvement there, just a couple of seconds. 

Introducing another compiler as esbuild or swc here would make us lose type checking, which is far more valuable for me than saving 20s of the build time. 

Also, we need to be careful with these compilers. This are alternative versions to the TS one, which means that they will probably output different code, or they might not even support al language features.

Every tool has its own pros and cons, and what I'm trying to say here is that if we introduce a new tool, we must make sure that we are maximising its pros, and minimising its cons. In this particular case I don't think it is worth to lose type checking and to risk losing future TypeScript features just to save a few seconds in test time. Indeed, we can currently speed up our tests by disabling type checking, which is really expensive by adding  the config below to `jest.config.js`, and I still wouldn't do it in this case.

```js
  globals: {
    'ts-jest': {
      isolatedModules: true
    }
  }
```

packages/x-adapter/src/utils/interpolate.ts
javieri-empathy: I prefer the first one because I don't have to repeat that weird code twice, but I don't like the name, and I can't think something better 🤦

packages/x-adapter/src/utils/interpolate.ts
LuisMartinez15: You're right. I couldn't see the wood for the trees.

tiborux (approved):

LuisMartinez15 (approved):
LGTM

PR #376, opened 2022-03-18
chore: make eslint run against `.vue` files

javieri-empathy:
ESLint does not lint files with `.vue` extension as it differs from the default `.js` one. You have to explicitly add them through the `--ext` flag. Another option could be refactoring our configurations so it uses the `overrides` property, but I'm not entirely sure this would work, but it will for sure require more work to do it.


package.json
```diff
@@ -5,8 +5,8 @@
     "postinstall": "lerna bootstrap --hoist && npm run build",
     "postci": "lerna bootstrap --hoist --ci && npm run build",
     "build": "lerna run build",
-    "lint": "eslint packages --cache",
-    "lint:fix": "eslint packages --fix",
+    "lint": "eslint packages --ext .ts,.tsx,.vue",
```
diegopf, line 8: is it intended to remove the `--cache` option?

diegopf (changes requested):

package.json
javieri-empathy: Glad you asked!. Yes, it is intended. We used to lint files before committing them with cache a while ago, but due to the amount of files linted this cache ended up freezing webstorm. I don't know if now it has been fixed or improved. In any case the command used to lint staged files is not this one, it is the one below at the `lint-staged` section of the `package.json`.

In any case this command is just run from the CI, and as we don't store caches (yet) in the CI we can skip that flag, it simply does not make sense to have it right now.

diegopf (approved):

PR #384, opened 2022-03-29
fix(base-currency): format preference goes prop to injected to default

CachedaCodes:
[EX-5649](https://searchbroker.atlassian.net/browse/EX-5649)

## Motivation and context
The `base-result-previous-price` was sending a format by default to the `base-currency` component, this format was taking priority over the injected format. Per the task, the priority order for the format to apply should be prop -> injected -> default. Given this, the decision was to move the default value of the format to `base-currency` and adjust the priorities.

## Type of change

- [X] `Main`
- [ ] Other. Specify:

## How has this been tested?
Add `base-result-previous-price` to a result in a view, the default format is applied.
Pass a format for the prop. This should be the format applied.
Inject a different format to that result, the format of the prop should still apply.
Remove the prop format, the injected format should apply.

### DIFF
--- a/packages/x-components/src/components/currency/__tests__/base-currency.spec.ts
+++ b/packages/x-components/src/components/currency/__tests__/base-currency.spec.ts
@@ -26,7 +26,7 @@ function renderInjectedBaseCurrency({ value, format }: RenderBaseCurrencyOptions
     {
       template: `
           <Provider>
-            <BaseCurrency v-bind="{ value, ...(format && { format }) }" />
+          <BaseCurrency v-bind="{ value, format }" />
           </Provider>
         `,
       components: {
--- a/packages/x-components/src/components/currency/__tests__/base-currency.spec.ts
+++ b/packages/x-components/src/components/currency/__tests__/base-currency.spec.ts
@@ -35,7 +35,7 @@ function renderInjectedBaseCurrency({ value, format }: RenderBaseCurrencyOptions
       }
     },
     {
-      data: function () {
+      data() {
         return {
           value,
           format
--- a/packages/x-components/src/components/currency/__tests__/base-currency.spec.ts
+++ b/packages/x-components/src/components/currency/__tests__/base-currency.spec.ts
@@ -13,7 +13,46 @@ function renderBaseCurrency({ value, format }: RenderBaseCurrencyOptions): Wrapp
   });
 }
 
+function renderInjectedBaseCurrency({ value, format }: RenderBaseCurrencyOptions): Wrapper<any> {
+  @Component({
+    template: '<div><slot /></div>'
+  })
+  class Provider extends Vue {
+    @XProvide('currencyFormat')
+    public providedFormat = '$i,iii.ddd';
+  }
+
+  return mount(
+    {
+      template: `
+          <Provider>
+            <BaseCurrency v-bind="{ value, ...(format && { format }) }" />
+          </Provider>
+        `,
+      components: {
+        Provider,
+        BaseCurrency
+      }
+    },
+    {
+      data: function () {
+        return {
+          value,
+          format
+        };
+      }
+    }
+  );
+}
+
 describe('testing BaseCurrency component', () => {
+  it('renders the default format', () => {
+    const wrapper = renderBaseCurrency({
+      value: 12345678
+    });
+    expect(wrapper.text()).toEqual('12.345.678,00');
+  });
+
   it('renders the provided format correctly with "?" after the decimal separator', () => {
     const wrapper = renderBaseCurrency({
       value: 12345678,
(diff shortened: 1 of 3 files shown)


packages/x-components/src/components/result/base-result-previous-price.vue
```diff
@@ -9,7 +9,7 @@
          @binding {result} result - Result data
     -->
     <slot :result="result">
-      <BaseCurrency :value="result.price.originalValue" :format="format" />
+      <BaseCurrency v-bind="{ value: result.price.originalValue, ...(format && { format }) }" />
```
CachedaCodes, line 12: Due to how `propsData` works I can't send anything as `format` prop if I don't want it to appear there.

packages/x-components/src/components/currency/base-currency.vue
```diff
@@ -83,13 +83,9 @@
...
-      return (
-        this.format ??
-        this.injectedFormat ??
-        //TODO: add here logger
-        //eslint-disable-next-line no-console
-        console.warn('It is necessary to pass a prop or inject the format')
-      );
+      return 'format' in this.$options.propsData!
+        ? this.format
+        : this.injectedFormat ?? this.format;
```
JaviIG, line 88: Small suggestion.

If you simply remove the prop default value and make it optional again, and do this change here:
```suggestion
      return this.format ?? this.injectedFormat ?? 'i.iii,dd'
```

Then you should be able to keep using this component easily without the boolean destructure trick.

```html
<BaseCurrency v-bind="{ value: result.price.originalValue, format }" />
```

JaviIG (changes requested):
GJ, looks great from a functional point of view.

packages/x-components/src/components/currency/base-currency.vue
```diff
@@ -64,8 +64,8 @@
      *
```
tiborux, line 64: Can we use this opportunity to update the documentation? I mean, update it to use the new style that we are using in the new components, like in the `my-history.vue`

packages/x-components/src/components/result/base-result-previous-price.vue
```diff
@@ -9,7 +9,7 @@
          @binding {result} result - Result data
```
tiborux, line 9: Same here for the documentation

tiborux (changes requested):

packages/x-components/src/components/currency/base-currency.vue
CachedaCodes: When I thought I was being smart but was being dumb instead. Good catch!

packages/x-components/src/components/currency/base-currency.vue
```diff
@@ -115,33 +109,272 @@ HTML element.
...
+```
+
+```vue
+<template>
+  <BaseCurrency :value="12345678.87654321" format="i iii - dd €" />
+  <!-- output: '12 345 678 - 87 €' -->
+</template>
+
+<script>
+  import { BaseCurrency } from '@empathyco/x-components/currency';
```
javieri-empathy, line 280:
```suggestion
  import { BaseCurrency } from '@empathyco/x-components';
```

packages/x-components/src/components/currency/base-currency.vue
```diff
@@ -115,33 +109,272 @@ HTML element.
...
+```
+
+```vue
+<template>
+  <BaseCurrency :value="12345678.87654321" format="i.iii,dddddd €" />
+  <!-- output: '12.345.678,876543 €' -->
+</template>
+
+<script>
+  import { BaseCurrency } from '@empathyco/x-components/currency';
```
javieri-empathy, line 298:
```suggestion
  import { BaseCurrency } from '@empathyco/x-components';
```

packages/x-components/src/components/currency/base-currency.vue
```diff
@@ -115,33 +109,272 @@ HTML element.
...
+```
+
+```vue
+<template>
+  <BaseCurrency :value="12345678.87654321" format="i.iii,dd €" />
+  <!-- output: '12.345.678,87 €' -->
+</template>
+
+<script>
+  import { BaseCurrency } from '@empathyco/x-components/currency';
```
javieri-empathy, line 352:
```suggestion
  import { BaseCurrency } from '@empathyco/x-components';
```

packages/x-components/src/components/currency/__tests__/base-currency.spec.ts
```diff
@@ -13,7 +13,46 @@ function renderBaseCurrency({ value, format }: RenderBaseCurrencyOptions): Wrapp
...
+  class Provider extends Vue {
+    @XProvide('currencyFormat')
+    public providedFormat = '$i,iii.ddd';
+  }
+
+  return mount(
+    {
+      template: `
+          <Provider>
+            <BaseCurrency v-bind="{ value, ...(format && { format }) }" />
```
javieri-empathy, line 29:
```suggestion
            <BaseCurrency v-bind="{ value, format }" />
```

javieri-empathy (approved):

tiborux (approved):

PR #386, opened 2022-03-29
feat(types): add `ExtractPaths`

LuisMartinez15:


packages/x-utils/src/types/utils.types.ts
```diff
@@ -100,3 +100,45 @@ export type PropertyType<
   : SomeObject extends any[]
   ? SomeObject[number]
   : never;
+
```
tajespasarela, line 103: would be better to split this big file in 2 smaller?

tajespasarela (changes requested):

packages/x-utils/src/types/utils.types.ts
LuisMartinez15: :sisi3: :callaito:

tajespasarela (changes requested):

packages/x-utils/src/types/property-path.types.ts
```diff
@@ -0,0 +1,106 @@
...
+ *
+ * @public
+ */
+export type ExtractPaths<SomeObject, Type, ExcludeOptional extends boolean = false> = {
+  [Path in PropertyPath<SomeObject> as PropertyType<SomeObject, Path> extends (
+    ExcludeOptional extends true ? Type : Type | undefined
+  )
+    ? Path
+    : never]: Type;
+};
```
JaviIG, line 106: Same questions like with the `TreeShakeObject` type. 
* How is this going to be used in the adapter?
* Can we remove the `ExcludeOptional` parameter
* I'm missing also a test for the optionals

packages/x-utils/src/types/__tests__/property-path.types.spec.ts
```diff
@@ -31,6 +31,18 @@ interface Person {
...
 
+interface AnotherExample {
+  stringKey: string;
+  optionalStringKey?: string;
+  numberKey: number;
+  booleanKey: boolean;
+  objectKey: {
+    innerStringKey: string;
+  };
+}
```
JaviIG, line 42: Move the interface near the test that uses it. I would do the same with the rest of the interfaces. Unless they are used in every test, it is better to locate them near the one that is using it. Easier to read and easier to maintain.

JaviIG (changes requested):
GJ. You can escape @javieri-empathy but you can't scape @JaviIG

packages/x-utils/src/types/object.types.ts
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * @remarks By default, ExcludeOptional is false, so optional properties are included.
+ *
+ * @param SomeObject - The object to tree-shake.
+ * @param Type - The type of the properties to preserve.
+ * @param ExcludeOptional - Flag to exclude optional properties.
+ *
+ * @public
+ */
+export type TreeShakeObject<SomeObject, Type, ExcludeOptional extends boolean = false> = {
```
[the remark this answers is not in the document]
LuisMartinez15: I used this name (`tree-shaking`) because it basically tree shakes an object, returning another one with the same shape but removing only those properties that don't match the passed type.

If you don't find it clear or sounds ambiguous, feel free to propose another one. We accept both cookies and better names.

packages/x-utils/src/types/utils.types.ts
LuisMartinez15: Yes, both have been refactored.

packages/x-utils/src/__tests__/utils.types.spec.ts
```diff
@@ -186,3 +196,54 @@ describe('PropertyType', () => {
...
+      objectKey: {
+        innerStringKey: 'inner'
+      },
+      stringKey: 'hi',
+      optionalStringKey: 'optional',
+      // @ts-expect-error
+      booleanKey: false
+    };
+
+    expect(test).toMatchObject<TreeShakeObjectByType<AnotherExample, string>>({
```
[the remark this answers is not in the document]
LuisMartinez15: That's right. The checks are the `@ts-expect-error` comments.

packages/x-utils/src/types/property-path.types.ts
LuisMartinez15: We are using a combination of both in the `x-adapter` to `narrow` the available `property paths` to assign to a `target` `key` in a `schema`. I forgot to add some examples in the documentation (both for this type and the `TreeShakeObject` one, so expect an update with some examples _soontm_)

packages/x-utils/.eslintrc.js
```diff
@@ -0,0 +1,19 @@
+module.exports = {
+  parserOptions: {
+    tsconfigRootDir: __dirname,
+    project: 'tsconfig.eslint.json'
+  },
+  overrides: [
```
LuisMartinez15, line 6: I've done this to avoid having to disable in each `.types.spec.ts` file the `@typescript-eslint/ban-ts-comment` rule to be able to use `@ts-expect-error` for _type-checking_

packages/x-utils/src/types/object.types.ts
```diff
@@ -0,0 +1,31 @@
+/**
+ * Picks the keys from the provided object that match the given type.
+ *
+ * @param SomeObject - The object to pick the keys from.
+ * @param Type - The type of the properties to pick.
+ *
+ * @example
```
LuisMartinez15, line 7: Added missing examples

packages/x-utils/src/types/object.types.ts
```diff
@@ -0,0 +1,31 @@
...
+ *   images: string[]
+ * }
+ *
+ * type ResultOnlyStrings = PickByType<Result, string>; // { id: string; price: { symbol: string; }}
+ * type ResultOnlyNumbers = PickByType<Result, number>; // { price: { max: number; min: number; }}
+ * ```
+ *
+ * @public
+ */
+export type PickByType<SomeObject, Type> = {
```
LuisMartinez15, line 25: Renamed `TreeShakeObject` to `PickByType`, which should be more straightforward to understand. Removed the `ExcludeOptionals` generic as @JaviIG suggested, as you can always pass the union of your desired `type` and `undefined` (i.e. `string | undefined`). Changed `object` to `Record<string, unknown>` as @JaviIG pointed out `functions` and `arrays` were being identified as `object` and therefore also being included.

packages/x-utils/src/types/property-path.types.ts
```diff
@@ -0,0 +1,116 @@
...
+  ? SomeObject[number]
+  : never;
+
+/**
+ * Extracts the property paths of the provided object that match the given type.
+ *
+ * @param SomeObject - The object to extract the property paths from.
+ * @param Type - The type of the property paths to extract.
+ *
+ * @example
```
LuisMartinez15, line 95: Added missing example

packages/x-utils/tsconfig.build.json
```diff
@@ -0,0 +1,15 @@
+{
```
LuisMartinez15, line 1: I've created this `tsconfig.build.json` file to be extended by both the `tsconfig.cjs.json` and `tsconfig.esm.json` to extract the common properties.

packages/x-utils/tsconfig.eslint.json
```diff
@@ -2,6 +2,5 @@
   "extends": "./tsconfig.json",
   "compilerOptions": {
     "noEmit": true
-  },
-  "exclude": []
+  }
```
LuisMartinez15, line 5: After removing this `exclude` property everything seems to be working as expected

packages/x-utils/tsconfig.json
```diff
@@ -2,18 +2,13 @@
...
     "esModuleInterop": true,
     "types": ["jest"],
     "lib": ["esnext"],
-    "sourceMap": true,
-    "inlineSources": true,
     "importHelpers": true
   },
   "include": ["src/**/*.ts"],
-  "exclude": ["node_modules", "**/__tests__/**/*"]
+  "exclude": ["node_modules"]
```
LuisMartinez15, line 13: I've removed the `tests` files from being `excluded` as this `tsconfig.json` is now the one used by the `IDE`/`compilation` process and, as such, needs to include those files.

LuisMartinez15:
I've refactored the `tsconfig` files so we have one for the `IDE`/`compilation` process, another one with the common config for every `build`, and one specific config per `build`.

LuisMartinez15:
Removed unnecessary `PickByType` as it is no longer required for the `x-adapter` schema types.

tajespasarela (changes requested):

tajespasarela (approved):

PR #403, opened 2022-04-03
feat(typeguards): add isFunction

LuisMartinez15:
EX-5796

### DIFF
--- a/packages/x-utils/src/__tests__/typeguards.spec.ts
+++ b/packages/x-utils/src/__tests__/typeguards.spec.ts
@@ -14,6 +14,8 @@ describe('typeguards test', () => {
       c: 'c'
     }
   };
+  const nil = null;
+  const undef = undefined;
 
   describe('isFunction', () => {
     it('should return true when the passed value is a Function', () => {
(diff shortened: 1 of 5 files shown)


packages/x-utils/src/__tests__/typeguards.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+  const func = (): any => {};
+  const arr = [1, 2, 3];
+  const obj = {
+    a: 1,
+    b: {
+      c: 'c'
+    }
+  };
+
+  describe('isFunction', () => {
```
tajespasarela, line 18: Should we test `null` and `undefined` values too?

packages/x-utils/src/__tests__/typeguards.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+      expect(isFunction(str)).toBe(false);
+      expect(isFunction(num)).toBe(false);
+      expect(isFunction(bool)).toBe(false);
+      expect(isFunction(func)).toBe(true);
+      expect(isFunction(arr)).toBe(false);
+      expect(isFunction(obj)).toBe(false);
+    });
+  });
+
+  describe('isObject', () => {
```
tajespasarela, line 29: Should we test null and undefined values too?

tajespasarela (changes requested):

packages/x-utils/src/__tests__/typeguards.spec.ts
LuisMartinez15: You're right, done!

tiborux (approved):

tajespasarela (approved):

PR #410, opened 2022-04-05
feat(tailwindcss-x): create tailwind plugin package

CachedaCodes:
[EX-5627](https://searchbroker.atlassian.net/browse/EX-5627)

## Motivation and context
Package for the Tailwind plugin with TS, PostCSS, and Vite. Added basic PostCSS plugins.

## Type of change

- [X] `Main`
- [ ] Other. Specify:


packages/tailwindcss-x/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "tailwindcss-x",
+  "private": true,
```
diegopf, line 3: I think we should add the description, author, license, homepage and similar fields like we do for other packages

packages/tailwindcss-x/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "version": "0.1.0",
+  "scripts": {
+    "dev": "vite",
+    "build": "tsc && vite build",
+    "preview": "vite preview"
+  },
+  "devDependencies": {
+    "autoprefixer": "^10.4.4",
```
diegopf, line 11: I see that we are using the version `~9.7.4` can we keep both projects in sync? This way we will not have a package-lock.json here.

packages/tailwindcss-x/package.json
```diff
@@ -0,0 +1,23 @@
...
+    "cssnano": "^5.0.15",
+    "postcss": "^8.4.12",
+    "postcss-import": "^14.1.0",
+    "tailwindcss": "^3.0.23",
+    "typescript": "~4.6.2",
+    "vite": "^2.9.0"
```
diegopf, line 17: I think we should use always the ~ and not the ^ modifier for external dependencies versions

diegopf (changes requested):
I see the readme.md file is missing, I think we should add one, shouldn't we?

.github/dependabot.yaml
```diff
@@ -19,3 +19,4 @@ updates:
       day: "friday"
       time: "08:00"
       timezone: "Europe/Madrid"
+
```
diegopf, line 22: Remove empty line, please.

packages/tailwindcss-x/tsconfig.eslint.json
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "noEmit": true
```
diegopf, line 4: We already have this set on the parent tsconfig.json, so we should skip it here.

diegopf (changes requested):

packages/tailwindcss-x/index.html
```diff
@@ -0,0 +1,13 @@
...
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <link rel="icon" type="image/svg+xml" href="favicon.svg" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <link rel="stylesheet" href="/src/css/index.pcss" />
+    <title>tailwindcss-x</title>
+  </head>
+  <body>
+    <div id="app"></div>
```
tajespasarela, line 11: add an example to see tw working:

```html
    <button class="x-bg-amber-400 x-p-1 x-border-amber-500 x-border-2 x-border-solid x-rounded">
      Button
    </button>
```

packages/tailwindcss-x/tsconfig.eslint.json
tajespasarela: actually, I think we are going to export types from this package so we have to remove that option from the parent.

packages/tailwindcss-x/index.html
```diff
@@ -0,0 +1,13 @@
...
+  </body>
```
tajespasarela, line 12: Include the script that already imports the CSS:
```html
    <script type="module" src="src/main.ts"></script>
```

packages/tailwindcss-x/index.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <link rel="icon" type="image/svg+xml" href="favicon.svg" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <link rel="stylesheet" href="/src/css/index.pcss" />
```
tajespasarela, line 7: No need this as it is imported from the script

packages/tailwindcss-x/tsconfig.json
```diff
@@ -0,0 +1,20 @@
...
+    "moduleResolution": "Node",
+    "strict": true,
+    "sourceMap": true,
+    "resolveJsonModule": true,
+    "esModuleInterop": true,
+    "noEmit": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "noImplicitReturns": true,
+    "skipLibCheck": true
```
tajespasarela, line 16: Include:
```js
    "importHelpers": true,
    "allowSyntheticDefaultImports": true,
    "declaration": true,
    "declarationMap": true,
```

and remove : 
```js
    "noEmit": true
```

tajespasarela (changes requested):

tajespasarela (dismissed):

packages/tailwindcss-x/tsconfig.json
```diff
@@ -0,0 +1,23 @@
...
+    "moduleResolution": "Node",
+    "strict": true,
+    "sourceMap": true,
+    "resolveJsonModule": true,
+    "esModuleInterop": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "noImplicitReturns": true,
+    "skipLibCheck": true,
+    "importHelpers": true,
```
diegopf, line 16: Enabling importHelpers, now we need to install tslib as dependency. See https://www.typescriptlang.org/tsconfig#importHelpers

diegopf (changes requested):

diegopf (approved):

tajespasarela (approved):

PR #429, opened 2022-04-14
Extract session service

bllucia:
Extract session service to x-utils (and fixing ttl bug)


packages/x-utils/src/services/session.service.ts
```diff
@@ -21,7 +25,9 @@ export class DefaultSessionService implements SessionService {
   public static instance: SessionService = new DefaultSessionService();
 
   public constructor(
-    protected storageService: StorageService = new BrowserStorageService(localStorage, 'x'),
+    protected storageService: StorageService = typeof localStorage !== 'undefined'
+      ? new BrowserStorageService(localStorage, 'x')
+      : new InMemoryStorageService(),
```
javieri-empathy, line 30: I'd like to remove this singleton and the default parameters so it is not coupled to the browser. The service instantiation should be done somewhere else in x-components. Initially the idea we had with services was to add them to the x-module. But for this PR I think it is fine. We can work this in the future.

javieri-empathy (dismissed):

packages/x-components/src/x-modules/tagging/wiring.ts
```diff
@@ -8,8 +9,7 @@ import { wireService, wireServiceWithoutPayload } from '../../wiring/wires.facto
 import { filter, mapWire } from '../../wiring/wires.operators';
 import { Wire } from '../../wiring/wiring.types';
 import { createWiring } from '../../wiring/wiring.utils';
-import { DefaultSessionService } from './service/session.service';
-import { DefaultPDPAddToCartService } from './service/index';
+import { DefaultPDPAddToCartService } from './service';
```
LuisMartinez15, line 12: Avoid importing from barrel file

```suggestion
import { DefaultPDPAddToCartService } from './service/pdp-add-to-cart.service';

```

tiborux (approved):

LuisMartinez15 (changes requested):
Good job!

tiborux (approved):

LuisMartinez15 (approved):

PR #443, opened 2022-04-20
refactor(schema): remove undefined from return type of mapper factory

diegopf:
Show a warning if a schema cannot be applied and remove union type with undefined from return type.

EX-5926

### DIFF
--- a/packages/x-adapter/src/endpoint-adapter.factory.ts
+++ b/packages/x-adapter/src/endpoint-adapter.factory.ts
@@ -37,10 +37,10 @@ export const endpointAdapterFactory: EndpointAdapterFactory = <Request, Response
     const endpoint = getEndpoint(rawEndpoint ?? requestEndpoint, request);
     const requestParameters = requestMapper(request, { endpoint });
 
-    return httpClient<Response>(
+    return httpClient(
       endpoint,
       deepMerge({}, defaultRequestOptions, requestOptions, { parameters: requestParameters })
-    ).then(response => responseMapper(response, { endpoint, requestParameters }));
+    ).then(response => responseMapper(response, { endpoint, requestParameters }) as Response);
   };
 
   endpointAdapter.extends = <NewRequest, NewResponse>(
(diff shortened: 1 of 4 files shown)


packages/x-adapter/src/endpoint-adapter.factory.ts
```diff
@@ -37,10 +37,10 @@ export const endpointAdapterFactory: EndpointAdapterFactory = <Request, Response
     const endpoint = getEndpoint(rawEndpoint ?? requestEndpoint, request);
     const requestParameters = requestMapper(request, { endpoint });
 
-    return httpClient<Response>(
+    return httpClient(
       endpoint,
       deepMerge({}, defaultRequestOptions, requestOptions, { parameters: requestParameters })
-    ).then(response => responseMapper(response, { endpoint, requestParameters }));
+    ).then(response => responseMapper(response, { endpoint, requestParameters }) as Response);
```
javieri-empathy, line 43: The issue we found yesterday comes from the type of the response mapper. As the `responseMapper` falls back to the `identityMapper`, the type is messed up. If you explicitly set the type of the destructured options this is fixed. I would fix it up there.

```ts
  const {
    endpoint: rawEndpoint,
    httpClient = fetchHttpClient,
    requestMapper = identityMapper,
    responseMapper = identityMapper,
    defaultRequestOptions = {}
  }: EndpointAdapterOptions<Request, Response> = options;
```

javieri-empathy (changes requested):

packages/x-adapter/src/endpoint-adapter.factory.ts
```diff
@@ -32,12 +32,12 @@ export const endpointAdapterFactory: EndpointAdapterFactory = <Request, Response
       requestMapper = identityMapper,
       responseMapper = identityMapper,
       defaultRequestOptions = {}
-    } = options;
+    }: EndpointAdapterOptions<Request, Response> = options;
 
     const endpoint = getEndpoint(rawEndpoint ?? requestEndpoint, request);
```
tajespasarela, line 37: The endpoints are in the wrong order. The request must override the options endpoint:

```suggestion
    const endpoint = getEndpoint(requestEndpoint ?? rawEndpoint, request);
```

tajespasarela (changes requested):

packages/x-adapter/src/endpoint-adapter.factory.ts
diegopf: Okay, I've also refactored a test that I think was misleading regarding this behaviour

tajespasarela (approved):

javieri-empathy (approved):

PR #447, opened 2022-04-22
build(deps): update eslint-plugin dependencies

tiborux:
Update eslint-plugin dependencies to the latest version. Change the versions in the package to be 'patch' instead of minor.

This task: https://searchbroker.atlassian.net/browse/EX-5964 has the different rules applied. We should check it in the future and remove them to fix the problems.

EX-5383


LuisMartinez15 (changes requested):
I think we should set the new rules to `warn` instead of entirely disabling them with `off`, so in the future we know we have to refactor our code to make them work properly. Disabling them is going to make it harder for us to know which ones we should be reactivating later.

packages/eslint-plugin-x/lib/configs/ts.js
```diff
@@ -15,10 +15,10 @@ const syntaxRules = {
   '@typescript-eslint/no-empty-interface': 'off',
   '@typescript-eslint/no-explicit-any': 'off',
   '@typescript-eslint/no-extra-parens': ['error', 'all', { nestedBinaryExpressions: false }],
+  '@typescript-eslint/no-loss-of-precision': 'off',
```
diegopf, line 18: IMO we can leave this rule enabled and just refactor the [filters test](https://github.com/empathyco/x/blob/main/packages/x-components/src/utils/__tests__/filters.spec.ts#L94) because at least I don't see a reason for such numbers there.

diegopf (changes requested):
Good job dealing with the huge amount of errors and warnings!👏

packages/eslint-plugin-x/lib/configs/jsdoc.js
```diff
@@ -6,26 +6,36 @@ module.exports = {
...
       'jsdoc/require-param-type': 'off',
       'jsdoc/require-returns': 'error',
       'jsdoc/require-returns-check': 'error',
       'jsdoc/require-returns-description': 'error',
       'jsdoc/require-returns-type': 'off',
+      'jsdoc/tag-lines': [
+        'error',
+        'any',
+        {
+          dropLines: 'true'
```
LuisMartinez15, line 36: Isn't this option `dropEndLines` ([docs](https://github.com/gajus/eslint-plugin-jsdoc#dropendlines-defaults-to-false))?

```suggestion
          dropEndLines: 'true'
```

packages/eslint-plugin-x/lib/configs/vue.js
```diff
@@ -29,9 +29,13 @@ module.exports = {
         }
       ],
       'vue/eqeqeq': 'error',
+      'vue/multi-word-component-names': 'warn',
       'vue/no-template-shadow': 'off',
       'vue/no-v-html': 'off',
-      'vue/v-on-function-call': 'error'
+      'vue/component-tags-order': 'warn',
+      'vue/v-on-function-call': 'error',
+      'vue/v-slot-style': 'off',
```
LuisMartinez15, line 37:
```suggestion
      'vue/v-slot-style': 'warn',
```

LuisMartinez15 (changes requested):

LuisMartinez15 (dismissed):
amazing job 👏🏿 👏🏿

diegopf (dismissed):

diegopf (approved):

PR #451, opened 2022-04-25
feat(theme): add custom default values

LuisMartinez15:
EX-5717

Add custom default values to the plugin's `theme`


diegopf (changes requested):
I see on figma that there are defined highlight colors but missing here. Do we need them?

LuisMartinez15:
@diegopf Those were added afterwards by Cristina and she didn't tell me 😞 

Thanks for pointing it out, tho. I'll include them.

diegopf (approved):

tiborux (approved):

PR #458, opened 2022-04-26
feat(utils): log request errors by default

javieri-empathy:
EX-5981

While collaborating with @joseacabaneros we spotted that the request errors were silenced, making it hard to fix the issue. 

This happens because in the `createFetchAndSaveActions` we are catching all errors to check if they are a cancelation instead of a real error. 

My proposal is to use the `console.error` as the default of the `onError` hook. This way if you catch the error you will silence it. Otherwise it will be logged in the console.

To test this, simply try throwing an error in the result mapper that we have configured.


LuisMartinez15 (approved):
good job! thanks!

diegopf (approved):

PR #460, opened 2022-04-26
fix: unnable to add breakpoints in vue components

tiborux:
We were investigating a little bit, we found this [workaround](https://github.com/vuejs/vue-jest/issues/448) but we should try to find another solution in the future (maybe we can try to find the problem in [vue-jest ](https://github.com/vuejs/vue-jest/tree/master/packages/vue2-jest)and create a pr)

The problem is related to the source maps, they are not linked with the source code.

EX-5325

### DIFF
--- a/packages/x-components/jest.config.js
+++ b/packages/x-components/jest.config.js
@@ -1,7 +1,7 @@
 module.exports = {
   preset: 'ts-jest',
   transform: {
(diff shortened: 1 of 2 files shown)


diegopf (approved):

diegopf:
works like a charm👏

PR #461, opened 2022-04-27
feat(search): extract `fetchAndSaveResponse` `onSuccess` callback to its own `saveSearchResponse` action

LuisMartinez15:
EX-5982

❗ ❗ I've renamed the `increasePageAppendingResults` to `increasePageAppendingResultsWire`. This should be a `BREAKING CHANGE` as it is exported since `version 2.x.x`

BREAKING-CHANGE: Rename `increasePageAppendingResults` to `increasePageAppendingResultsWire` in `search wiring`.

Motive will need to override the logic for saving the search response. Right now that logic is in the onSuccess hook of the fetchAndSaveSearchResponse action. 

My fastest proposal here would be to extract this logic to a new saveSearchResponse action in the search module. Then if you need to override it to save more data, you can create a new action, import the default saveSearchResponse, call it, and then save more entities.

## Type of change

## How has this been tested?

Added new unit test cases.

## Checklist:


packages/x-components/src/x-modules/search/store/__tests__/actions.spec.ts
```diff
@@ -230,6 +232,47 @@ describe('testing search module actions', () => {
...
+
+    // eslint-disable-next-line max-len
+    it('saves default values of optional or undefined response properties in the search state', () => {
+      store.dispatch('saveSearchResponse', {
+        ...searchResponseStub,
+        partialResults: undefined,
+        redirections: undefined,
+        banners: undefined,
+        promoteds: undefined,
+        spellcheck: undefined
```
diegopf, line 261: Looks like the tests are failing because these undefined, adding `as any` should fix them.

diegopf (changes requested):

packages/x-components/src/x-modules/search/store/__tests__/actions.spec.ts
LuisMartinez15: My bad. I had a `search-adapter` build locally which had a modified `SearchResponse` type version. It should be fine now, as that change got merged into `main` and I've updated the pull request with it.

diegopf (approved):

PR #470, opened 2022-05-03
feat: add alt-text accessibility

mavmaf:
## Motivation and context
_The required `alt` attribute specifies an alternate text for an image, if the image cannot be displayed.

The `alt` attribute provides alternative information for an image if a user for some reason cannot view it (because of slow connection, an error in the src attribute, or if the user uses a screen reader)._

The aim of this PR is to include an `alt` attribute in every `` tag in order to fulfill the alt-text rule for accessibility. 

## Type of change

- [x] `Main`
- [ ] Other. Specify:

## Checklist:


packages/x-components/.eslintrc.js
```diff
@@ -6,8 +6,8 @@ module.exports = {
   },
   rules: {
     'no-dupe-class-members': 'off',
-    '@typescript-eslint/no-unused-vars': 'off',
-    'vuejs-accessibility/alt-text': 'warn',
+    '@typescript-eslint/no-unused-vars-experimental': 'off',
+    'vuejs-accessibility/alt-text': 'error',
```
herrardo, line 10: Should we set this here, or should we set it in the eslint-plugin-x? 

I know this will make all the projects that use the plugin to update the alt images, but i am concerned about enabling this rule in every single project one by one.

packages/x-components/src/views/design-system/design-system-picture.vue
```diff
@@ -5,35 +5,35 @@
...
-        <img class="x-picture__image" :src="whiteBackgroundImageUrl" />
+        <img class="x-picture__image" :src="whiteBackgroundImageUrl" alt="random image" />
       </picture>
     </article>
 
     <article class="x-list x-list--wrap x-list--gap-06 x-list--align-start x-picture-show-case">
       <h2 class="x-title2">Fixed Ratio</h2>
       <picture class="x-picture x-picture--fixed-ratio">
-        <img class="x-picture__image" :src="whiteBackgroundImageUrl" />
+        <img class="x-picture__image" :src="whiteBackgroundImageUrl" alt="random image" />
```
herrardo, line 36: I know this is an internal view, but should we set a alt like:

```suggestion
        <img class="x-picture__image" :src="whiteBackgroundImageUrl" alt="Fixed Ratio Image" />
```

The same applies to the previous ones

herrardo (changes requested):

packages/x-components/.eslintrc.js
mavmaf: Actually `'vuejs-accessibility/alt-text': 'error',` should be removed from here as the recommended accessibility configuration (which is the same as listed here but with `'error'`s instead of `'warn'`s was already added to `packages/eslint-plugin-x/lib/configs/vue.js`, so what we are actually doing here is to avoid the build breaks due to code not (yet) fulfilling the rules

herrardo (approved):

mnavarroespinosa (approved):

PR #472, opened 2022-05-05
bugfix(tagging): track click event on add to cart click

LuisMartinez15:
EX-5898

We want to detect `addtocart` as an interaction of the user to improve `findability`, so we also need to track the `click` event when an add to cart button is clicked.

## Type of change

## How has this been tested?

Added new step in the add to cart tagging scenario.

## Checklist

### DIFF
--- a/packages/x-components/src/x-modules/tagging/wiring.ts
+++ b/packages/x-components/src/x-modules/tagging/wiring.ts
@@ -190,7 +190,8 @@ export const taggingWiring = createWiring({
     storeClickedResultWire
   },
   UserClickedResultAddToCart: {
-    trackAddToCartWire
+    trackAddToCartWire,
+    trackResultClickedWire
   },
   UserClickedPDPAddToCart: {
     trackAddToCartFromSessionStorage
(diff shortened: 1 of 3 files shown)


tiborux (dismissed):

herrardo (changes requested):
Seeing the two different verb tenses is making me crazy :

Could we change lines 35 and 62 from `has been` to `is` in `packages/x-components/tests/e2e/tagging/tagging.spec.ts`

Also, should we add a TODO for SERP AddToCartButton Scenario? Something like:

Scenario: X. Clicking a result AddToCart triggers both the result click and add to cart tagging.

packages/x-components/src/x-modules/tagging/wiring.ts
```diff
@@ -190,7 +190,8 @@ export const taggingWiring = createWiring({
     storeClickedResultWire
   },
   UserClickedResultAddToCart: {
-    trackAddToCartWire
+    trackAddToCartWire,
+    trackResultClickedWire
   },
   UserClickedPDPAddToCart: {
     trackAddToCartFromSessionStorage
```
diegopf, line 197: If we want to track the add2cart click on PDP we should also add the wire here

herrardo: Then i don't understand why the test about PDP tragging is passing...

herrardo (approved):

PR #478, opened 2022-05-10
feat: add preselected filters component

tiborux:
Description

This component adds the possibility to use preselectedFilters in x-components

Implementation

A component was created to emit an event where there is filters in the snippet config or in the component's prop. The snippet config has priority over the prop.

How to test it

Run the tests

- Add a component in the home.vue: (<PreselectedFilters />)

- Modify the base-config.ts to add some preselected filters like this ones: `filters: ['{!tag=brand_facet}brand_facet:"Lego"', '{!tag=age_facet}age_facet:"toddler"']`

EX-6023


packages/x-components/src/components/__tests__/preselected-filters.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+    jest.clearAllMocks();
+  });
+
+  it('does not emit the event when neither filters nor snippet config filters are provided', () => {
+    const { emit } = renderPreselectedFilters();
+
+    expect(emit).not.toHaveBeenCalled();
+  });
+  it('emits the event when filters are provided in the snippet config', () => {
+    const snippetFilters = [
```
CachedaCodes, line 39: Move this to a constant outside of the tests and use that instead of repeating the same filters in the other tests. Also, rename it to `filters` or something less specific to the snippet and then use it with `snippetFilters: filtersConst`

packages/x-components/src/components/__tests__/preselected-filters.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+      }
+    }
+  });
+
+  return {
+    emit
+  };
+}
+
+describe('testing Preselected filters component', () => {
```
CachedaCodes, line 28: Add a test for when the `filters` array is empty.

CachedaCodes (changes requested):

CachedaCodes:
Nice job! 👏

packages/x-components/src/components/preselected-filters.vue
```diff
@@ -0,0 +1,106 @@
...
+     * @internal
+     */
+    @Inject()
+    public snippetConfig?: SnippetConfig;
+
+    /**
+     * A list of preselected filters.
+     *
+     * @remarks Emits the {@link XEventsTypes.PreselectedFiltersProvided} when the
+     * component is rendered or the values changed.
```
javieri-empathy, line 28: This says that it emits the event whenever the value changes, but it is not true. This is done only once in the `mounted` event, without any watcher or XEmit decorator. 

This is a use case for dutti which wants to change this filters dynamically. Maybe it can be postponed until someone requests it. I don't know if Vodafone needs this.

javieri-empathy (changes requested):

packages/x-components/src/components/__tests__/preselected-filters.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  it('does not emit the event when neither filters nor snippet config filters are provided', () => {
+    const { emit } = renderPreselectedFilters();
+
+    expect(emit).not.toHaveBeenCalled();
+  });
+  it('emits the event when filters are provided in the snippet config', () => {
+    const snippetFilters = [
+      '{!tag=brand_facet}brand_facet:"Lego"',
+      '{!tag=age_facet}age_facet:"toddler"'
```
herrardo, line 41: Could you add how to pass this object in the [Snippet Configuration](https://github.com/empathyco/x/blob/main/packages/x-components/static-docs/build-search-ui/web-archetype-integration-guide.md)

I understand this is an extra param, there is a row in the table for <extra parameters>, we could add in the description cell an example as the warehouse one.

herrardo (changes requested):

packages/x-components/src/components/preselected-filters.vue
```diff
@@ -0,0 +1,107 @@
...
+  import { PreselectedFilters, baseSnippetConfig } from '@empathyco/x-components';
+  export default {
+    name: 'PreselectedFiltersDemo',
+    components: {
+      PreselectedFilters
+    },
+    provide: {
+      ...baseSnippetConfig,
+      filters: ['{!tag=brand_facet}brand_facet:"Lego"', '{!tag=age_facet}age_facet:"toddler"']
+    }
```
javieri-empathy, line 77: I think this would be more like this:

```suggestion
    provide: {
      snippetConfig: {
        filters: ['{!tag=brand_facet}brand_facet:"Lego"', '{!tag=age_facet}age_facet:"toddler"']
      }
    }
```

No need to do nothign with the `baseSnippetConfig`, and you have to write the provided key `snippetConfig`

packages/x-components/tests/e2e/preselected-filters.feature
```diff
@@ -0,0 +1,38 @@
...
+    And   a recommendations API with a known response
+    And   a results API with a known response
+
+  Scenario Outline: 1. Preselected filter are selected is there is no more filters in the url
+    Given   an application the "<filter>" filter preselected
+    When  start button is clicked
+    And  "<query>" is searched
+    Then  related results are displayed
+    And   url contains parameter "filter" with value "<filter>"
+    And filters "<filter>" are shown in the selected filters
```
javieri-empathy, line 17: I hate this code style, but format the steps in both scenarios so they are aligned.

javieri-empathy (changes requested):

CachedaCodes (approved):

javieri-empathy (approved):

javieri-empathy (dismissed):

CachedaCodes (approved):

LuisMartinez15 (approved):

PR #485, opened 2022-05-12
feat(facets): Add `FilterModify` by `FilterModelName`.

tajespasarela:
BREAKING CHANGE: The method `registerFilterModifier` from `FilterEntityFactory` renamed to `registerModifierByFacetId`.
EX-6160

The intention of this PR is to add another option to register `FilterEntityModifier`s by `FilterModelName`.
The current way to register by Facet id is kept but renamed.

The use case is when you don't know the facets, therefore the facets ids, but you need to make single selectable the Hierarchical Filters, for example. This is a real case for Motive.

### DIFF
--- a/packages/x-components/src/x-modules/facets/entities/filter-entity.factory.ts
+++ b/packages/x-components/src/x-modules/facets/entities/filter-entity.factory.ts
@@ -1,4 +1,4 @@
-import { Facet, Filter, isFacetFilter } from '@empathyco/x-types';
+import { Facet, FilterModelName, Filter, isFacetFilter } from '@empathyco/x-types';
 import { Store } from 'vuex';
 import { RootXStoreState } from '../../../store/store.types';
 import { EditableNumberRangeFilterEntity } from './editable-number-range-filter.entity';
@@ -36,7 +36,13 @@ export class FilterEntityFactory {
   /**
    * The registered modifiers grouped by the facetId to be applied.
    */
-  protected modifiers: Record<Facet['id'], FilterEntityModifier[]> = {};
+  protected modifiersByFacetId: Record<Facet['id'], FilterEntityModifier[]> = {};
+
+  /**
+   * The registered modifiers grouped by the filter `ModelName` to be applied.
+   */
+  protected modifiersByFilterModelName: Partial<Record<FilterModelName, FilterEntityModifier[]>> =
+    {};
 
   /**
    * Contains the instantiated entities for each facet.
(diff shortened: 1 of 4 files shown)


packages/x-components/src/x-modules/facets/entities/filter-entity.factory.ts
```diff
@@ -36,7 +36,13 @@ export class FilterEntityFactory {
...
    * The registered modifiers grouped by the facetId to be applied.
    */
-  protected modifiers: Record<Facet['id'], FilterEntityModifier[]> = {};
+  protected modifiersByFacetId: Record<Facet['id'], FilterEntityModifier[]> = {};
+
+  /**
+   * The registered modifiers grouped by the filter `ModelName` to be applied.
+   */
+  protected modifiersByFilterModelName: Partial<Record<FilterModelName, FilterEntityModifier[]>> =
+    {};
```
LuisMartinez15, line 45: Do we need two different records to store the same thing? I don't think keys will overlap, so we can have just one list:

```typescript
protected modifiers: Record<Facet['id'] | FilterModelName, FilterEntityModifier[]> = {};
```

Then, we can get rid of the new register method, and just use the previous one:

```typescript
/**
   * Registers a list of modifiers to be used with the {@link FilterEntity | FilterEntities} of a
   * particular facet.
   *
   * @param modifierKey - The facet id or filter model name whose Entities will be modified.
   * @param modifiers - The list of modifiers to be registered.
   */
  registerModifier(
    modifierKey: Facet['id'] | FilterModelName,
    ...modifiers: FilterEntityModifier[]
  ): void {
    this.updateModifiers(this.modifiers, modifierKey, modifiers);
  }
```

And in `creatingFilterEntity()`

```typescript
const modifiers = isFacetFilter(filter)
  ? this.modifiers[filter.facetId] ?? this.modifiers[filter.modelName] ?? []
  : [];
```

LuisMartinez15 (changes requested):
Good job!

javieri-empathy (approved):

packages/x-components/src/x-modules/facets/entities/filter-entity.factory.ts
javieri-empathy: I was thinking the same when I saw the PR. It might be slightly easier for debuging issues if we have two separate records, but code is slightly simpler with your proposal.
I don't see any big issue or advantage in any of the both ways. So do as you wish.

LuisMartinez15 (approved):

PR #491, opened 2022-05-16
fix(url): URL now requires a query to store rest of fields.

javieri-empathy:
EX-6134

This just prevents the url handler from storing any values when there is no query. This happens with the preselected filters PR.


LuisMartinez15 (approved):
good job!

packages/x-components/src/x-modules/url/components/__tests__/url-handler.spec.ts
```diff
@@ -21,6 +21,7 @@ Object.defineProperty(window, 'performance', {
 /**
  * Renders the {@link UrlHandler} component, exposing a basic API for testing.
  *
+ * @param root0
```
mnavarroespinosa, line 24:
```suggestion
 * @param options - The options to render the component with.
```

mnavarroespinosa (approved):
Looks good to me!

Issue #385, opened 2022-03-29
[FEATURE]: As a shopper, I want the search box to have an animated placeholder


herrardo:
### How can the project be improved?

To have an animated placeholder showing different messages to inspire the shopper to type different types of queries.

https://user-images.githubusercontent.com/4663897/160593353-64ff623c-af1f-4957-bab3-948efa45d839.mov

Even though in the video only one part of the string is animated, the idea is to animate the whole string to make things easier. So `find` won't be in a fixed position and will move with the rest of the string.

#### Hover scenario
Given a shopper who is going to type in the search box
when the user hovers the search box
then the placeholder is animated showing different messages
when the user clicks on the search box and the input get focused
then the placeholder disappears to start typing.

#### Normal scenario
Given a shopper who is going to type in the search box
then the placeholder is animated showing different messages
when the user clicks on the search box and the input get focused
then the placeholder disappears to start typing.

### Proposed solution

An idea could be to create a `SearchInputPlaceholder` component that can coexist with the [`SearchInput`](https://github.com/empathyco/x/blob/main/packages/x-components/src/x-modules/search-box/components/search-input.vue) component.

### Customizations supported

- The text of the messages have to be configurable
- The number of different texts is configurable
- The animation can happen on hover
- The animation can happen before the hover

### Additional information

This is the module where this task has to be done: https://github.com/empathyco/x/tree/main/packages/x-components/src/x-modules/search-box

[SearcInput documentation
](https://docs.empathy.co/develop-empathy-platform/ui-reference/components/search-box/x-components.search-input.html)

Have a look at the [Contributing Guidelines](https://github.com/empathyco/x/blob/main/.github/CONTRIBUTING.md) ;)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

jtomasdias00:
Hi @herrardo 

My colleague @brunorodrigues1999 and I are students at the masters program in software engineering at FEUP (https://paginas.fe.up.pt/~estudar/mesw/). As part of evaluating one of our curricular units, we need to choose an open-source project and fix an issue. We would like to know if we could help with this issue? Thanks!

herrardo:
Yes sure @jtomasdias00 and @BrunoRodrigues1999!

Have a look to the [Contrib Guidelines](https://github.com/empathyco/x/blob/main/.github/CONTRIBUTING.md) first

herrardo:
[...]

PR #503, opened 2022-05-22
Feature/385 search box animated placeholder

BrunoRodrigues1999:
**Purpose**
To have an animated placeholder showing different messages to inspire the shopper to type different types of queries.

## Motivation and context
The motivation for this issue is to bring a more welcoming experience to the user by providing an idea of the products being made available to the same.

A ```searchinputplaceholder``` component was made an the same can coexist with the already existant ```searchinput``` component.

A new emitter ```emitUserHoveringOverSearchBox``` was created for the hover on the ```searchinput``` component so that the animation can start on hover.
The animation stops when the user clicks on the input element or if the user stops hovering. Also if there is text in the placeholder, the animation won't trigger so that the text doesn't overlap the inputted text.

- [ ] Dependencies. If any, specify: None
- [ ] Open issue: **[issue] (https://github.com/empathyco/x/issues/385)

## Type of change
New feature 

Main

## How has this been tested?
Testing has been performed in the development server by extensive usage
We tried different test scenarios by changing the number of placeholders in the prop

## Checklist:

### DIFF
--- a/packages/x-components/src/design-system/components/search-placeholder/default.scss
+++ b/packages/x-components/src/design-system/components/search-placeholder/default.scss
@@ -1,0 +1,42 @@
+.form-group{
+    position: relative;
+}
+.placeholder {
+    position: absolute;
+    left: 10px;
+    top: 12px;
+    opacity: 0;
+}
+    @keyframes fadeIn {
+        0% {
+            opacity: 0;
+        }
+
+        50% {
+            opacity: 1;
+        }
+
+        100% {
+            opacity: 0;
+        }
+    }
+
+    @keyframes slideIn {
+        0% {
+            transform: translateY(-30%);
+        }
+
+        50% {
+            transform: translateY(0);
+        }
+
+        100% {
+            transform: translateY(30%);
+        }
+    }
+
+    .visible{
+        opacity: 1;
+        animation: fadeIn 2s, slideIn 2s;
+        animation-iteration-count: infinite;
+    }
--- a/packages/x-components/src/x-modules/search-box/components/search-input-placeholder.vue
+++ b/packages/x-components/src/x-modules/search-box/components/search-input-placeholder.vue
@@ -1,0 +1,54 @@
+<template>
+  <label ref="placeholder" class="placeholder"></label>
+</template>
+
+<script lang="ts">
+  import Vue from 'vue';
+  import { Component, Prop } from 'vue-property-decorator';
+  import { State } from '../../../components/decorators/store.decorators';
+  import { XOn } from '../../../components/decorators/bus.decorators';
+  import { xComponentMixin } from '../../../components/x-component.mixin';
+  import { searchBoxXModule } from '../x-module';
+  import SearchInput from './search-input.vue';
+  @Component({
+    components: { SearchInput },
+    mixins: [xComponentMixin(searchBoxXModule)]
+  })
+  export default class SearchInputPlaceholder extends Vue {
+    @State('searchBox', 'query')
+    public query!: string;
+
+    @Prop({ default: () => [''] })
+    protected placeholdersArray!: string[];
+
+    @Prop({ default: 1 })
+    protected numberOfStrings!: number;
+
+    protected get isQueryEmpty(): boolean {
+      return this.query.length === 0;
+    }
+
+    @XOn(['UserHoveringOverSearchBox'])
+    animatePlaceholder(): void {
+      if (this.isQueryEmpty) {
+        var array = this.placeholdersArray;
+        var element = this.$refs.placeholder as HTMLElement;
+        element.classList.add('visible');
+        for (let i = 0; i < array.length; i++) {
+          setTimeout(function () {
+            element.innerText = array[i];
+            if (i === array.length - 1) {
+              element.classList.remove('visible');
+            }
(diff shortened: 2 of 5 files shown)


packages/x-components/src/x-modules/search-box/components/search-input-placeholder.vue
```diff
@@ -0,0 +1,51 @@
...
+  import { xComponentMixin } from '../../../components/x-component.mixin';
+  import { searchBoxXModule } from '../x-module';
+  import SearchInput from './search-input.vue';
+  @Component({
+    components: { SearchInput },
+    mixins: [xComponentMixin(searchBoxXModule)]
+  })
+  export default class SearchInputPlaceholder extends Vue {
+    @State('searchBox', 'query')
+    public query!: string;
```
diegopf, line 19: You can get access to the current query through the $x object.  You can see what's available [here](https://github.com/empathyco/x/blob/main/packages/x-components/src/plugins/x-bus.types.ts)

So you don't need to listen to the state directly

packages/x-components/src/x-modules/search-box/components/search-input-placeholder.vue
```diff
@@ -0,0 +1,51 @@
...
+
+    @Prop({ default: () => [''] })
+    protected placeholdersArray!: string[];
+
+    protected get isQueryEmpty(): boolean {
+      return this.query.length === 0;
+    }
+
+    @XOn(['UserHoveringOverSearchBox'])
```
diegopf, line 28: I think there's no need to create new events to handle this. You can just listen to the mouse events on this component.

packages/x-components/src/x-modules/search-box/components/search-input-placeholder.vue
```diff
@@ -0,0 +1,51 @@
...
+        for (let i = 0; i < array.length; i++) {
+          setTimeout(function () {
+            element.innerText = array[i];
+            if (i === array.length - 1) {
+              element.classList.remove('visible');
+            }
+          }, i * 2000);
+        }
+      }
+    }
```
diegopf, line 43: Currently the animation skips the last value and is not running on a loop 😅. Instead of setTimeout you should use setInterval and clear the interval when the input is not hovered. 

In order to loop through the array elements you can store the active element index, check if the next one exists and if not then set the active index to 0. This way the elements will be changed in an infinite loop. 

Take this just as an idea:

```
let activeIndex = 0;

let activePlaceholder = 0;

setInterval(() => {
const next = activeIndex++;
activePlaceholder = placeholders[next] ? next : 0;
})
```

packages/x-components/src/design-system/components/search-placeholder/default.scss
```diff
@@ -0,0 +1,42 @@
+.form-group{
```
diegopf, line 1: The styles does not need to be here. They can be in the component styles.

packages/x-components/src/x-modules/search-box/components/search-input-placeholder.vue
```diff
@@ -0,0 +1,51 @@
...
+        var array = this.placeholdersArray;
+        var element = this.$refs.placeholder as HTMLElement;
+        element.classList.add('visible');
+        for (let i = 0; i < array.length; i++) {
+          setTimeout(function () {
+            element.innerText = array[i];
+            if (i === array.length - 1) {
+              element.classList.remove('visible');
+            }
+          }, i * 2000);
```
diegopf, line 40: Might be a good idea to use a prop to control the interval time.

packages/x-components/src/x-modules/search-box/events.types.ts
```diff
@@ -7,57 +7,61 @@
...
    */
   UserFocusedSearchBox: void;
   /**
    * The user is typing/pasting a query
-   * * Payload: the partial query that the user is typing.
+   * Payload: the partial query that the user is typing.
    */
+  UserHoveringOverSearchBox: void;
+
+  UserHoveringOutOfSearchBox: void;
```
diegopf, line 44: As said before I think there's no need for new events🙂

packages/x-components/src/x-modules/search-box/components/search-input.vue
```diff
@@ -5,6 +5,8 @@
     @click="emitUserClickedSearchBox"
     @focus="emitUserFocusedSearchBox"
     @input="emitUserIsTypingAQueryEvents"
+    @mouseover="emitUserHoveringOverSearchBox"
+    @mouseout="emitUserHoveringOutOfSearchBox"
```
diegopf, line 9: I think we should listen for the mouse events on the label inside of the `search-input-placeholder.vue` and not here.

diegopf (changes requested):
Thanks for the contribution! 🚀 

I gave you a few hints to cover all the requirements. We really appreciate your effort. Please, take a look to the comments and feel free to ask anything.

herrardo:
@BrunoRodrigues1521 @BrunoRodrigues1999 @jtomasdias00  This PR is waiting for your update, could you have a look?

herrardo (changes requested):
Please, if you don't check the review From Diego we will proceed to decline the Pull Request.

PR #522, opened 2022-06-08
feat(mappers): add `combineMappers` util

mnavarroespinosa:
EX-6294

### DIFF
--- a/packages/x-adapter/src/mappers/__tests__/combine-mappers.spec.ts
+++ b/packages/x-adapter/src/mappers/__tests__/combine-mappers.spec.ts
@@ -2,7 +2,7 @@ import { DeepPartial } from '@empathyco/x-utils';
 import { Mapper } from '../../types/mapper.types';
 import { combineMappers } from '../combine-mappers';
 
-describe('combineMappers tests', () => {
+describe('combineMappers util', () => {
   const testMapper: Mapper<TestFrom, DeepPartial<TestTo>> = ({ query, count }) => {
     return {
       q: query,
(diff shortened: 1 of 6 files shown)


packages/x-adapter/src/mappers/__tests__/combine-mappers.spec.ts
```diff
@@ -0,0 +1,46 @@
+import { DeepPartial } from '@empathyco/x-utils';
+import { Mapper } from '../../types/mapper.types';
+import { combineMappers } from '../combine-mappers';
+
+describe('combineMappers tests', () => {
```
tajespasarela, line 5: `'combineMappers util'` we already know this is a test

packages/x-adapter/src/mappers/combine-mappers.ts
```diff
@@ -0,0 +1,29 @@
...
+  ...mappers: Mapper<From, DeepPartial<To>>[]
+): Mapper<From, To> {
+  return mappers.length === 1
+    ? (mappers[0] as Mapper<From, To>)
+    : (from, context) => {
+        if (!context.to) {
+          context.to = {};
+        }
+
+        return mappers.reduce<DeepPartial<To>>((result, mapper) => {
```
tajespasarela, line 23: `result` may be confusing (we have a model with the same name). Maybe change to:
`value` and `mappedValue`

packages/x-adapter/src/mappers/combine-mappers.ts
```diff
@@ -0,0 +1,29 @@
...
+ * @public
+ */
+export function combineMappers<From, To>(
+  ...mappers: Mapper<From, DeepPartial<To>>[]
+): Mapper<From, To> {
+  return mappers.length === 1
+    ? (mappers[0] as Mapper<From, To>)
+    : (from, context) => {
+        if (!context.to) {
+          context.to = {};
```
tajespasarela, line 20: I understand the use of this `to` parameter in the context but the name is confusing.

Maybe something more clear like: `previousMappedValue`,  `previousValue`, `preValue`...

packages/x-adapter/src/mappers/__tests__/combine-mappers.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+  });
+
+  it('combines multiple mappers', () => {
+    const combinedMappers = combineMappers<TestFrom, TestTo>(testMapper, extraMapper);
+    expect(combinedMappers({ query: 'patata', count: 0, extra: 'croqueta' }, {})).toEqual({
+      q: 'patata',
+      c: 0,
+      e: 'croqueta'
+    });
+  });
```
tajespasarela, line 33: Should we test the context parameter too?

tajespasarela (changes requested):

packages/x-adapter/src/mappers/combine-mappers.ts
mnavarroespinosa: We used the `to` parameter because it is defined in the `MapperContext` interface (packages/x-adapter/src/types/mapper.types.ts) and made sense to us to use it because of its documentation. If you think that another name fits better we can either rename it or just use a new name as the interface allows it.

tajespasarela: Yes, we can rename it in the `MapperContext`

packages/x-adapter/src/mappers/__tests__/combine-mappers.spec.ts
mnavarroespinosa: Sure, I've changed it!

packages/x-adapter/src/mappers/__tests__/combine-mappers.spec.ts
mnavarroespinosa: Good point, I've updated the tests.

packages/x-adapter/src/mappers/combine-mappers.ts
LuisMartinez15: Renamed from `to` to `mappedValue`

tajespasarela (approved):

PR #526, opened 2022-06-09
fix: Fix Request types.

tajespasarela:
EX-6302

### DIFF
--- a/packages/search-types/src/request/request.model.ts
+++ b/packages/search-types/src/request/request.model.ts
@@ -53,6 +53,6 @@ export interface TrackableRequest {
  */
 export interface ExtraParamsRequest {
   extraParams: {
-    [key: string]: any;
+    [key: string]: unknown;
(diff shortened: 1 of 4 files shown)


packages/search-types/src/request/request.model.ts
```diff
@@ -52,5 +52,7 @@ export interface TrackableRequest {
  * @public
  */
 export interface ExtraParamsRequest {
-  [key: string]: any;
+  extraParams: {
+    [key: string]: any;
```
LuisMartinez15, line 56: I think @javieri-empathy suggested to revert this `any` to `unknown`, but he can correct me if I'm wrong.

tajespasarela: changed

mnavarroespinosa (approved):

tiborux (approved):

PR #540, opened 2022-06-13
refactor: search in platform adapter

mnavarroespinosa:
EX-6313


packages/x-adapter-platform/src/types/platform-adapter.types.ts
```diff
@@ -6,15 +6,12 @@ import {
...
 } from '@empathyco/x-types';
-import { BaseRequest, SearchRequest, TaggingRequest } from './request.types';
-import {
-  NextQueriesResponse,
-  RelatedTagsResponse,
-  SearchResponse,
-  TopClickedResponse
-} from './response.types';
+import { BaseRequest, TaggingRequest } from './request.types';
+import { NextQueriesResponse, RelatedTagsResponse, TopClickedResponse } from './response.types';
```
tiborux, line 14: NextQueriesResponse  and TopClickedResponse should not be imported from here, right?

packages/x-adapter-platform/src/types/request.types.ts
```diff
@@ -1,6 +1,3 @@
-import { Filter } from '@empathyco/x-types';
-import { Dictionary } from '@empathyco/x-utils';
-
 export interface BaseRequest {
```
tiborux, line 1: should we move this to a base-request file?

tiborux (changes requested):

packages/x-adapter-platform/src/types/platform-adapter.types.ts
tiborux: I think that this change is made on the abraham's PR

mnavarroespinosa: Yes, you're right, it should not be imported from there and that change is made in Abraham's PR.

packages/x-adapter-platform/src/schemas/filters/simple-filter.schema.ts
```diff
@@ -1,14 +1,12 @@
 import { createMutableSchema, Schema } from '@empathyco/x-adapter-next';
 import { SimpleFilter } from '@empathyco/x-types';
-import { PlatformFacetFilter } from '../../types';
+import { PlatformFilter } from '../../types/models/facet.model';
 
-export const simpleFilterSchema: Schema<PlatformFacetFilter, SimpleFilter> = {
+export const simpleMutableFilterSchema = createMutableSchema<Schema<PlatformFilter, SimpleFilter>>({
```
diegopf, line 5:
```suggestion
export const simpleFilterMutableSchema = createMutableSchema<Schema<PlatformFilter, SimpleFilter>>({
```

packages/x-adapter-platform/src/schemas/models/facet.schema.ts
```diff
@@ -7,14 +7,16 @@ import {
   NumberRangeFacet,
   SimpleFacet
 } from '@empathyco/x-types';
-import { PlatformFacet, PlatformHierarchicalFilter } from '../types/models.types';
-import { numberFilterMutableSchema } from './filters/number-filter.schema';
-import { simpleMutableFilterSchema } from './filters/simple-filter.schema';
+import { numberFilterMutableSchema } from '../filters/number-filter.schema';
+import { simpleMutableFilterSchema } from '../filters/simple-filter.schema';
```
diegopf, line 11:
```suggestion
import { simpleFilterMutableSchema } from '../filters/simple-filter.schema';
```

packages/x-adapter-platform/src/types/request.types.ts
diegopf: will be used when we finish this refactor? I thought we are getting rid of this BaseRequest after all.

packages/x-adapter-platform/src/types/requests/index.ts
```diff
@@ -1,3 +1,4 @@
-export * from './popular-searches-request.model';
-export * from './query-suggestions-request.model';
 export * from './identifier-results-request.model';
+export * from './query-suggestions-request.model';
+export * from './popular-searches-request.model';
+export * from './search-request.model';
```
diegopf, line 4: Next queries is missing here but I think that is solved on main already.

diegopf (changes requested):

packages/x-adapter-platform/src/schemas/models/facet.schema.ts
mnavarroespinosa: Changed! Thanks!

packages/x-adapter-platform/src/types/requests/index.ts
mnavarroespinosa: Updated with main and now Next queries are there.

tiborux (approved):

diegopf (approved):

PR #544, opened 2022-06-14
feat(identifier-results): use platform adapter

tiborux:
EX-6316


packages/x-components/src/adapter/e2e-adapter.ts
```diff
@@ -0,0 +1,96 @@
...
+    ...properties
+  }).then(response => response.json());
+};
+
+/**
+ * Mock beacon httpClient.
+ *
+ * @param endpoint - The endpoint to use.
+ * @param options - Additional options to make the request with.
+ * @param _
```
mnavarroespinosa, line 45:
```suggestion
 * @param _
 * @param options - Additional options to make the request with.
```

mnavarroespinosa (approved):
gj!

PR #549, opened 2022-06-15
feat(popular-searches): use platform adapter

mnavarroespinosa:
EX-6327

### DIFF
--- a/packages/x-components/src/__tests__/utils.ts
+++ b/packages/x-components/src/__tests__/utils.ts
@@ -5,7 +5,7 @@ import { createLocalVue } from '@vue/test-utils';
 import Vue from 'vue';
 import { Store } from 'vuex';
 import { PlatformAdapter } from '@empathyco/x-adapter-platform';
-import { IdentifierResultsResponse } from '@empathyco/x-types';
+import { IdentifierResultsResponse, PopularSearchesResponse } from '@empathyco/x-types';
 import { XPluginOptions } from '../plugins';
 import { BaseXBus } from '../plugins/x-bus';
 import { XPlugin } from '../plugins/x-plugin';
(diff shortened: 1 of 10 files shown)


packages/x-components/src/__tests__/utils.ts
```diff
@@ -174,7 +176,9 @@ export function getMockedPlatformAdapter(
     /* eslint-disable @typescript-eslint/no-non-null-asserted-optional-chain */
     identifierResults: getMockedAdapterFunction(responseFeatures?.identifierResults!),
     nextQueries: getMockedAdapterFunction(responseFeatures?.nextQueries!),
+    popularSearches: getMockedAdapterFunction(responseFeatures?.popularSearches!),
     querySuggestions: getMockedAdapterFunction(responseFeatures?.querySuggestions!)
+    /* eslint-enable @typescript-eslint/no-non-null-asserted-optional-chain */
```
tiborux, line 181: remove this line

tiborux (approved):

packages/x-components/src/__tests__/utils.ts
mnavarroespinosa: By removing it, the following lines will have the `@typescript-eslint/no-non-null-asserted-optional-chain` disabled because this rule has been disabled using `eslint-disable` instead `eslint-disable-next-line`.

packages/x-components/src/x-modules/popular-searches/store/__tests__/getters.spec.ts
```diff
@@ -5,7 +5,7 @@ import Vue from 'vue';
 import Vuex, { Store } from 'vuex';
```
ajperezbau, line 5: I think that in this file the interface `PopularSearchesRequest` should be used instead of `SuggestionsRequest`.

ajperezbau (changes requested):

ajperezbau (approved):

PR #563, opened 2022-06-20
docs(api-extractor): add API extractor to `x-adapter-platform`

mnavarroespinosa:
[EX-6384](https://searchbroker.atlassian.net/browse/EX-6384)

### DIFF
--- a/packages/x-adapter-platform/package.json
+++ b/packages/x-adapter-platform/package.json
@@ -42,7 +42,7 @@
   },
   "devDependencies": {
     "@microsoft/api-documenter": "~7.15.3",
-    "@microsoft/api-extractor": "~7.19.4",
+    "@microsoft/api-extractor": "~7.20.0",
     "@types/jest": "~27.0.3",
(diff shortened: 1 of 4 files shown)


tiborux (approved):

packages/x-adapter-platform/package.json
```diff
@@ -38,6 +41,8 @@
     "tslib": "~2.3.0"
   },
   "devDependencies": {
+    "@microsoft/api-documenter": "~7.15.3",
+    "@microsoft/api-extractor": "~7.19.4",
```
diegopf, line 45: I would use the same version as x-components package, which is the 7.20.0

diegopf (changes requested):

packages/x-adapter-platform/package.json
mnavarroespinosa: Good point! I've upgraded it.

diegopf (approved):

tiborux (approved):

PR #575, opened 2022-06-23
fix: Remove `@empathyco/x-platform-adapter` dependency from library c…

javieri-empathy:
- Removes `Adapter` type. This has no sense and it was complicated.
- Adds `XComponentsAdapter` type to `@empathyco/x-types`. Together with the rest of the request/responses info. `x-types` can be considered as a package where all the shared types needed mainly by x-components, and other packages are.
- Moves `getTaggingInfoFromUrl` to `@empathyco/x-platform-adapter`. This utility was previously in `@empathyco/x-utils`. Due to the refactor of the types it created a cyclic dependency.
- Makes `PlatformAdapter` extendable.

### DIFF
--- a/packages/x-components/src/plugins/x-plugin.ts
+++ b/packages/x-components/src/plugins/x-plugin.ts
@@ -413,8 +413,10 @@ export class XPlugin implements PluginObject<XPluginOptions> {
  * {@link XComponentAPI | X Component API }.
  *
  * @example
- * Minimal installation example. A search adapter is needed for the plugin to work, and connect to
- * the API.
+ * Minimal installation example. An API adapter is needed to connect the X Components with the
+ * suggestions, search, or tagging APIs. In this example we are using the default Empathy's platform
+ * adapter.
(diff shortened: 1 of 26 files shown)


packages/x-components/src/plugins/x-plugin.ts
```diff
@@ -24,15 +24,15 @@ import { assertXPluginOptionsAreValid } from './x-plugin.utils';
  */
 export class XPlugin implements PluginObject<XPluginOptions> {
   /**
-   * {@link @empathyco/x-adapter-platform#PlatformAdapter | PlatformAdapter} Is the middleware
+   * {@link @empathyco/x-typesm#XComponentsAdapter | XComponentsAdapter} Is the middleware
```
tajespasarela, line 27: Update the doc `@example` in line 419

tajespasarela (changes requested):

tajespasarela (approved):

PR #578, opened 2022-06-27
chore: Change schemas naming.

tajespasarela:
EX-6435

This PR does the following changes in `@empathyco/x-adapter-platform`:
1. Renaming all the schemas from `XXXXMutableSchema` to `XXXXSchema`. All the schemas here are mutable by default. No need to reiterate in naming
2. Moving `hierarchicalFilterSchema` to its own file like the others filters schemas.

### DIFF
--- a/packages/x-adapter-platform/src/schemas/models/facet.schema.ts
+++ b/packages/x-adapter-platform/src/schemas/models/facet.schema.ts
@@ -2,17 +2,17 @@ import { createMutableSchema, Schema } from '@empathyco/x-adapter';
 import {
   EditableNumberRangeFacet,
   HierarchicalFacet,
-  HierarchicalFilter,
   NumberRangeFacet,
   SimpleFacet
 } from '@empathyco/x-types';
-import { numberFilterMutableSchema } from '../filters/number-filter.schema';
-import { simpleFilterMutableSchema } from '../filters/simple-filter.schema';
-import { PlatformFacet, PlatformHierarchicalFilter } from '../../types/models/facet.model';
+import { PlatformFacet } from '../../types/models/facet.model';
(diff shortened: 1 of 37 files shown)


packages/x-adapter-platform/src/schemas/models/facet.schema.ts
```diff
@@ -2,17 +2,17 @@ import { createMutableSchema, Schema } from '@empathyco/x-adapter';
...
 } from '@empathyco/x-types';
-import { numberFilterMutableSchema } from '../filters/number-filter.schema';
-import { simpleFilterMutableSchema } from '../filters/simple-filter.schema';
-import { PlatformFacet, PlatformHierarchicalFilter } from '../../types/models/facet.model';
+import { PlatformFacet } from '../../types/models/facet.model';
 import { getFacetConfig } from '../facets/utils';
 import { FacetsConfig } from '../facets/types';
+import { hierarchicalFilterSchema } from './filters/hierarchical-filter.schema';
+import { numberFilterSchema } from './filters/number-filter.schema';
+import { simpleFilterSchema } from './filters/simple-filter.schema';
```
CachedaCodes, line 13: Could these 3 imports be bundled together?

```suggestion
import { hierarchicalFilterSchema, numberFilterSchema, simpleFilterSchema } from './filters';
```

CachedaCodes (changes requested):

packages/x-adapter-platform/src/schemas/models/facet.schema.ts
diegopf: but usually we don't want to import from barrels, do we?

diegopf (approved):

CachedaCodes (approved):

PR #592, opened 2022-07-07
docs: Add module type to integration script docs

javieri-empathy:
Adds the type="module" to the x archetype scripts so they can be integrated by the clients.


herrardo (approved):
😘

PR #593, opened 2022-07-07
feat(history-queries): add `HistoryQueriesSwitch` component

tiborux:
If you wanna test it, you can add in the predictive layer this code:

```
<ToggleHistoryQueries />
<MyHistory />
```

Right now, when you disable the history queries using this button, nothing gonna happen, because we are emitting an event that it's going to be used by the new dialog to confirm if you wanna disable or not the history queries.

If you wanna test it anyways, you can modify the toggle method by this:

```
    protected toggle(): void {
      if (this.isEnabled) {
        this.$x.emit('UserToggledHistoryQueries', false);
        this.$x.emit('UserDisableHistoryQueries');
      } else {
        this.$x.emit('UserToggledHistoryQueries', true);
      }
    }
```

EX-6515


packages/x-components/src/x-modules/history-queries/components/toggle-history-queries.vue
```diff
@@ -0,0 +1,78 @@
...
+- [`UserDisableHistoryQueries`](./x-components.historyqueriesxevents.userdisablehistoryqueries.md):
+  the event is emitted after the user clicks the toggle and `isEnabled` is `true`.
+- [`UserToggledHistoryQueries`[(x-components.historyqueriesxevents.usertoggledhistoryqueries.md):
+  the event is emitted after the user clicks the toggle and `isEnabled` is `false`.
+
+## See it in action
+
+Here you have a basic example of how the toggle is rendered.
+
+_Try clicking it to see how it changes its state_
```
diegopf, line 60: I'm not sure how the vue live works but I would say this will not work since the `ToggleHistoryQueries`  is only emitting the `UserDisableHistoryQueries` and there is nothing that will actually disable the history queries. I think the switch state will stay true regardless you click on it.

packages/x-components/src/x-modules/history-queries/components/toggle-history-queries.vue
```diff
@@ -0,0 +1,78 @@
...
+    @State('historyQueries', 'isEnabled')
+    public isEnabled!: boolean;
+
+    /**
+     * Emit events depending on the switch value.
+     *
+     * @internal
+     */
+    protected toggle(): void {
+      if (this.isEnabled) {
```
javieri-empathy, line 37: I would always emit the `UserToggledHistoryQueries` event, which I proposed to rename to `UserClickedToggleHistoryQueries` to better reflect the intention.

Then we can modify the wiring using a `filter` function so it only listens to the events with payload `true`, and maybe we can get rid of the trick of `UserDisableHistoryQueries` event. But I'm not sure about what we talked the other day in the refinement. I thought we wrote the new proposal in the task but I couldn't find it.

packages/x-components/src/x-modules/history-queries/components/toggle-history-queries.vue
```diff
@@ -0,0 +1,78 @@
...
+   * Toggle history queries component to enable or disable the history queries.
+   * This component emits events depending on the `isEnabled` value.
+   *
+   * @public
+   */
+  @Component({
+    mixins: [xComponentMixin(historyQueriesXModule)],
+    components: { BaseSwitch }
+  })
+  export default class ToggleHistoryQueries extends Vue {
```
javieri-empathy, line 22: Should we name it `HistoryQueriesSwitch`? Just to match the control that is being rendered

packages/x-components/src/x-modules/history-queries/components/toggle-history-queries.vue
```diff
@@ -0,0 +1,78 @@
...
+  import { ToggleHistoryQueries } from '@empathyco/x-components';
+
+  export default {
+    name: 'ToggleHistoryQueriesDemo',
+    components: {
+      ToggleHistoryQueries
+    }
+  };
+</script>
+```
```
javieri-empathy, line 77: I would add a slightly more complex example so it really shows how it works.

```html
```vue live
<template>
  <div>
    <div>
      <SearchInput :instant="false" />
      <SearchButton>Search</SearchButton>
    </div>
    <label>
      History queries:
      <ToggleHistoryQueries />
      <HistoryQueries />
    </label>
  </div>
</template>

<script>
  import { ToggleHistoryQueries, HistoryQueries } from '@empathyco/x-components/history-queries';
  import { SearchInput, SearchButton } from '@empathyco/x-components/search';

  export default {
    name: 'ToggleHistoryQueriesDemo',
    components: {
      ToggleHistoryQueries,
      HistoryQueries,
      SearchInput,
      SearchButton
    }
  };
</script>
```

packages/x-components/src/x-modules/history-queries/components/toggle-history-queries.vue
LuisMartinez15: Sure. I'll rename it.

packages/x-components/src/x-modules/history-queries/components/toggle-history-queries.vue
LuisMartinez15: I agree with the renaming of the event.

I've also refactored the wiring to use a `filter` function as you suggested. I had to change the `toggleHistoryQueries` action param type so TS doesn't shout.

packages/x-components/src/x-modules/history-queries/components/toggle-history-queries.vue
LuisMartinez15: Thanks. I've added it.

packages/x-components/src/x-modules/history-queries/components/toggle-history-queries.vue
LuisMartinez15: You're right. Without the confirmation dialog you can't really disable the switch.

packages/x-components/src/x-modules/history-queries/components/toggle-history-queries.vue
javieri-empathy: Thanks for pointing this because I totally forgot about it haha. When I refined the task I think that [we also agreed to](https://searchbroker.atlassian.net/browse/EX-6514?focusedCommentId=153413) refactor the action so it does not depend on a payload. 

Also I just remembered why we decided to emit multiple events reading that comment. It was to ease the usage of that modal component. And right now seeing your updated example with this use case It completely makes sense even if it is not the desired approach.

So I have to say sorry for the misunderstanding and ask you to add the 2 events again.
* Add `UserClickedEnableHistoryQueries` and `UserClickedDisableHistoryQueries`.
* Remove `UserClickedToggleHistoryQueries` event.
* Rename `toggleHistoryQueries` to `setHistoryQueriesEnabled`

Do you think that would work?

packages/x-components/src/x-modules/history-queries/components/history-queries-switch.vue
```diff
@@ -0,0 +1,129 @@
...
+     * Emit events depending on the switch value.
+     *
+     * @internal
+     */
+    protected toggle(): void {
+      if (this.isEnabled) {
+        this.$x.emit('UserClickedToggleHistoryQueries', undefined);
+      } else {
+        this.$x.emit('UserClickedToggleHistoryQueries', true);
+      }
```
javieri-empathy, line 42:
```suggestion
        this.$x.emit(this.isEnabled ? 'UserClickedDisableHistoryQueries' : 'UserClickedEnableHistoryQueries');
```

packages/x-components/src/x-modules/history-queries/store/actions/toggle-history-queries.action.ts
```diff
@@ -8,6 +8,6 @@ export const toggleHistoryQueries: HistoryQueriesXStoreModule['actions']['toggle
       await dispatch('setHistoryQueries', []);
     }
 
-    commit('setIsEnabled', isEnabled);
+    commit('setIsEnabled', isEnabled as boolean);
```
javieri-empathy, line 11:
```suggestion
    commit('setIsEnabled', isEnabled);
```

packages/x-components/src/x-modules/history-queries/store/types.ts
```diff
@@ -174,7 +174,7 @@ export interface HistoryQueriesActions {
    *
    * @param isEnabled - Whether to enable or disable the history queries.
    */
-  toggleHistoryQueries(isEnabled: boolean): void;
+  toggleHistoryQueries(isEnabled?: boolean): void;
```
javieri-empathy, line 177:
```suggestion
  toggleHistoryQueries(isEnabled: boolean): void;
```

packages/x-components/src/x-modules/history-queries/wiring.ts
```diff
@@ -138,7 +142,7 @@ export const historyQueriesWiring = createWiring({
   UserPressedRemoveHistoryQuery: {
     removeHistoryQuery
   },
-  UserToggledHistoryQueries: {
+  UserClickedToggleHistoryQueries: {
     toggleHistoryQueries
   }
```
javieri-empathy, line 147:
```suggestion
  UserClickedEnableHistoryQueries: {
    setHistoryQueriesEnabled
  }
```

packages/x-components/src/x-modules/history-queries/wiring.ts
```diff
@@ -96,7 +97,10 @@ export const removeHistoryQuery = wireDispatch('removeFromHistory');
  *
  * @public
  */
-export const toggleHistoryQueries = wireDispatch('toggleHistoryQueries');
+export const toggleHistoryQueries = filter(
+  wireDispatch('toggleHistoryQueries'),
+  ({ eventPayload: isEnabled }) => isEnabled !== undefined
+);
```
javieri-empathy, line 103:
```suggestion
export const setHistoryQueriesEnabled = wireDispatch('setHistoryQueriesEnabled', true);
```

packages/x-components/src/x-modules/history-queries/events.types.ts
```diff
@@ -9,43 +9,43 @@ import { HistoryQuery } from '@empathyco/x-types';
...
    */
   UserSelectedAHistoryQuery: HistoryQuery;
   /**
-   * The user has toggled the history queries.
-   * * Payload: whether it has been set to enable or disable.
+   * The user has clicked the toggle history queries switch.
+   * Payload: whether it has been set to enable or disable.
    */
-  UserToggledHistoryQueries: boolean;
+  UserClickedToggleHistoryQueries: boolean | undefined;
```
javieri-empathy, line 50:
```ts
  /**
   * The user has clicked a control to enable the history queries.
   * Payload: None.
   */
  UserClickedEnableHistoryQueries: void;
  /**
   * The user has clicked a control to disable the history queries.
   * Payload: None.
   */
  UserClickedDisableHistoryQueries: void;
```

javieri-empathy (changes requested):

javieri-empathy (approved):

packages/x-components/src/x-modules/history-queries/components/history-queries-switch.vue
```diff
@@ -0,0 +1,130 @@
...
+```vue live
+<template>
+  <div>
+    <div>
+      <SearchInput :instant="false" />
+      <SearchButton>Search</SearchButton>
+    </div>
+    <label>
+      History queries:
+      <ToggleHistoryQueries />
```
diegopf, line 91: Shouldn't this be `HistoryQueriesSwitch` now ?

diegopf (changes requested):

diegopf (approved):

javieri-empathy (approved):

PR #598, opened 2022-07-11
feat(my-history): bind `formatTime` to `suggestion-content` slot

diegopf:
binds `formatTime` function to allow formatting dates when overriding slot content.

EX-6611

### DIFF
--- a/packages/x-components/src/x-modules/history-queries/components/my-history.vue
+++ b/packages/x-components/src/x-modules/history-queries/components/my-history.vue
@@ -67,7 +67,6 @@
   import { groupItemsBy, isArrayEmpty } from '../../../utils/array';
   import { SnippetConfig } from '../../../x-installer/api/api.types';
   import { historyQueriesXModule } from '../x-module';
-  import HistoryIcon from '../../../components/icons/history.vue';
   import HistoryQueryComponent from './history-query.vue';
 
   /**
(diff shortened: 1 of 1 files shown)


packages/x-components/src/x-modules/history-queries/components/my-history.vue
```diff
@@ -33,10 +33,13 @@
                   @binding {Suggestion} suggestion - History Query suggestion data
                   @binding {number} index - History Query suggestion index
             -->
-                <slot name="suggestion-content" v-bind="{ suggestion, index }">
-                  <div class="x-list x-list--vertical">
-                    <span>{{ suggestion.query }}</span>
-                    <span>{{ formatTime(suggestion.timestamp) }}</span>
+                <slot name="suggestion-content" v-bind="{ suggestion, index, formatTime }">
+                  <div class="x-list x-list--horizontal">
+                    <HistoryIcon />
```
LuisMartinez15, line 38: I'd just bind the `formatTime` function to the slot and avoid adding the icon here in x. I'd leave that to the `archetype`, where we are usually customising the icons and styles. What do you think?

LuisMartinez15 (changes requested):
good job!

LuisMartinez15 (approved):
gj!

ajperezbau (approved):

PR #599, opened 2022-07-12
feat: Move `XAPI` from `window.X` to `window.InterfaceX`

javieri-empathy:
BREAKING-CHANGE: `XAPI` is now available now at `window.InterfaceX`.
EX-6645

Renames X to InterfaceX so it is less likely to collide with customer clients.


LuisMartinez15 (approved):
thanks!

diegopf (approved):

PR #600, opened 2022-07-13
Bugfix/ex 6457 allow extending all transitions

javieri-empathy:
There were 2 issues before this PR:
- Directional animations (like translate) didn't support extra `$listeners`.
- There was a trick that relied on adding beforeLeave and afterLeave hooks to the transition component so we could delay the overlay animation, but this only worked for the leaving transition

Knowing that, this PR implements following changes:
* Normalises all animation components, so they can both receive extra listeners and props
* Fixes modal open/close logic so it can support independent animations for both the content and the overlay.
* Adds CSS variables for opacity and overlay color (so we can stop using `!important` in clients to override it.
* Adds a `Fade` animation, used by default for the overlay.

### DIFF
--- a/packages/x-components/src/components/animations/collapse-width.vue
+++ b/packages/x-components/src/components/animations/collapse-width.vue
@@ -26,7 +26,7 @@
    */
   @Component({
     mixins: [createCollapseAnimationMixin('width')],
-    inheritAttrs: true
+    inheritAttrs: false
   })
   export default class CollapseWidth extends Vue {
     // TODO Add support for extending enter, after-enter and leave transitions
--- a/packages/x-components/src/components/animations/collapse-width.vue
+++ b/packages/x-components/src/components/animations/collapse-width.vue
@@ -6,6 +6,7 @@
     @leave="collapse"
     appear
     name="x-collapse-width-"
+    v-bind="$attrs"
   >
     <!-- @slot (Required) to add content to the transition -->
     <slot />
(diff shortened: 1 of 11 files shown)


packages/x-components/src/components/animations/collapse-width.vue
```diff
@@ -26,7 +27,9 @@
   @Component({
     mixins: [createCollapseAnimationMixin('width')]
```
LuisMartinez15, line 28: Shouldn't we add `inheritAttrs` here too?

```suggestion
    mixins: [createCollapseAnimationMixin('width')],
    inheritAttrs: true
```

packages/x-components/src/components/animations/fade.vue
```diff
@@ -0,0 +1,70 @@
...
+animate it. The animation just fades in/out the element by changing its opacity.
+
+Wrapping a component:
+
+```vue live
+<template>
+  <div>
+    <button @click="shouldRender = !shouldRender">Toggle</button>
+    <Fade>
+      <p v-if="open">León is southern Spain</p>
```
LuisMartinez15, line 51:
```suggestion
      <p v-if="shouldRender">León is southern Spain</p>
```

LuisMartinez15 (changes requested):
Why are we setting different values to `inheritAttrs` but always binding them to the root element?

packages/x-components/src/components/animations/fade.vue
javieri-empathy: Vaya ojo vida TQ+ (What eye life LY+)

LuisMartinez15 (approved):
good job!

CachedaCodes (approved):

PR #610, opened 2022-07-19
fix(translations): Fix '\r' issue on csv-to-json script

CachedaCodes:
[EX-6631](https://searchbroker.atlassian.net/browse/EX-6631)

## Motivation and context

After the Kenay and Izas setups we found some issues with the `csv-to-json` scripts. 
The first issue was extra `"` on some messages. In this case, the issue didn't come from the script but from the CSV provided by the customers.
The second one was the script inserting `\r` in some points of the JSON and in messages that should be empty. This was caused by the format of the exports of the "Numbers" program. Instead of ending the CSV line with just `\n` it ends them with `\r\n`

## Type of change

- [X] `Main`
- [ ] Other. Specify:

## How has this been tested?

Grab the example CSV file `it.messages.csv` from the task and use the script on it (I changed the source file from the script's test to do so). Test the result with and without the change of the PR applied.

### DIFF
--- a/packages/x-translations/src/csv-to-json.ts
+++ b/packages/x-translations/src/csv-to-json.ts
@@ -30,7 +30,7 @@ function generateJSONFile(sourcePath: string): JSON {
  * @returns A JSON object with the contents of the CSV file.
  */
 function transformToJson(csv: string): JSON {
-  const [header, ...rows] = csv.trim().split(/\r\n|\n/);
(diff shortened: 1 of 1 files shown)


tiborux (approved):

LuisMartinez15 (approved):

LuisMartinez15 (changes requested):
good job!

tiborux (approved):

PR #611, opened 2022-07-19
feat(next-queries): add next query preview logic

ajperezbau:
EX-6119


packages/x-components/src/x-modules/next-queries/store/actions/fetch-and-save-next-query-preview.action.ts
```diff
@@ -0,0 +1,43 @@
+import { NextQueriesXStoreModule } from '../types';
+import { cancellablePromise, CancelSymbol } from '../../../../utils/cancellable-promise';
+
+let cancelPromise: undefined | (() => void);
```
javieri-empathy, line 4: Like below, multiple requests can happen at the same time, and they will override this variable, so this global variable for me does not make sense. 
Probably we can skip cancellation handling for now and simplify this code. I wasn't thinking about adding it when I refined the task, as the worst think that can happen is that you have an additional entry of previews stored. Maybe there are some corner cases like extra parameters that change while you are requesting the preview, but I don't think it is enough to complicate the code right now.

packages/x-components/src/x-modules/next-queries/store/actions/fetch-and-save-next-query-preview.action.ts
```diff
@@ -0,0 +1,43 @@
...
+
+const fetchAndSave: NextQueriesXStoreModule['actions']['fetchAndSaveNextQueryPreview'] = (
+  { dispatch, commit },
+  nextQuery
```
javieri-empathy, line 8: I'm fine with this being either a `query: string` or a `nextQuery: NextQuery`. But I don't think it is a good idea to mix both types and say `nextQuery: string`, as it is confusing for the reader. Indeed in the type definition you named it `query`.

packages/x-components/src/x-modules/next-queries/store/actions/fetch-next-query-preview.action.ts
```diff
@@ -0,0 +1,31 @@
+import { SearchRequest } from '@empathyco/x-types';
+import { NextQueriesXStoreModule } from '../types';
+import { XPlugin } from '../../../../plugins/index';
```
javieri-empathy, line 3: Avoid importing from barrel

packages/x-components/src/x-modules/next-queries/store/types.ts
```diff
@@ -127,3 +164,20 @@ export type NextQueriesActionContext = XActionContext<
   NextQueriesMutations,
   NextQueriesActions
 >;
+
+/**
+ * Interface to type the next query preview objects.
+ */
+interface NextQueryPreviewResults {
```
tiborux, line 171: should be these types on the x-types?

javieri-empathy (dismissed):

packages/x-components/src/x-modules/next-queries/store/module.ts
```diff
@@ -25,9 +30,11 @@ export const nextQueriesXStoreModule: NextQueriesXStoreModule = {
     config: {
       maxItemsToRequest: 20,
       hideSessionQueries: true,
-      loadOnInit: true
+      loadOnInit: true,
+      resultsPreviewCount: 8
     },
-    params: {}
+    params: {},
+    results: {}
```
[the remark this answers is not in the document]
tiborux: agree

packages/x-components/src/x-modules/next-queries/store/types.ts
```diff
@@ -70,6 +78,18 @@ export interface NextQueriesMutations extends StatusMutations, QueryMutations {
    * @param params - The new extra params.
    */
   setParams(params: Dictionary<unknown>): void;
+
+  /**
+   * Adds a new entry to the result's dictionary.
+   *
+   * @param payload - Object containing the next query, the totalResults and the results to add.
+   */
+  setResults(payload: SetNextQueryPreviewResultsPayload): void;
```
tiborux, line 87: Can we simplify the type to be like this? makes sense?

```suggestion
  setResults(payload: Dictionary<NextQueryPreviewResults>): void;
```

packages/x-components/src/x-modules/next-queries/store/module.ts
```diff
@@ -44,13 +51,22 @@ export const nextQueriesXStoreModule: NextQueriesXStoreModule = {
     setStatus,
     setParams(state, params) {
       state.params = params;
+    },
+    setResults(state, { nextQuery, results }) {
+      state.results[nextQuery] = results;
+    },
```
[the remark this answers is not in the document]
ajperezbau: You are right, I forgot that 🤦 .
Fixed.

packages/x-components/src/x-modules/next-queries/store/actions/fetch-next-query-preview.action.ts
```diff
@@ -0,0 +1,31 @@
...
+  { state },
+  query
+) => {
+  if (!query) {
+    return null;
+  }
+  const request: SearchRequest = {
+    query,
+    rows: state.config.resultsPreviewCount,
+    extraParams: state.params
```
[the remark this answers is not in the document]
ajperezbau: Ook, which should be the origin?
Something like `next_query:results`?

packages/x-components/src/x-modules/next-queries/store/module.ts
```diff
@@ -44,13 +48,21 @@ export const nextQueriesXStoreModule: NextQueriesXStoreModule = {
     setStatus,
     setParams(state, params) {
       state.params = params;
+    },
+    setResultsPreview(state, resultsPreview) {
+      state.resultsPreview = Object.assign({}, state.resultsPreview, resultsPreview);
```
tiborux, line 53:
```suggestion
      state.resultsPreview = Object.assign({}, state.resultsPreview, resultsPreview);
```
```suggestion
      state.resultsPreview = {...state.resultsPreview, ...resultsPreview };
```

packages/x-components/src/x-modules/next-queries/store/actions/fetch-next-query-preview.action.ts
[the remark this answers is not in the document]
tiborux: in this case if not sure because the query feature is the next-query, but the feature location is the predictive_layer? Because in this case, they are not still the query inside the grid, but im not sure about this or if we should create a new feature location?

packages/search-types/src/query-signals/preview-results.model.ts
```diff
@@ -0,0 +1,13 @@
+import { Result } from '../result/result.model';
+
+/**
+ * Interface to type the query preview objects.
+ */
+export interface PreviewResults {
```
tajespasarela, line 6: Move this file outside the `query-signals` folder, as it is no longer a next-queries dependant feature

packages/x-components/src/x-modules/next-queries/store/actions/fetch-next-query-preview.action.ts
```diff
@@ -0,0 +1,31 @@
...
+ * @returns A Promise of a SearchResponse when it fetches the results, `null` if the request was
+ * not made.
+ */
+export const fetchNextQueryPreview: NextQueriesXStoreModule['actions']['fetchNextQueryPreview'] = (
+  { state },
+  query
+) => {
+  if (!query) {
+    return null;
+  }
```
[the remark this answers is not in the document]
tajespasarela: I think Abraham is right, and I would leave the empty string check. Specially thinking of reusing the wire with custom wiring config in setups. Better prevent an error request to API

tajespasarela (changes requested):

tiborux (dismissed):

tajespasarela (approved):

tiborux (approved):

PR #615, opened 2022-07-21
feat(history-queries): disable `MyHistory` without confirmation when the history queries list is empty

ajperezbau:
EX-6696

### DIFF
--- a/packages/x-components/src/x-modules/history-queries/components/__tests__/history-queries-switch.spec.ts
+++ b/packages/x-components/src/x-modules/history-queries/components/__tests__/history-queries-switch.spec.ts
@@ -2,20 +2,25 @@ import { DeepPartial } from '@empathyco/x-utils';
 import { createLocalVue, mount, Wrapper } from '@vue/test-utils';
 import Vue from 'vue';
 import Vuex, { Store } from 'vuex';
+import { HistoryQuery } from '@empathyco/x-types';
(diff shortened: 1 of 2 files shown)


packages/x-components/src/x-modules/history-queries/components/__tests__/history-queries-switch.spec.ts
```diff
@@ -52,8 +57,31 @@ describe('testing HistoryQueriesSwitch component', () => {
...
+    const { wrapper } = renderHistoryQueriesSwitch({
+      historyQueries: [],
+      isEnabled: true
+    });
+    const listener = jest.fn();
+    wrapper.vm.$x.on('UserClickedConfirmDisableHistoryQueries').subscribe(listener);
+
+    wrapper.trigger('click');
+
+    expect(listener).toHaveBeenCalledTimes(1);
```
LuisMartinez15, line 71: Can you also check the `isEnabled` flag in the `store` `state`?

```suggestion
  await new Promise(resolve => setTimeout(resolve));

  expect(wrapper.vm.$store.state.x.historyQueries.isEnabled).toBe(false);
  expect(listener).toHaveBeenCalledTimes(1);
```

LuisMartinez15 (changes requested):
good job!

LuisMartinez15 (approved):
good job!

tiborux (approved):

PR #631, opened 2022-07-28
feat(next-queries): add next query preview component

ajperezbau:
EX-6646


annacv (approved):

packages/x-components/src/views/home/Home.vue
```diff
@@ -491,4 +533,7 @@
     height: calc(100% - 20px);
     margin: 10px;
   }
+  .x-next-query-preview__result {
+    max-width: 180px !important;
```
tajespasarela, line 537: inline the styles in the element, no need to add more complex CSS for this view:

`style="max-width: 180px"`

packages/x-components/src/x-modules/next-queries/components/next-query-preview.vue
```diff
@@ -0,0 +1,166 @@
...
+    }
+  };
+</script>
+```
+
+### Play with the default slot
+
+In this example, the results will be rendered inside a sliding panel.
+
+```vue live
```
tajespasarela, line 118: add also an example of the new slot

tajespasarela (changes requested):

tajespasarela (approved):

annacv (approved):

PR #636, opened 2022-07-29
feature/EX-6463-change-aspect-ratio

javieri-empathy:
Modifies the aspect ratio logic to use the modern CSS aspect-ratio property. This also prevents from creating unnecessary layers for the images in the view due to the `position: absolute` property. 
If we use `mix-blend-mode`, this will create another layer, that's why I also removed from the demo the `--colored` modifier.
From my own tests there is no performance improvement on either the paint or composite phases, at least for standard use cases (search & select an RT, search & select a suggestion...). I'd love to have more benchmarks here from different computers. What I did was simply record the performance, and in the bottom-up panel you should be able to see this.

image

To see the layers you can use the `Layers` tool with different css properties to see how it affects. Right now you can see that images, results, and grid are on the same layer:

image

### DIFF
--- a/packages/x-components/src/design-system/components/picture/fixed-ratio.scss
+++ b/packages/x-components/src/design-system/components/picture/fixed-ratio.scss
@@ -1,5 +1,5 @@
 .x-picture--fixed-ratio.x-picture {
   // size
-  aspect-ratio: 1 / var(--x-number-aspect-ratio-picture);
+  aspect-ratio: var(--x-number-aspect-ratio-picture);
   width: 100%;
 }
--- a/packages/x-components/src/design-system/components/picture/fixed-ratio.scss
+++ b/packages/x-components/src/design-system/components/picture/fixed-ratio.scss
@@ -1,19 +1,5 @@
 .x-picture--fixed-ratio.x-picture {
-  // layout
-  position: relative;
-  height: 0;
-
   // size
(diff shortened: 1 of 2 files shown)


annacv (dismissed):
I see some little improvements here: 

Left side: aspect-ratio change applied,
Right side: main branch
afterChange-vs-beforeChange

packages/x-components/src/design-system/components/picture/fixed-ratio.scss
```diff
@@ -1,19 +1,5 @@
...
-  // layout
-  position: relative;
-  height: 0;
-
   // size
-  padding-block-start: calc(var(--x-number-aspect-ratio-picture) * 100%);
-  padding-block-end: 0;
-  padding-inline-start: 0;
-  padding-inline-end: 0;
+  aspect-ratio: 1 / var(--x-number-aspect-ratio-picture);
```
CachedaCodes, line 3: Wasn't the agreement to use just the `var(--x-number-aspect-ratio-picture)` here and marking the merge as BREAKING CHANGE?

LuisMartinez15:
I also see an improvement:

Previous
image

Current
image

LuisMartinez15 (dismissed):
good job!

LuisMartinez15 (approved):

CachedaCodes (approved):

PR #643, opened 2022-08-05
fix(facets): fix `SingleSelectModifier` handling for selecting hierar…

javieri-empathy:
Fixes the issue where selecting a 3rd level hierarchical filter caused it to be deselected when saving the search response. This happened because of the `SingleSelectModifier` retrieved wrongly the list of filter descendants for `HierarchicalFilter`s.

### DIFF
--- a/packages/x-components/src/x-modules/facets/entities/single-select.modifier.ts
+++ b/packages/x-components/src/x-modules/facets/entities/single-select.modifier.ts
@@ -89,9 +89,11 @@ export class SingleSelectModifier extends BaseFilterEntityModifier {
     filter: HierarchicalFilter,
     ids: Array<Filter['id']> = [filter.id]
   ): Array<Filter['id']> {
-    return filter?.children
-      ? filter?.children.flatMap(descendant =>
-          this.getDescendantsIds(descendant, [descendant.id, ...ids])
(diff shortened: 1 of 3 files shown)


packages/x-components/src/x-modules/facets/entities/single-select.modifier.ts
```diff
@@ -89,9 +89,11 @@ export class SingleSelectModifier extends BaseFilterEntityModifier {
     filter: HierarchicalFilter,
     ids: Array<Filter['id']> = [filter.id]
   ): Array<Filter['id']> {
-    return filter?.children
-      ? filter?.children.flatMap(descendant =>
-          this.getDescendantsIds(descendant, [descendant.id, ...ids])
+    return filter?.children?.length
+      ? filter?.children.reduce(
+          (descentantIdsList, descendant) =>
```
LuisMartinez15, line 94: Maybe we should rename the `descendantIdsList` to `_` as it is unused

```suggestion
          (_, descendant) =>
```

LuisMartinez15 (approved):
good job!

LuisMartinez15 (approved):

mnavarroespinosa (approved):

PR #644, opened 2022-08-08
feat(result): add variants support

mnavarroespinosa:
EX-6760

BREAKING CHANGE: `Result` model properties are optional, except `id`, `modelName`, and `tagging`. 

In order to support result variants:
- Now the `Result` properties are optional, except `id`, `modelName`, and `tagging`.
- A result variant can contain more variants.
- As result properties now are optional, I'm not defining a variant as a `Partial<Result>`, let me know if you feel it should be a partial result. 

## Motivation and context
- [x] Open issue. If applicable, link: https://searchbroker.atlassian.net/browse/EX-6760

## Type of change

- [x] `Main`
- [ ] Other. Specify:

### DIFF
--- a/packages/x-components/src/x-modules/identifier-results/components/identifier-result.vue
+++ b/packages/x-components/src/x-modules/identifier-results/components/identifier-result.vue
@@ -57,14 +57,14 @@
      * @public
      */
     protected get highlightedQueryHTML(): string {
-      const identifierValue = this.result.identifier?.value;
+      const identifierValue = this.result.identifier?.value ?? '';
       if (identifierValue && this.identifierHighlightRegexp) {
         return identifierValue.replace(
           this.identifierHighlightRegexp,
           '<span class="x-identifier-result__matching-part">$1</span>'
         );
       }
-      return identifierValue ?? '';
+      return identifierValue;
     }
   }
 </script>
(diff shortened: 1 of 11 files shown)


packages/x-components/src/x-modules/identifier-results/components/identifier-result.vue
```diff
@@ -57,14 +57,14 @@
      * @public
      */
     protected get highlightedQueryHTML(): string {
-      const identifierValue = this.result.identifier.value;
+      const identifierValue = this.result.identifier?.value;
```
LuisMartinez15, line 60: I think we can move the `?? ''` (nullish coalescing operator) here:

```suggestion
      const identifierValue = this.result.identifier?.value ?? '';
```

packages/x-components/src/x-modules/identifier-results/components/identifier-result.vue
```diff
@@ -57,14 +57,14 @@
...
       if (identifierValue && this.identifierHighlightRegexp) {
         return identifierValue.replace(
           this.identifierHighlightRegexp,
           '<span class="x-identifier-result__matching-part">$1</span>'
         );
       }
-      return identifierValue;
+      return identifierValue ?? '';
```
LuisMartinez15, line 67: If you end up moving the nullish coalescing operator to the first line:

```suggestion
      return identifierValue;
```

LuisMartinez15 (dismissed):
good job!

annacv (approved):

LuisMartinez15 (approved):
good job!

PR #655, opened 2022-08-12
feat(design-system): add `tailwindcss` to `x-components` and configure it to use `x-tailwindcss`

mnavarroespinosa:
EX-6793

Configures `x-components` project to use `tailwindcss` and configures it to use `x-tailwindcss` plugin.


packages/x-components/package.json
```diff
@@ -114,9 +116,11 @@
     "rollup-plugin-vue": "~5.1.9",
     "sass": "~1.49.7",
     "sass-loader": "~8.0.0",
+    "tailwindcss": "~3.1.8",
     "ts-jest": "~27.0.7",
     "ts-node": "~8.9.1",
     "typescript": "~4.6.2",
+    "vue-cli-plugin-tailwind": "~3.0.0",
```
CachedaCodes, line 123: With everything now in place do we need this dependency?

packages/x-tailwindcss/src/index.ts
```diff
@@ -1,2 +1,2 @@
 export * from './helpers';
-export { default } from './plugin';
+export { default as xTailwindPlugin } from './plugin';
```
CachedaCodes, line 2: This si a Breaking Change. Should it be added as such in the merge? Are we doing that for the `x-tailwindcss`?

CachedaCodes (changes requested):

packages/x-components/tailwind.config.js
```diff
@@ -0,0 +1,12 @@
+const { xTailwindPlugin } = require('@empathyco/x-tailwindcss');
+
+module.exports = {
+  content: [
+    './index.html',
+    './src/components/**/*.vue',
+    './src/views/**/*.vue',
```
javieri-empathy, line 7: Could we simply include all `.vue` files that are under `src`?: `./src/**/*.vue`

packages/x-components/tailwind.config.js
```diff
@@ -0,0 +1,12 @@
+const { xTailwindPlugin } = require('@empathyco/x-tailwindcss');
+
+module.exports = {
+  content: [
+    './index.html',
```
javieri-empathy, line 5: This file does not exist here. it is located under the `public` folder

javieri-empathy (changes requested):

packages/x-components/package.json
diegopf: Actually not, I will remove it.

packages/x-components/tailwind.config.js
```diff
@@ -0,0 +1,12 @@
...
+    './src/components/**/*.vue',
```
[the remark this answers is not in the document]
diegopf: I don't think so, according to the [tailwind docs](https://tailwindcss.com/docs/configuration#content) we should add the paths to any file that could contain `Tailwind` class names and as far as I can see this is not the case for those files, is it? At least we are not doing that yet.

packages/x-components/package.json
```diff
@@ -79,23 +79,25 @@
...
     "@vue/cli-plugin-typescript": "~5.0.4",
     "@vue/cli-plugin-vuex": "~5.0.4",
     "@vue/cli-service": "~5.0.4",
     "@vue/test-utils": "~1.0.3",
     "@vue/vue2-jest": "~27.0.0-alpha.3",
-    "autoprefixer": "~9.7.4",
+    "autoprefixer": "~10.4.4",
     "cypress": "~8.4.0",
-    "cypress-cucumber-preprocessor": "~4.0.0",
+    "cypress-cucumber-preprocessor": "^4.3.1",
```
javieri-empathy, line 100: Why has this been updated? and why with a minor version?

packages/x-components/tailwind.config.js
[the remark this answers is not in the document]
javieri-empathy: We have some components, like animations built in `.ts` files instead of `.vue`, but they are not using tailwind classes yet. We can keep it like this then.

packages/x-components/tailwind.config.js
```diff
@@ -0,0 +1,12 @@
...
+    './src/views/**/*.vue',
+    './src/x-modules/**/components/**/*.vue'
+  ],
+  prefix: 'x-',
```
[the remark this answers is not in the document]
javieri-empathy: The issue with the prefix is that tailwind does not eliminate all unused CSS when we use the same `x` prefix that we currently have for the old design system. It includes some classes that have the same name but different use cases (x-filter for example), so that's why I would encourage using a different prefix.

javieri-empathy (changes requested):

packages/x-components/tailwind.config.js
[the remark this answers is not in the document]
javieri-empathy: As agreed with @tajespasarela, we will keep the `x` prefix. The idea of using `xds` was to ease the migration process, but it would end up being more complicated, as we would have to keep both classes alive while the migration is ongoing, and once the first full version of the XDS builder is finished remove all `xds` classes. 

To do things step by step and avoid conflicting classes, we are also going to [disable all tailwind core plugins](https://tailwindcss.com/docs/theme#disabling-an-entire-core-plugin), but one or two to ensure that tailwind keeps working because they are 112 plugins, here is the config below:

```
{
    "screens": false,
    "colors": false,
    "columns": false,
    "spacing": false,
    "animation": false,
    "aspectRatio": false,
    "backdropBlur": false,
    "backdropBrightness": false,
    "backdropContrast": false,
    "backdropGrayscale": false,
    "backdropHueRotate": false,
    "backdropInvert": false,
    "backdropOpacity": false,
    "backdropSaturate": false,
    "backdropSepia": false,
    "backgroundColor": false,
    "backgroundImage": false,
    "backgroundOpacity": false,
    "backgroundPosition": false,
    "backgroundSize": false,
    "blur": false,
    "brightness": false,
    "borderColor": false,
    "borderOpacity": false,
    "borderRadius": false,
    "borderSpacing": false,
    "borderWidth": false,
    "boxShadow": false,
    "boxShadowColor": false,
    "caretColor": false,
    "accentColor": false,
    "contrast": false,
    "container": false,
    "content": false,
    "cursor": false,
    "divideColor": false,
    "divideOpacity": false,
    "divideWidth": false,
    "dropShadow": false,
    "fill": false,
    "grayscale": false,
    "hueRotate": false,
    "invert": false,
    "flex": false,
    "flexBasis": false,
    "flexGrow": false,
    "flexShrink": false,
    "fontFamily": false,
    "fontSize": false,
    "fontWeight": false,
    "gap": false,
    "gradientColorStops": false,
    "gridAutoColumns": false,
    "gridAutoRows": false,
    "gridColumn": false,
    "gridColumnEnd": false,
    "gridColumnStart": false,
    "gridRow": false,
    "gridRowStart": false,
    "gridRowEnd": false,
    "gridTemplateColumns": false,
    "gridTemplateRows": false,
    "height": false,
    "inset": false,
    "keyframes": false,
    "letterSpacing": false,
    "lineHeight": false,
    "listStyleType": false,
    "margin": false,
    "maxHeight": false,
    "maxWidth": false,
    "minHeight": false,
    "minWidth": false,
    "objectPosition": false,
    "opacity": false,
    "order": false,
    "padding": false,
    "placeholderColor": false,
    "placeholderOpacity": false,
    "outlineColor": false,
    "outlineOffset": false,
    "outlineWidth": false,
    "ringColor": false,
    "ringOffsetColor": false,
    "ringOffsetWidth": false,
    "ringOpacity": false,
    "ringWidth": false,
    "rotate": false,
    "saturate": false,
    "scale": false,
    "scrollMargin": false,
    "scrollPadding": false,
    "sepia": false,
    "skew": false,
    "space": false,
    "stroke": false,
    "strokeWidth": false,
    "textColor": false,
    "textDecorationColor": false,
    "textDecorationThickness": false,
    "textUnderlineOffset": false,
    "textIndent": false,
    "textOpacity": false,
    "transformOrigin": false,
    "transitionDelay": false,
    "transitionDuration": false,
    "transitionProperty": false,
    "transitionTimingFunction": false,
    "translate": false,
    "width": false,
    "willChange": false,
    "zIndex": false
}
```

Then I'll create another task to create a system to allow dynamic css classes in the components. It doesn't make sense to allow tailwind to set a custom class prefix, while x-components don't support it. This system should also allow to add custom class names to every node in the DOM.

packages/x-components/package.json
diegopf: I honestly don't know, I think is just a mistake. I've reverted it.

javieri-empathy (approved):

CachedaCodes (approved):

Issue #190, opened 2021-10-15
[FEATURE]: Add support to React.context to the ReactWrapper


herrardo:
### How can the project be improved?

The [React Wrapper](https://github.com/empathyco/x/tree/main/packages/react-wrapper) library allows wrapping a Vue component inside a React Component. These React components are ready to work with props, events, slots, scoped slots... but no context.

Context provides a way to pass data through the component tree without having to pass props down manually at every level.

In a typical React application, data is passed top-down (parent to child) via props, but such usage can be cumbersome for certain types of props that are required by many components within an application. Context provides a way to share values like these between components without having to explicitly pass a prop through every level of the tree.

[React.context documentation
](https://reactjs.org/docs/context.html#when-to-use-context)

[Simple example of use](https://stackblitz.com/edit/react-ts-rl5bpy)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

herrardo:
Closing this task as we label it with no priority, thanks Felix for your contribution.

PR #656, opened 2022-08-15
[FEATURE]: Add support to React.context to the ReactWrapper #190

w6e:
Hi guys 
I saw this issue open and decided to give it a try..

I am sure there are some more elegant ways to do this, but to open the conversation I propose here a simple solution. If you have suggestions how to improve, I am happy to take them..

---

I pass a 'context' prop to the ReactWrapper. This prop is fed by a context consumer, which makes it reactive. It also contains a setter to update the context. 

In the VueCreator, I add a provider that makes the context prop globally accessible in the part of the dom tree controlled by the vue instance. 

To read and write the context, any Vue child component can then inject the context value and setter into its scope and use them.

I added a demo page localhost:1234/?view=contextDemo

---

What do you think?

### DIFF
--- a/packages/react-wrapper/tests/demo/stubs/react-user-context.stub.tsx
+++ b/packages/react-wrapper/tests/demo/stubs/react-user-context.stub.tsx
@@ -1,0 +1,67 @@
+import React, { ReactElement, useContext, useState } from 'react'
+import { ReactWrapper } from '../../../src'
+import { VueUserContext } from './vue-user-context'
+import { UserContext } from '../contexts/user-context';
+import styled from 'styled-components'
+
+const Container = styled.div`
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: space-around;
+  max-width: 1000px;
+  margin: 0 auto;
+  background-color: beige;
+  padding: 2rem;
+  
+`
+
+const Column = styled.div`
+  min-width: 400px;
+  
+`
+
+export function ReactUserContext({...props}): ReactElement {
+
+  const [ userContext, setUserContext ] = useContext(UserContext);
+  const [ textBoxState, setTextBoxState ] = useState('')
+
+  const context = {
+    user: {
+      userContext,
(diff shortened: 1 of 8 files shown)


packages/react-wrapper/tests/demo/stubs/react-user-context.stub.tsx
```diff
@@ -0,0 +1,67 @@
...
+const Container = styled.div`
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: space-around;
+  max-width: 1000px;
+  margin: 0 auto;
+  background-color: beige;
+  padding: 2rem;
+  
+`
```
mnavarroespinosa, line 16: If the dependency from `styled-components` is removed, this would need to be refactored, for example to: 

```
const containerStyle: React.CSSProperties = {
  display: 'flex',
  flexWrap: 'wrap',
  justifyContent: 'space-around',
  maxWidth: '1000px',
  margin: '0 auto',
  backgroundColor: 'beige',
  padding: '2rem'
};
```

And then use it in the render as: 
```
  ...
  return (
    <div style={containerStyle}>
    ...
```

mnavarroespinosa (changes requested):
Thanks a lot for the contribution, Felix!

herrardo (commented):
Good morning @w6e 

@mnavarroespinosa left some comments in the PR, but in the second revision, we are seeing that we are not really resolving the original problem, and we are adding complexity, so we might be addressed this in the wrong direction. Please, stand by, we will come up with a final answer soon. Thanks!

mnavarroespinosa:
Hey @w6e, as @herrardo mentioned, we'll try to restate the feature purpose and return with a more defined use case to solve. We're sorry for the inconvenience, and thanks again for the effort.

w6e:
Hi @herrardo , @mnavarroespinosa no problem, I would still be interested to solve it if you come up with a new spec

PR #658, opened 2022-08-17
feat(adapter): disable cancel http client

javieri-empathy:
Due to now using `fetch` with `keepalive` true, we have to add support for our current http client to handle empty responses, and avoid cancelling requests.

EX-6875


packages/x-adapter/src/http-clients/__tests__/fetch.http-client.spec.ts
```diff
@@ -56,58 +56,75 @@ describe('fetch httpClient testing', () => {
...
-          q: 'shirt'
-        }
-      }).then(response => expect(response).toBeDefined()),
-      fetchHttpClient(searchUrl, {
-        parameters: {
-          q: 'jeans'
-        }
-      }).then(response => expect(response).toBeDefined())
-    ]);
+    expect(
```
ajperezbau, line 98: Shouldn't this test be done with cancellable param to true and using different IDs for the requests?

ajperezbau (changes requested):

packages/x-adapter-platform/src/__tests__/__mocks__/fetch.mock.ts
```diff
@@ -4,16 +4,21 @@
  * @param response - The value to return.
  * @returns A Promise that will resolve to the response given.
  */
-export function getFetchMock(response: any): any {
-  return (_url: string, { signal }: any) => {
-    return new Promise((resolve: any, reject: any) => {
+export function getFetchMock(
```
diegopf, line 7: well, my comments from `fetch.mock.ts` file can be applied to this file too😅

packages/x-adapter/src/http-clients/utils.ts
```diff
@@ -13,7 +13,7 @@ import { RequestError } from './errors/request-error';
  */
 export function toJson(response: Response): Promise<any> {
   if (response.ok) {
-    return response.json();
+    return response.text().then(text => text && JSON.parse(text));
```
diegopf, line 16: Maybe I'm thinking too much here but what if the response is not a valid JSON? Should we consider such scenario hand and have handler for it?

diegopf (changes requested):

packages/x-adapter/src/http-clients/__mocks__/fetch.mock.ts
```diff
@@ -19,14 +16,21 @@ export const koFetchMock = jest.fn(() =>
  *
  * @internal
  */
-export function fetchMock(response: any): any {
-  return (_url: string, { signal }: any) => {
-    return new Promise((resolve: AnyFunction, reject: AnyFunction) => {
+export function fetchMock(
+  response: unknown
```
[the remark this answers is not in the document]
javieri-empathy: `unknown` is still type safe, while `any` is not. With `unknown` you prevent the developer to access/use any of the variable properties. With `any` you still allow them. There are only a few good `any` use cases, which essentially means "no types here", usually it can be replaced with `unknown` which is stricter.

packages/x-adapter/src/http-clients/utils.ts
javieri-empathy: You are right, but we haven't developed a proper error handling yet, so I don't know what you would like to do here, if just raising an error with a friendlier message or what. I'm open to it though. 

This code should be equivalent with the previous one but with support for empty strings. What I'm thinking right now is if it makes sense to return a string when we expect a JSON object, or should I return an empty object instead? This wouldn't have happened if this function wasn't using `any` 🤣. Truth is that it returns something `unknown` here, and as we don't know the type the only safe way would be to cast it whenever this `toJson` function is used.

diegopf: I was thinking on catching the error, logging a descriptive message and return `null` and this way we prevent the uncaught exception. However, you are right and this is not within the scope of this PR and just something to think about in the future.

packages/x-adapter/src/http-clients/__tests__/fetch.http-client.spec.ts
ajperezbau: image

ajperezbau (approved):

PR #667, opened 2022-08-19
feat: merge a variant with the result

ajperezbau:
Keep in mind that this implementation is setting the variants in the array of selected variants instead of the indexes, to access the variants of a level and check the selected variant in a simpler way.

Also, the behavior of the auto select will be implemented in another task.

To test it, place the `ResultProvider` wrapping the result in the `Home` component and add the `ResultSelector` in the `Result` component.

Right now the API is not returning variants, so to test it an option is to modify the ResultProvider component to hardcode them, like this:

```
    /**
     * The original result containing the variants.
     *
     * @public
     */
    @Prop({
      required: true
    })
    public result!: Result;
    @XProvide(RESULT_KEY)
    public modifiedResult = {
      ...this.result,
      ...{
        variants: [
          {
            name: 'test',
            images: ['https://assets.empathy.co/images-demo/3935.jpg'],
            variants: [{ name: 'test 2' }]
          },
          { name: 'test 3', images: ['https://assets.empathy.co/images-demo/3935.jpg'] }
        ]
      }
    };
```

EX-6762


packages/x-components/src/components/result/result-selector.vue
```diff
@@ -0,0 +1,336 @@
...
+</script>
+
+<docs lang="mdx">
+## Events
+
+This component doesn't emit events.
+
+## See it in action
+
+Here you have a basic example of how the ResultSelector component is rendered.
```
diegopf, line 142:
```suggestion
Here you have a basic example of how the `ResultSelector` component is rendered.
```

diegopf (commented):
LGTM apart from the doubt I have with the variant selector

packages/x-components/src/components/result/result-provider.vue
```diff
@@ -0,0 +1,192 @@
...
+     * @param variant - The variant to set.
+     * @public
+     */
+    @XProvide(SET_RESULT_VARIANT_KEY)
+    setResultVariant(level: number, variant: ResultVariant): void {
+      if (this.selectedVariants[level] === variant) {
+        return;
+      }
+      this.selectedVariants = this.selectedVariants.slice(0, level);
+      this.$set(this.selectedVariants, level, variant);
```
tajespasarela, line 61: You can do it in one step keeping the reactivity
```suggestion
      this.selectedVariants.splice(level, Number.POSITIVE_INFINITY, variant);
```

tajespasarela (changes requested):

packages/x-components/src/components/result/result-provider.vue
```diff
@@ -0,0 +1,192 @@
...
+     * Selects a variant of the result.
+     * When called, it slices the array of selected variants to remove the selected child variants.
+     * Emits the {@link XEventsTypes.UserSelectedAResultVariant} when called.
+     *
+     * @param level - The nest level where the variant is placed inside the result.
+     * @param variant - The variant to set.
+     * @public
+     */
+    @XProvide(SET_RESULT_VARIANT_KEY)
+    setResultVariant(level: number, variant: ResultVariant): void {
```
tajespasarela, line 56: maybe we can swap the order of the parameters and make the level optional with `0` as the default value:
```suggestion
    setResultVariant(variant: ResultVariant, level: number = 0): void {
```

this case when we only have one level of variants (which will be the usual case), it would be easier to use.

packages/x-components/src/components/result/__tests__/result-provider.spec.ts
```diff
@@ -0,0 +1,247 @@
...
+    expect(wrapper.find(getDataTestSelector('result-name')).text()).toBe('red jacket L');
+
+    // It won't deselect the child variant if the parent is clicked.
+
+    await setRedJacketButton.trigger('click');
+
+    expect(wrapper.find(getDataTestSelector('result-name')).text()).toBe('red jacket L');
+  });
+
+  it('provides the original result', async () => {
```
tajespasarela, line 153: This is not a feature of the Provider itself, but the auxiliary `Child` component.

Maybe renaming to `'keeps the original result unmodified'` or something like that would be clearer

packages/x-components/src/components/result/__tests__/result-selector.spec.ts
```diff
@@ -0,0 +1,290 @@
...
+import { Result, ResultVariant } from '@empathyco/x-types';
+import ResultSelector from '../result-selector.vue';
+import { createResultStub } from '../../../__stubs__/results-stubs.factory';
+import { findTestDataById, getDataTestSelector } from '../../../__tests__/utils';
+import {
+  RESULT_KEY,
+  SELECTED_VARIANTS_KEY,
+  SET_RESULT_VARIANT_KEY
+} from '../../decorators/injection.consts';
+
```
tajespasarela, line 12: Since the `ResultSelector` is hard coupled to the `ResultProvider` I would test that using the `ResultProvider` here in this test.

Only testing `ResultSelector` features, but using the `ResultProvider` and not replicating its features here.

tajespasarela (changes requested):

packages/x-components/src/components/result/__tests__/result-provider.spec.ts
```diff
@@ -0,0 +1,247 @@
...
+
+    await setRedJacketButton.trigger('click');
+
+    expect(wrapper.find(getDataTestSelector('result-name')).text()).toBe('red jacket');
+
+    await setRedJacketXlButton.trigger('click');
+
+    expect(wrapper.find(getDataTestSelector('result-name')).text()).toBe('red jacket L');
+
+    // It won't deselect the child variant if the parent is clicked.
```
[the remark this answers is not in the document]
ajperezbau: Yep, it will be implemented in that task.
But this behavior will depend on the client, in Oysho for example the first color should be selected by default, but not the first size.

packages/x-components/src/components/result/result-provider.vue
```diff
@@ -0,0 +1,192 @@
...
+   *
+   * It receives the original result and keeps track of the selected variant.
+   *
+   * It provides the original result, the array containing the selected variants and a callback to
+   * set the selected variant to be used from a child.
+   *
+   * @public
+   */
+  @Component
+  export default class ResultProvider extends Vue {
```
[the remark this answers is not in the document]
ajperezbau: Yes, I think that `ResultVariantsProvider` describes better the component 👍

packages/x-components/src/components/result/result-provider.vue
ajperezbau: I am not sure about this being an improvement for the actual use case.
That function is intended to be used by the selector, and there the level is always provided to the function.
I can change it anyway to make it easier in case of a new component using only the first level of variants is created.

tajespasarela: yep, but you have to think ahead, if this function is used in a future refactors from anywhere else (other component or whatever), you allow an easy use of it.

diegopf (changes requested):
After the refactor we have `result-variant-selector` and `variants-result-provider` components. Maybe it is just me, but isn't the different naming convention annoying? I'd stick with the same convention and start always with `result-variants-...` and follow that with selector or provider.

tajespasarela (approved):

diegopf (approved):

PR #669, opened 2022-08-22
fix: remove unnecessary token from github action allowing PRs from forks

CachedaCodes:
[EX-6308](https://searchbroker.atlassian.net/browse/EX-6307)

## Motivation and context
PRs from outside the project (specifically from forks) we're failing at the build step because they tried to access a token they didn't have access tot. This token isn't necessary for the workflow, so it has been removed.

## Type of change

- [X] `Main`
- [ ] Other. Specify:

## How has this been tested?

Fork the repo and open a PR from the fork to main with the fix applied. The workflow should work.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,6 @@
 name: Check branch is releasable and release alpha on main branch update
 on: [push, pull_request]
-concurrency: 
+concurrency:
   group: ${{ github.workflow }}-${{ github.ref_name }}
   cancel-in-progress: true
 jobs:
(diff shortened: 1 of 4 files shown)


diegopf (changes requested):
I think we can safely remove this token usage on the other places as well, at least where it is used for the git checkout. See:

[prepare-release.yml](https://github.com/empathyco/x/blob/main/.github/workflows/prepare-release.yml#L10)
[release-alpha.yml](https://github.com/empathyco/x/blob/main/.github/workflows/release-alpha.yml#L11)
[publish-release](https://github.com/empathyco/x/blob/main/.github/workflows/build.yml#L14)

mnavarroespinosa (commented):
Looks good to me, I agree with @diegopf on removing this token usage in other places as well.

mnavarroespinosa (approved):
LGTM!

diegopf (approved):

PR #670, opened 2022-08-23
feat: create `queriesPreview` module

mnavarroespinosa:
EX-6638

## Motivation and context
Create an empty results-preview module with all boilerplate code

## Type of change

- [X] `Main`
- [ ] Other. Specify:

### DIFF
--- a/packages/x-components/src/x-modules/queries-preview/events.types.ts
+++ b/packages/x-components/src/x-modules/queries-preview/events.types.ts
@@ -1,0 +1,7 @@
+/**
+ * Dictionary of the events of QueriesPreview XModule, where each key is the event name, and the
+ * value is the event payload type or `void` if it has no payload.
+ *
+ * @public
+ */
+export interface QueriesPreviewXEvents {}
--- a/packages/x-components/src/x-modules/queries-preview/index.ts
+++ b/packages/x-components/src/x-modules/queries-preview/index.ts
@@ -1,0 +1,4 @@
+export * from './events.types';
+export * from './store';
+export * from './x-module';
+export * from './wiring';
--- a/packages/x-components/src/x-modules/queries-preview/store/types.ts
+++ b/packages/x-components/src/x-modules/queries-preview/store/types.ts
@@ -8,13 +8,6 @@ import { XActionContext, XStoreModule } from '../../../store/index';
 export interface QueriesPreviewState {}
 
 /**
- * QueriesPreview store actions.
- *
- * @public
- */
-export interface QueriesPreviewActions {}
-
-/**
  * QueriesPreview store getters.
  *
  * @public
@@ -27,6 +20,13 @@ export interface QueriesPreviewGetters {}
  * @public
  */
 export interface QueriesPreviewMutations {}
+
+/**
+ * QueriesPreview store actions.
+ *
+ * @public
+ */
+export interface QueriesPreviewActions {}
 
 /**
  * QueriesPreview type safe store module.
--- a/packages/x-components/src/x-modules/queries-preview/store/types.ts
+++ b/packages/x-components/src/x-modules/queries-preview/store/types.ts
@@ -1,0 +1,53 @@
+import { XActionContext, XStoreModule } from '../../../store/index';
+
+/**
+ * QueriesPreview store state.
+ *
+ * @public
+ */
+export interface QueriesPreviewState {}
+
+/**
+ * QueriesPreview store actions.
+ *
+ * @public
+ */
+export interface QueriesPreviewActions {}
+
+/**
+ * QueriesPreview store getters.
+ *
+ * @public
+ */
+export interface QueriesPreviewGetters {}
+
+/**
+ * QueriesPreview store mutations.
+ *
+ * @public
+ */
+export interface QueriesPreviewMutations {}
+
+/**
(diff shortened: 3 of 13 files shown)


packages/x-components/src/x-modules/queries-preview/index.ts
```diff
@@ -0,0 +1,4 @@
+export * from './events.types';
+export * from './store';
+export * from './x-module';
+export * from './wiring';
```
CachedaCodes, line 4: Include and export of the `queries-preview` module in `packages/x-components/src/index.ts` too.

packages/x-components/src/x-modules/queries-preview/store/types.ts
```diff
@@ -0,0 +1,53 @@
...
+ * @public
+ */
+export interface QueriesPreviewState {}
+
+/**
+ * QueriesPreview store actions.
+ *
+ * @public
+ */
+export interface QueriesPreviewActions {}
```
CachedaCodes, line 15: Can you move the actions under the mutations? 🙏 The other modules use that order.

packages/x-components/src/x-modules/queries-preview/wiring.ts
```diff
@@ -0,0 +1,8 @@
+import { createWiring } from '../../wiring/index';
+
+/**
+ * QueriesPreview wiring.
```
CachedaCodes, line 4:
```suggestion
 * Wiring configuration for the {@link QueriesPreviewXModule | queriesPreview module}.
```

packages/x-components/src/x-modules/queries-preview/x-module.ts
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * QueriesPreview {@link XModule} alias.
+ *
+ * @public
+ */
+export type QueriesPreviewXModule = XModule<QueriesPreviewXStoreModule>;
+
+/**
+ * QueriesPreview {@link XModule} implementation. This module is auto-registered as soon as you
+ * import any component from the `queries-preview` entry point.
```
CachedaCodes, line 16: I think here should be `queriesPreview` instead of `queries-preview`. Not 100% if it matters, though.

packages/x-components/src/x-modules/queries-preview/x-module.ts
```diff
@@ -0,0 +1,25 @@
+import { XModule } from '../x-modules.types';
+import { queriesPreviewEmitters } from './store/emitters';
+import { queriesPreviewXStoreModule } from './store/module';
+import { QueriesPreviewXStoreModule } from './store/types';
```
CachedaCodes, line 4: Not from this PR, but, can you change the imports in `x-modules/extra-params/x-module.ts` so it doesn't use the './store' barrel?

````javascript
import { extraParamsEmitters } from './store/emitters';
import { extraParamsXStoreModule } from './store/module';
import { ExtraParamsXStoreModule } from './store/types';
````

File: packages/x-components/src/x-modules/queries-preview/events.types.ts [hunk 6]
```diff
@@ -0,0 +1,7 @@
+/**
+ * Dictionary of the events of QueriesPreview XModule, where each key is the event name, and the
+ * value is the event payload type or `void` if it has no payload.
+ *
+ * @public
+ */
+export interface QueriesPreviewXEvents {}
```
CachedaCodes, line 7: Import this in `src/wiring/events.types.ts`.

CachedaCodes (changes requested):

packages/x-components/src/x-modules/queries-preview/index.ts
mnavarroespinosa: Done!

packages/x-components/src/x-modules/queries-preview/store/types.ts
mnavarroespinosa: Sure!

packages/x-components/src/x-modules/queries-preview/x-module.ts
mnavarroespinosa: I would say it should be `queries-previews` as it is the folder name where the module is defined and the path used when importing the module.

packages/x-components/src/x-modules/queries-preview/x-module.ts
mnavarroespinosa: Sure!

packages/x-components/src/x-modules/queries-preview/events.types.ts
mnavarroespinosa: Done!

packages/x-components/src/x-modules/queries-preview/x-module.ts
diegopf: yep, agree with @mnavarroespinosa. Actually in other x-modules we do the same. See [search-box x-module](https://github.com/empathyco/x/blob/main/packages/x-components/src/x-modules/search-box/x-module.ts#L17).

diegopf (changes requested):

diegopf (approved):

CachedaCodes (approved):

PR #674, opened 2022-08-24
build: Support SSR with createInjectorSSR for styles.

tajespasarela:
EX-6892


packages/x-components/build/rollup.config.ts
```diff
@@ -93,10 +93,13 @@ export const rollupConfig = createRollupOptions({
       mode: [
         'inject',
         (varname: string, id: string) =>
-          // eslint-disable-next-line max-len
-          `import {createInjector} from 'vue-runtime-helpers';const injector=createInjector({});injector('${normalizePath(
-            id
-          )}',{source:${varname}})`
+          `import {createInjector, createInjectorSSR} from 'vue-runtime-helpers';
+           const isBrowser = (function(){try {return this!==global;}catch(e){return true;}})();
```
mnavarroespinosa, line 97: This function is returning `true` when I execute it in a Node.js environment, what do you think of changing it to?
```suggestion
           const isBrowser = (function(){try {return this===window;}catch(e){return false;}})();
```

mnavarroespinosa (changes requested):

packages/x-components/build/rollup.config.ts
javieri-empathy: FYI, because we are executing this check in a function, rollup is not able to remove it. whenever it feels the resultant variant is not needed. To fix it you can either not execute it in an iife, or flag the function as pure (no side effects) with the `/*#__PURE__*/`.
image
image

mnavarroespinosa: Update: apparently, it was returning `true` because I ran the script in a project using ES modules instead of CommonJS, which uses `strict` mode by default causing `this` to behave differently. This is another approach, by @javieri-empathy, which seems to work properly:
```suggestion
           const isBrowser = (function(){return (Object.prototype.toString.call(typeof process!=='undefined'? process:0)!=='[object process]');})();
```

mnavarroespinosa (changes requested):
@tajespasarela what do you think about @javieri-empathy [comment](https://github.com/empathyco/x/pull/674#discussion_r955951142)? about not using an iife or flagging the function as pure so it can be removed by rollup

tajespasarela:
> @tajespasarela what do you think about @javieri-empathy [comment](https://github.com/empathyco/x/pull/674#discussion_r955951142)? about not using an iife or flagging the function as pure so it can be removed by rollup

The function was necessary when the `try{}catch()` was necessary. It is not anymore, so we can replace the function by an stament.

javieri-empathy (approved):

mnavarroespinosa (approved):

PR #677, opened 2022-08-25
feat(queries-preview): add wiring and actions

mnavarroespinosa:
[EX-6844](https://searchbroker.atlassian.net/browse/EX-6844)

## Motivation and context
Added wiring and actions to the `queries-preview` module.

- The module's state now stores a dictionary of `QueryPreviewItems`, where each item contains, mainly, the results of a specific query and the request used for the search.

- `fetchQueryPreview` and `fetchAndSaveQueryPreview` actions have been added in order to perform the request and save the results, and the request, used for the search.

#### Events
- Whenever the `QueryPreviewRequestChange` is emitted, a search will be performed with the passed request.
> The component described by [EX-6845](https://searchbroker.atlassian.net/browse/EX-6845) will be in charge of emitting this event if the `extraParams` change.

- `QueryPreviewRemoved` event will remove the entry of a specific query from the `queriesPreview` dictionary.

## Type of change

- [x] `Main`
- [ ] Other. Specify:

### DIFF
--- a/packages/x-components/src/x-modules/queries-preview/events.types.ts
+++ b/packages/x-components/src/x-modules/queries-preview/events.types.ts
@@ -11,7 +11,7 @@ export interface QueriesPreviewXEvents {
    * Any property of the queries preview request has changed.
    * Payload: The new {@link @empathyco/x-types#SearchRequest | request}.
    */
-  QueryPreviewRequestChange: SearchRequest;
+  QueryPreviewRequestChanged: SearchRequest;
   /**
    * The component that shows a Query preview has been unmounted.
    * Payload: The query whose preview has been removed.
--- a/packages/x-components/src/x-modules/queries-preview/events.types.ts
+++ b/packages/x-components/src/x-modules/queries-preview/events.types.ts
@@ -8,7 +8,7 @@ import { SearchRequest } from '@empathyco/x-types';
  */
 export interface QueriesPreviewXEvents {
   /**
-   * Query preview request have been changed.
+   * Any property of the queries preview request has changed.
    * Payload: The new {@link @empathyco/x-types#SearchRequest | request}.
    */
   QueryPreviewRequestChange: SearchRequest;
--- a/packages/x-components/src/x-modules/queries-preview/events.types.ts
+++ b/packages/x-components/src/x-modules/queries-preview/events.types.ts
@@ -1,7 +1,20 @@
+import { SearchRequest } from '@empathyco/x-types';
+
 /**
  * Dictionary of the events of QueriesPreview XModule, where each key is the event name, and the
  * value is the event payload type or `void` if it has no payload.
  *
  * @public
  */
-export interface QueriesPreviewXEvents {}
+export interface QueriesPreviewXEvents {
+  /**
+   * Query preview request have been changed.
+   * Payload: The new {@link @empathyco/x-types#SearchRequest | request}.
+   */
+  QueryPreviewRequestChange: SearchRequest;
+  /**
+   * The component that shows a Query preview has been unmounted.
+   * Payload: The query whose preview has been removed.
+   */
+  QueryPreviewRemoved: string;
+}
--- a/packages/x-components/src/x-modules/queries-preview/store/__tests__/actions.spec.ts
+++ b/packages/x-components/src/x-modules/queries-preview/store/__tests__/actions.spec.ts
@@ -65,10 +65,10 @@ describe('testing queries preview module actions', () => {
       const query = 'tshirt';
       const request = getQueryPreviewRequest(query);
 
-      const promise = store.dispatch('fetchAndSaveQueryPreview', request);
+      await store.dispatch('fetchAndSaveQueryPreview', request);
       await promise;
 
-      const expectedResults = {
+      const expectedResults:  QueryPreviewItem = {
         totalResults: mockedSearchResponse.totalResults,
         results: mockedSearchResponse.results,
         query,
--- a/packages/x-components/src/x-modules/queries-preview/store/__tests__/actions.spec.ts
+++ b/packages/x-components/src/x-modules/queries-preview/store/__tests__/actions.spec.ts
@@ -12,6 +12,7 @@ import {
   QueriesPreviewActions,
   QueryPreviewItem
 } from '../types';
+import { resetQueriesPreviewStateWith } from './utils';
 
 describe('testing queries preview module actions', () => {
   const mockedSearchResponse = getSearchResponseStub();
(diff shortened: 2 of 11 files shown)


packages/x-components/src/x-modules/queries-preview/events.types.ts
```diff
@@ -1,7 +1,20 @@
...
 /**
  * Dictionary of the events of QueriesPreview XModule, where each key is the event name, and the
  * value is the event payload type or `void` if it has no payload.
  *
  * @public
  */
-export interface QueriesPreviewXEvents {}
+export interface QueriesPreviewXEvents {
+  /**
+   * Query preview request have been changed.
```
tiborux, line 11:
```suggestion
   * Any property of the queries preview request has changed.
```

To match other modules

packages/x-components/src/x-modules/queries-preview/store/__tests__/actions.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+
+  describe('fetchAndSaveQueryPreview', () => {
+    it('should request and store query preview results in the state', async () => {
+      const query = 'tshirt';
+      const request = getQueryPreviewRequest(query);
+
+      const promise = store.dispatch('fetchAndSaveQueryPreview', request);
+      await promise;
+
+      const expectedResults = {
```
javieri-empathy, line 71: For complex types that are no inferred like this one, it is always better to have them explicitly typed to ease future refactors.

```suggestion
      const expectedResults:  QueryPreviewItem = {
```

packages/x-components/src/x-modules/queries-preview/store/__tests__/actions.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+    });
+  });
+
+  describe('fetchAndSaveQueryPreview', () => {
+    it('should request and store query preview results in the state', async () => {
+      const query = 'tshirt';
+      const request = getQueryPreviewRequest(query);
+
+      const promise = store.dispatch('fetchAndSaveQueryPreview', request);
+      await promise;
```
javieri-empathy, line 69: Why store the promise and await it in another line?

packages/x-components/src/x-modules/queries-preview/store/__tests__/actions.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+  localVue.config.productionTip = false; // Silent production console messages.
+  localVue.use(Vuex);
+
+  const store: SafeStore<
+    QueriesPreviewState,
+    QueriesPreviewGetters,
+    QueriesPreviewMutations,
+    QueriesPreviewActions
+  > = new Store(queriesPreviewXStoreModule as any);
+  installNewXPlugin({ adapter, store }, localVue);
```
javieri-empathy, line 39: You are creating a global store, and you never reset its state. Therefore each test case result is dependant from the previous test. You can either do something similar like we do with component render factories, or reset the store state before every test.

packages/x-components/src/x-modules/queries-preview/store/__tests__/actions.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+      expect(results).toEqual(mockedSearchResponse);
+    });
+
+    it('should return `null` if the query is empty', async () => {
+      const request = getQueryPreviewRequest('');
+      expect(await store.dispatch('fetchQueryPreview', request)).toBeNull();
+    });
+  });
+
+  describe('fetchAndSaveQueryPreview', () => {
```
javieri-empathy, line 63: Can we have some tests asserting the different status values?

javieri-empathy (changes requested):

packages/x-components/src/x-modules/queries-preview/wiring.ts
```diff
@@ -1,8 +1,64 @@
...
+ * @public
+ */
+export const fetchAndSaveQueryPreviewWire = wireDispatch('fetchAndSaveQueryPreview');
+
+/**
+ * Clears a query preview from queries preview module.
+ *
+ * @public
+ */
+export const clearQueryPreview = wireCommit('clearQueryPreview');
```
tiborux, line 41: Add new empty line.

packages/x-components/src/x-modules/queries-preview/store/__tests__/actions.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+    it('should request and store query preview results in the state', async () => {
+      const query = 'tshirt';
+      const request = getQueryPreviewRequest(query);
+
+      const promise = store.dispatch('fetchAndSaveQueryPreview', request);
```
tiborux, line 68:
```suggestion
      await store.dispatch('fetchAndSaveQueryPreview', request);
```

packages/x-components/src/x-modules/queries-preview/store/__tests__/actions.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+  localVue.use(Vuex);
+
+  const store: SafeStore<
+    QueriesPreviewState,
+    QueriesPreviewGetters,
+    QueriesPreviewMutations,
+    QueriesPreviewActions
+  > = new Store(queriesPreviewXStoreModule as any);
+  installNewXPlugin({ adapter, store }, localVue);
+
```
tiborux, line 40: Use the util that you created, right now is not. being used

```
beforeEach(() => {
    resetQueriesPreviewStateWith(store);
  });
  ```

**Review by tiborux — CHANGES_REQUESTED** [review]
gj

File: packages/x-components/src/x-modules/queries-preview/store/__tests__/actions.spec.ts [hunk 9]
```diff
@@ -0,0 +1,106 @@
...
+  describe('fetchQueryPreview', () => {
+    it('should make a search request with a unique id', async () => {
+      const request = getQueryPreviewRequest('sandals');
```
tiborux, line 43: This test and the next one can be combined in one.

packages/x-components/src/x-modules/queries-preview/store/__tests__/actions.spec.ts
mnavarroespinosa: You're right! I actually added the util to reset it but forgot to use it in the tests 😅. I've added it!

packages/x-components/src/x-modules/queries-preview/store/__tests__/actions.spec.ts
mnavarroespinosa: Sure, I've combined them.

packages/x-components/src/x-modules/queries-preview/store/__tests__/actions.spec.ts
mnavarroespinosa: Sure! I've change it, it did not make sense.

javieri-empathy (approved):

tiborux (approved):

PR #686, opened 2022-08-30
build(dependencies): fix Vue & Vuex versions

javieri-empathy:
EX-6918

The recent Vue 2.7 release breaks some things. Until we fix them, X-Components is only compatible with the 2.6.x version. To avoid accidental updates, we are setting `Vue` and `Vuex` as `peerDependencies` (we want to make sure we use the version installed by the user project), and setting them also as `devDependencies` with minor versions for our local development.


tiborux (approved):
gj

tajespasarela (approved):

PR #688, opened 2022-09-01
fix(components): Fix the error on console in `BaseResultImage` with `CrossFade` Animation.

tajespasarela:
EX-6924

- This PR fixes an error raised in the console when the `BaseResultImage` component is used with the `CrossFade` animation and the result image is not loaded (image URL not found).

- Also refactored the component to use native [lazy loading](https://developer.mozilla.org/en-US/docs/Web/Performance/Lazy_loading) and get rid of the Intersection observers.


packages/x-components/src/components/result/__tests__/base-result-image.spec.ts
```diff
@@ -34,7 +34,7 @@ describe('testing BaseResultImage component', () => {
         };
       }
     });
-    const image = customResultImageWrapper.get(getDataTestSelector('result-picture-image'));
+    const image = customResultImageWrapper.get(getDataTestSelector('result-picture-loader'));
```
javieri-empathy, line 37: Tests still have data from previous version.

javieri-empathy (changes requested):

packages/x-components/src/components/result/base-result-image.vue
```diff
@@ -1,36 +1,32 @@
 <template>
   <picture ref="image" class="x-picture x-result-picture" data-test="result-picture">
-    <component
-      :is="animation"
+    <img
       v-if="!hasImageLoaded && !hasAllImagesFailed"
```
javieri-empathy, line 4: Because of the bug in Safari about empty SRCs reported by Wences and Ivan,
Could we remove the `hasAllImagesFailed` computed property and just use `!imageSrc` instead?. Similar approach for below changes. 
```suggestion
      v-if="!hasImageLoaded && imageSrc"
```

tiborux (dismissed):

annacv (commented):
It would be nice if anyone else could check this: in my local, we have removed the error in console but now the fallback image is not being shown:

Captura de pantalla 2022-09-06 a les 9 25 05


Here, the behavior in main (error console but fallback ok):

Captura de pantalla 2022-09-06 a les 9 27 25

packages/x-components/src/components/result/base-result-image.vue
```diff
@@ -1,36 +1,30 @@
...
-    <component :is="animation">
+    <img
+      v-if="!hasImageLoaded && imageSrc"
+      @error="flagImageAsFailed"
+      @load="flagImageLoaded"
+      loading="lazy"
+      :src="imageSrc"
+      :style="loaderStyles"
+      data-test="result-picture-loader"
+      alt=""
```
tiborux, line 11: I know that this is not going to be seen, but should we add a right alt?

javieri-empathy: Trying to research a little bit into `alt` attributes, I found[ this awesome page](https://www.w3.org/WAI/tutorials/images/decision-tree/) explaining its use cases using a decision tree. In this case the image is not intended for the user, so we can use an empty alt. In the other case I was hesitant about using `alt="result.name"`. My concern is that the result name does not describe the image content. But as per that decision tree, when the image is used inside a link with no more text, which is the case that we currently have in our archetype, the alt can be used to identify the destination of the link, which is the result page, so we should be fine.

javieri-empathy:
> It would be nice if anyone else could check this: in my local, we have removed the error in console but now the fallback image is not being shown:
> 
> Captura de pantalla 2022-09-06 a les 9 25 05
> 
> Here, the behavior in main (error console but fallback ok):
> 
> Captura de pantalla 2022-09-06 a les 9 27 25

If you modify the HTML of the loaded image to trigger an error we won't listen to them. As we have proven the image to be valid, and loaded it using another instance of the `img` element, we don't have to listen to errors there, we can trust the image to load successfully and without an additional request.

packages/x-components/src/components/result/base-result-image.vue
tiborux: So, should we make that change in this pr?

tiborux (approved):

javieri-empathy (approved):

PR #699, opened 2022-09-02
feat(queries-preview): add query preview component 

mnavarroespinosa:
[EX-6845](https://searchbroker.atlassian.net/browse/EX-6845)

Create a `queryPreview` component that renders nothing (no carousel, no grid), by default it only displays a list of results names. It also makes a `Provide` of the results.  

**Slots**

It allows customizing the content through:
- the `default` slot
- or customize just each result through the `result` slot.

**Events**

- `QueryPreviewRequestChanged`: the event is emitted when the component is mounted and when the
  properties of the request object change. The event payload is the `queryPreviewRequest` object.
- When the component is destroyed, it is not emitting the `QueryPreviewRemoved` on purpose:

  I don't see clearly that removing the preview results from the state when the component is unmounted would be beneficial. In some scenarios, we could be asking for the same results many times, so I think we should add a config to determine if this event should be emitted or not, as a different task.

**Example**

As an example, three carousels have been added to the `Home` view to be displayed when there is no query in the search box.

## Type of change

- [x] `Main`
- [ ] Other. Specify:

## How has this been tested?

## Checklist:

### DIFF
--- a/packages/x-components/src/views/home/Home.vue
+++ b/packages/x-components/src/views/home/Home.vue
@@ -214,7 +214,7 @@
             </div>
 
             <template v-if="!$x.query.searchBox">
-              <h1 class="x-title1 x-margin--bottom-06">Discovery wall</h1>
+              <h1 class="x-title1 x-margin--bottom-06">Brand Recommendations</h1>
 
               <SlidingQueryPreview query="sunglasses" />
               <SlidingQueryPreview query="handbag" />
--- a/packages/x-components/src/views/home/Home.vue
+++ b/packages/x-components/src/views/home/Home.vue
@@ -213,6 +213,14 @@
               <p>You may be interested in these:</p>
             </div>
 
+            <template v-if="!$x.query.searchBox">
+              <h1 class="x-title1 x-margin--bottom-06">Discovery wall</h1>
+
+              <SlidingQueryPreview query="sunglasses" />
+              <SlidingQueryPreview query="handbag" />
+              <SlidingQueryPreview query="earrings" />
+            </template>
+
             <!-- Results -->
             <LocationProvider location="results">
               <ResultsList v-infinite-scroll:main-scroll>
@@ -399,13 +407,13 @@
   import Result from './result.vue';
   import { HomeControls } from './types';
   import SlidingNextQueryPreview from './sliding-next-query-preview.vue';
+  import SlidingQueryPreview from './sliding-query-preview.vue';
 
   @Component({
     directives: {
       infiniteScroll
     },
     components: {
-      SlidingNextQueryPreview,
       ArrowRight,
       Aside,
       AutoProgressBar,
@@ -458,7 +466,9 @@
       SearchButton,
       SearchIcon,
       SearchInput,
+      SlidingNextQueryPreview,
       SlidingPanel,
+      SlidingQueryPreview,
       SnippetCallbacks,
       SnippetConfigExtraParams,
       SortDropdown,
--- a/packages/x-components/src/views/home/sliding-query-preview.vue
+++ b/packages/x-components/src/views/home/sliding-query-preview.vue
@@ -1,0 +1,39 @@
+<template>
+  <QueryPreview :query="query" #default="{ results, totalResults, query }">
+    <div class="x-list x-list--gap-03 x-margin--bottom-05">
+      <h1 class="x-title2 x-text--bold">{{ query }} ({{ totalResults }})</h1>
+      <SlidingPanel :resetOnContentChange="false">
+        <div class="x-list x-list--gap-03">
+          <Result
+            v-for="result in results"
+            :key="result.id"
+            :result="result"
+            style="max-width: 180px"
+          />
+        </div>
+      </SlidingPanel>
+    </div>
+  </QueryPreview>
+</template>
+
+<script lang="ts">
+  import { Component, Prop } from 'vue-property-decorator';
+  import Vue from 'vue';
+  import QueryPreview from '../../x-modules/queries-preview/components/query-preview.vue';
+  import SlidingPanel from '../../components/sliding-panel.vue';
+  import Result from './result.vue';
+
+  @Component({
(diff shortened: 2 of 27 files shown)


packages/x-components/src/x-installer/api/api.types.ts
```diff
@@ -116,7 +116,7 @@ export interface SnippetConfig {
   /** The filters to be applied on the first request. */
   filters?: string[];
   /** List of queries to preview. */
-  queriesPreview?: QueryPreview[];
+  queriesPreview?: QueryPreviewConfig[];
```
tiborux, line 119: I think that this name its better for what its being used for, but I'm not 100% sure if i like it hahahaha

packages/x-components/src/x-modules/queries-preview/components/__tests__/query-preview.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+        request: {
+          query: 'milk'
+        },
+        results: [],
+        status: 'initial',
+        totalResults: 0
+      }
+    });
+
+    expect(wrapper.find(getDataTestSelector('query-preview')).element).toBeFalsy();
```
tiborux, line 200: method `find` is deprecated. in this case you can use `findComponent(QueryPreview)`

https://v1.test-utils.vuejs.org/api/wrapper/find.html

packages/x-components/src/views/home/Home.vue
```diff
@@ -213,6 +213,14 @@
               <p>You may be interested in these:</p>
             </div>
 
+            <template v-if="!$x.query.searchBox">
+              <h1 class="x-title1 x-margin--bottom-06">Discovery wall</h1>
```
herrardo, line 217: As the Discovery Wall concept I think is being deprecated...could you change the title?

```suggestion
              <h1 class="x-title1 x-margin--bottom-06">Brand Recommendations</h1>
```

herrardo (changes requested):

packages/x-components/src/views/home/sliding-query-preview.vue
```diff
@@ -0,0 +1,39 @@
+<template>
+  <QueryPreview :query="query" #default="{ results, totalResults, query }">
+    <div class="x-list x-list--gap-03 x-margin--bottom-05">
+      <h1 class="x-title2 x-text--bold">{{ query }} ({{ totalResults }})</h1>
```
herrardo, line 4: Also, in this task: https://searchbroker.atlassian.net/browse/EX-6925

We are describing that will need to receive a title and a query in the snippet. Do we need to add a prop in the title, or will we add the prop on a custom wrapper component in the Archetype?

packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
```diff
@@ -0,0 +1,316 @@
+<template>
+  <NoElement v-if="queryPreviewResults && queryPreviewResults.totalResults">
```
tiborux, line 2: Should we use the same formula that we used for other components? Like having a list that can be concatenated with another list and finally you can modify the component using the slot:

image

I know that this is already working, but im not sure if we want to have the same way of concatenating things that are going to render results.

tiborux (changes requested):
gj

diegopf (changes requested):

packages/x-components/src/x-installer/api/api.types.ts
diegopf: What about `QueriesPreviewList`?

packages/x-components/src/views/home/sliding-query-preview.vue
mnavarroespinosa: We don't need to add it here. This `SlidingQueryPreview` is used just for the demo of `x-components`. Later in the Archetype, we will use the title provided by the snippet config info in a custom wrapper component, as you said.

packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
```diff
@@ -0,0 +1,316 @@
...
+    })
+    protected query!: string;
+
+    /**.
+     *  The origin property for the request
+     *
+     * @public
+     */
+    @Prop()
+    protected queryOrigin!: QueryOrigin;
```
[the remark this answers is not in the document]
mnavarroespinosa: I don't think so, good point!

packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
```diff
@@ -0,0 +1,316 @@
...
+     */
+    public get queryPreviewResults(): QueryPreviewItem | undefined {
+      const previewResults = this.previewResults[this.query];
+      return previewResults?.results
+        ? {
+            ...previewResults,
+            results: previewResults.results.slice(0, this.maxItemsToRender)
+          }
+        : undefined;
+    }
```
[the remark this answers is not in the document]
mnavarroespinosa: Good catch! I totally agree, it makes more sense to just bind the needed properties. Changed!

packages/x-components/src/x-modules/queries-preview/components/__tests__/query-preview.spec.ts
mnavarroespinosa: Good point, I have changed it. Actually, `renderQueryPreview` is already using `findComponent` when returning the `wrapper` 🤦‍♂️

packages/x-components/src/x-installer/api/api.types.ts
mnavarroespinosa: I'm not sure about `QueriesPreviewList` as it should represent a single item instead of a list. What about `QueryPreviewDetails`, `QueryPreviewParams` or `QueryToPreview`?

herrardo (approved):

diegopf (approved):

PR #713, opened 2022-09-09
Move i18 plugin

annacv:
[EX-6966](https://searchbroker.atlassian.net/browse/EX-6966)


packages/x-archetype-utils/src/i18n/i18n.plugin.ts
```diff
@@ -0,0 +1,154 @@
+import { VueConstructor } from 'vue';
+import VueI18n, { LocaleMessageObject } from 'vue-i18n';
+import { deepMerge } from '@empathyco/x-deep-merge';
+import { AnyMessages, Device, I18nOptions, LoadLazyMessagesByDevice, Locale } from './i18n.types';
+
+/**
+ * I18n settings manager.
+ */
+export class I18n {
```
javieri-empathy, line 9: With the types that we modified, we are losing message types. Errors that previously were triggered  now they aren't. We can solve them if we add a generic parameter to this class representing the full messages object for a given lang, and we use it across the methods or properties.

packages/x-archetype-utils/src/i18n/i18n.plugin.ts
```diff
@@ -0,0 +1,160 @@
...
+    const index = keyParts.findIndex(keyPart => {
+      if (typeof keyPartMessages[keyPart] === 'object') {
+        keyPartMessages = <LocaleMessageObject>keyPartMessages[keyPart];
+      } else {
+        return true;
+      }
+    });
+
+    return index !== -1 ? <string>keyPartMessages?.[keyParts.slice(index).join('.')] : '';
+  }
```
javieri-empathy, line 87: I would propose a little refactor with my favorite tool, the `reduce`.

-  I find quite weird mutating a variable outside the `findIndex` loop.
- This function should return a nullish value as per the usage above.
- `typeof keyPartMessages[keyPart]` can return `true` if the value is `null`. We should use `isObject`
- Messages can be an array or a function. I would clarify that we do not support this yet (I tried, but I think the benefit is not much and it complicates the code.

```ts
  /**
   * Tries to get a value from the messages object accessing it key by key.
   * When a key fails, takes the remaining key parts and tries to access the message
   * joining them into a single key.
   *
   * @param path - A dot separated key to access the message object.
   * @returns The message corresponding to the provided key, or empty string if it doesn't exist.
   */
  protected getMessageWithDotsInKey(path: string): string | null {
    // TO-DO Support function or array messages
    const message = path
      .split('.')
      .reduce<LocaleMessage>(
        (messages, key, index, pathParts) =>
          isObject(messages)
            ? messages[key] ?? messages[pathParts.slice(index).join('.')]
            : messages,
        this.currentMessages
      );
    return typeof message === 'string' ? message : null;
  }
```

javieri-empathy (changes requested):

packages/x-archetype-utils/package.json
```diff
@@ -33,13 +33,23 @@
     "test": "jest",
     "prepublishOnly": "npm run build"
   },
+  "dependencies": {
+    "@empathyco/x-deep-merge": "^1.3.0-alpha.19",
+    "@empathyco/x-utils": "^1.0.0-alpha.5"
```
LuisMartinez15, line 38: This dependency is only use for testing purposes. I think it can be moved to the `devDependencies` section.

Maybe I'm wrong and setting it as a `peerDependency` is enough, but... shouldn't `VueI18n` be moved to the `dependencies` section? It is `imported` in the `i18n.plugin.ts` and we're creating a VueI18n instance there.

LuisMartinez15 (changes requested):
good job!

javieri-empathy (approved):

LuisMartinez15 (approved):

PR #714, opened 2022-09-09
feat: add `uiLang` parameter

javieri-empathy:
EX-6855

Adds suppport for the `uiLang` paramenter. This is an optional paramenter of the `SnippetConfig`. When no passed it should default to `lang`.


packages/x-components/src/x-installer/x-installer/x-installer.ts
```diff
@@ -292,6 +294,25 @@ export class XInstaller {
...
+   * @param snippetConfig - The snippet config to normalise.
+   * @returns The normalised version of the given snippet config.
+   * @internal
+   */
+  protected normaliseSnippetConfig(
+    snippetConfig: SnippetConfig | Partial<SnippetConfig>
+  ): NormalisedSnippetConfig | Partial<SnippetConfig> {
+    if (snippetConfig.lang) {
+      snippetConfig.uiLang ??= snippetConfig.lang;
+    }
```
javieri-empathy, line 312: The watcher approach, while ideal to have in one place the side effects of mutating the snippet config brought some issues, as you lost the information of which properties where modified, so instead I opted to do a more imperative approach in the only method that had all the information we needed to make a decision.

packages/x-components/src/x-installer/x-installer/x-installer.ts
```diff
@@ -276,13 +276,15 @@ export class XInstaller {
   ): Vue | undefined {
     if (this.options.app !== undefined) {
       const vue = this.getVue();
-      this.snippetConfig = vue.observable(snippetConfig);
+      const normalisedSnippetConfig = (this.snippetConfig = vue.observable(
+        this.normaliseSnippetConfig(snippetConfig)
+      ));
```
mnavarroespinosa, line 281: Just to understand it, why do we need to define `normalisedSnippetConfig`? Instead of:
```
this.snippetConfig = vue.observable(this.normaliseSnippetConfig(snippetConfig));
```

javieri-empathy: Good question! I did that because of testing code:

- One of the things I did previously was cloning the client snippet config. I did it because it felt natural for me that as we are going to modify an object provided by an external user, it felt safer to clone it. But I discarded this approach
- `provide` functions have bound Vue `this` context, so it wasn't accessible. I now realised that even if we don't need that normalised variable because we mutate rather than clone, I would change the `provide` syntax to use the object syntax.

mnavarroespinosa (approved):
good job!

packages/x-components/src/x-installer/x-installer/x-installer.ts
```diff
@@ -247,43 +248,35 @@ export class XInstaller {
...
    */
-  protected createApp(
-    extraPlugins: VueConstructorPartialArgument,
-    snippetConfig: SnippetConfig
-  ): Vue | undefined {
+  protected createApp(extraPlugins: VueConstructorPartialArgument): Vue | undefined {
     if (this.options.app !== undefined) {
       const vue = this.getVue();
-      this.snippetConfig = vue.observable(snippetConfig);
+      vue.observable(this.snippetConfig);
```
mnavarroespinosa, line 274: According to the [docs](https://v2.vuejs.org/v2/api/#Vue-observable), It seems that `Vue.observable` directly mutates the object passed to it in Vue 2, but it won't in Vue 3, do we want to already work with the returned object or we shouldn't worry about that now?

LuisMartinez15: @mnavarroespinosa You're right. I'd change it now as we will probably forget later on when migrating to `vue3` and it might turn into a bug 😅

LuisMartinez15, line 274:
```suggestion
      const snippetConfig = vue.observable(this.snippetConfig);
```

packages/x-components/src/x-installer/x-installer/x-installer.ts
```diff
@@ -247,43 +248,35 @@ export class XInstaller {
...
       return new vue({
         ...extraPlugins,
         ...this.options.vueOptions,
-        provide() {
-          return {
-            snippetConfig
-          };
+        provide: {
+          snippetConfig: this.snippetConfig
```
LuisMartinez15, line 279: If you end up refactoring the observed snippet config using the returned object instead.

```suggestion
          snippetConfig
```

packages/x-components/src/x-installer/x-installer/x-installer.ts
```diff
@@ -292,6 +285,25 @@ export class XInstaller {
...
+   *
+   * @param snippetConfig - The snippet config to normalise.
+   * @returns The normalised version of the given snippet config.
+   * @internal
+   */
+  protected normaliseSnippetConfig(
+    snippetConfig: SnippetConfig | Partial<SnippetConfig>
+  ): NormalisedSnippetConfig | Partial<SnippetConfig> {
+    if (snippetConfig.lang) {
+      snippetConfig.uiLang ??= snippetConfig.lang;
```
LuisMartinez15, line 302: TIL about the `??=` operator 🤦🏿

LuisMartinez15 (changes requested):
good job! @mnavarroespinosa suggested a change that I think might be interesting to do now. If you don't think so, let me know and I'll approve the pr.

packages/x-components/src/x-installer/x-installer/x-installer.ts
javieri-empathy: Do you mean `reactive`? Yes, I agree, In vue3 as reactivity is based in Proxy objects they don't mutate the source, however I would still use `this.snippetConfig` rather than a new variable for the same reasons you mentioned.

LuisMartinez15 (approved):

mnavarroespinosa (approved):

PR #716, opened 2022-09-12
feat(x-adapter): clean empty params

mnavarroespinosa:
EX-6959

### DIFF
--- a/packages/x-adapter/src/http-clients/fetch.http-client.ts
+++ b/packages/x-adapter/src/http-clients/fetch.http-client.ts
@@ -1,4 +1,4 @@
-import { Dictionary, flatObject } from '@empathyco/x-utils';
+import { Dictionary, cleanEmpty, flatObject } from '@empathyco/x-utils';
 import { HttpClient } from './types';
 import { buildUrl, toJson } from './utils';
 
@@ -18,8 +18,8 @@ export const fetchHttpClient: HttpClient = (
 ) => {
   const signal = cancelable ? { signal: abortAndGetNewAbortSignal(id) } : {};
   const flatParameters = flatObject(parameters);
-  const url = sendParamsInBody ? endpoint : buildUrl(endpoint, flatParameters);
-  const bodyParameters = sendParamsInBody ? { body: JSON.stringify(parameters) } : {};
+  const url = sendParamsInBody ? endpoint : buildUrl(endpoint, cleanEmpty(flatParameters));
+  const bodyParameters = sendParamsInBody ? { body: JSON.stringify(cleanEmpty(parameters)) } : {};
 
   return fetch(url, {
     ...properties,
--- a/packages/x-utils/src/object.ts
+++ b/packages/x-utils/src/object.ts
@@ -100,23 +100,28 @@ export function cleanUndefined<T>(obj: T): T {
  * - `null`.
  * - an empty string.
  * - an empty array.
+ * - an empty object.
  *
  * @param obj - The object from whom pick the values.
  * @returns A new object with the not empty properties of the source object.
  * @public
  */
-export function cleanEmpty<T extends Record<string, unknown>>(obj: T): T {
+export function cleanEmpty<SomeObject extends Record<string, unknown>>(
+  obj: SomeObject
+): SomeObject {
   return reduce(
     obj,
     (pickedObject, key, value) => {
       if (isObject(value)) {
-        pickedObject[key] = cleanEmpty(value);
+        if (Object.keys(value).length > 0) {
+          pickedObject[key] = cleanEmpty(value);
+        }
       } else if (value !== null && value !== '' && !(isArray(value) && value.length === 0)) {
         pickedObject[key] = value;
       }
       return pickedObject;
     },
-    {} as T
+    {} as SomeObject
   );
 }
 
--- a/packages/x-utils/src/object.ts
+++ b/packages/x-utils/src/object.ts
@@ -112,9 +112,11 @@ export function cleanEmpty<SomeObject extends Record<string, unknown>>(
   return reduce(
     obj,
     (pickedObject, key, value) => {
+      // FIXME: Clean nested empty arrays too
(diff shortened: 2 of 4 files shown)


CachedaCodes (approved):

CachedaCodes:
gj! 👍

packages/x-utils/src/object.ts
```diff
@@ -94,6 +94,32 @@ export function cleanUndefined<T>(obj: T): T {
...
+ *
+ * @param obj - The object from whom pick the values.
+ * @returns A new object with the not empty properties of the source object.
+ * @public
+ */
+export function cleanEmpty<T extends Record<string, unknown>>(obj: T): T {
+  return reduce(
+    obj,
+    (pickedObject, key, value) => {
+      if (isObject(value)) {
```
javieri-empathy, line 112: Should we clean also empty objects? It is not specified in the task, but i think it is the same case as the empty array. Doing that would also complicate the logic a little bit (one more `if`), as you mind end up removing all the properties from a child object, transforming it into an empty object that should be removed then.

I guess that is not needed for the http-client use case, as it flattens the parameters for the GET requests

packages/x-utils/src/object.ts
```diff
@@ -94,6 +94,32 @@ export function cleanUndefined<T>(obj: T): T {
...
+ * - `undefined`.
+ * - `null`.
+ * - an empty string.
+ * - an empty array.
+ *
+ * @param obj - The object from whom pick the values.
+ * @returns A new object with the not empty properties of the source object.
+ * @public
+ */
+export function cleanEmpty<T extends Record<string, unknown>>(obj: T): T {
```
javieri-empathy, line 108: I know this file is completely wrong in terms of generic naming, but given our contribution guide:

> - Generic values must have a meaningful name.

```ts
// ❌ Wrong
type Dictionary<T> = Record<string, T>;

// ✅ Good
type Dictionary<Value> = Record<string, Value>;
```

javieri-empathy: Also, should we type this function better? At least to say that it won't return null/undefined values:

```ts
export type ExcludeEmptyProperties<Something> = {
  [Key in keyof Something as RemoveEmpty<Something[Key]> extends never
    ? never
    : Key]: Something extends Record<string, unknown>
    ? ExcludeEmptyProperties<Something[Key]>
    : Exclude<Something[Key], '' | [] | undefined | null>;
};

type RemoveEmpty<Something> = Exclude<Something, '' | [] | undefined | null>;
```

javieri-empathy (changes requested):

packages/x-utils/src/object.ts
mnavarroespinosa: I have updated the generic naming but, as we agreed, the function typing will remain the same, as it got complicated to exclude null/undefined values from the return type.

packages/x-utils/src/object.ts
mnavarroespinosa: Sure, I totally agree! I've added the check so the empty objects are also cleaned.

packages/x-adapter/src/http-clients/fetch.http-client.ts
```diff
@@ -18,8 +18,8 @@ export const fetchHttpClient: HttpClient = (
 ) => {
   const signal = cancelable ? { signal: abortAndGetNewAbortSignal(id) } : {};
   const flatParameters = flatObject(parameters);
-  const url = sendParamsInBody ? endpoint : buildUrl(endpoint, flatParameters);
-  const bodyParameters = sendParamsInBody ? { body: JSON.stringify(parameters) } : {};
+  const url = sendParamsInBody ? endpoint : buildUrl(endpoint, cleanEmpty(flatParameters));
+  const bodyParameters = sendParamsInBody ? { body: JSON.stringify(cleanEmpty(parameters)) } : {};
```
mnavarroespinosa, line 22: @LuisMartinez15 suggested maybe doing the clean of the parameters first, so `cleanEmpty` is called just once: 
```
const cleanedParameters = cleanEmpty(parameters);
const flatParameters = flatObject(cleanedParameters);
const url = sendParamsInBody ? endpoint : buildUrl(endpoint, flatParameters);
const bodyParameters = sendParamsInBody ? { body: JSON.stringify(cleanedParameters) } : {};
```
In this case, we would have a different result for objects like this:
```
{
   q: 'hey',
   whatever: 'hola',
   extraParams: {
      whatever: ''
   }
}
```
In one case, we will have in `flatParameters` an object containing `whatever: 'hola'`, and in the other, `whatever` would be completely removed. 

@javieri-empathy @CachedaCodes which one do you think is the desired behavior?

CachedaCodes (dismissed):

packages/x-adapter/src/http-clients/fetch.http-client.ts
javieri-empathy: Hmm good catch!. After thinking about this and talking with you, I would leave it as it is (1st flat -> 2nd clean). If you use extra params to override a harcoded param of the request you have to accept the full consequences. If you provide an empty extra param property value it should remove the default parameter.

Checking this, we spotted that `flatObject` has a weird behavior. Depending on how you create the object, the order the keys are inserted, it will return different results. I have created a [task to check this.](https://searchbroker.atlassian.net/browse/EX-7019)

javieri-empathy (changes requested):

packages/x-utils/src/object.ts
mnavarroespinosa: Good point! I added it and also a comment to fix the function to clean nested empty arrays too in the future.

packages/x-utils/src/object.ts
javieri-empathy: Yep, my bad, I was blinded by the pleasure of writing TS types

javieri-empathy (approved):

CachedaCodes (approved):

PR #717, opened 2022-09-13
feat(facets): remove filters on new query

tiborux:
- Moved the util, `is-new-query,` to the base utils folder. (this part has not to be reviewed)

- Added a new event `FacetsQueryChanged` that is going to be emitted when the query changes. I found a problem when you add a preselected filter. If you do that, the first time that you open X and search for something, the preselected will be removed. To avoid that, I added a new property `status` to know if we are searching the first query or not. We will use this property to clear the filters when the query changes but only after the first check. (I dont know if we should refactor this property to be a boolean because right now its like a state machine).

- I was checking the util that we used in the history queries, but its not the same, maybe we can refactor both to be the same but I think that we should do it in another task.

EX-6196


packages/x-components/src/x-modules/facets/wiring.ts
```diff
@@ -110,6 +124,12 @@ export const facetsWiring = createWiring({
   FacetsGroupProvided: {
     setFacetsGroupWire
   },
+  UserAcceptedAQuery: {
+    setQuery
+  },
+  FacetsQueryChanged: {
+    clearFiltersOnStatusWire
```
ajperezbau, line 131: Maybe we can get rid of the status prop in the module if we clear the filters only if the facets had already a query.

To know that we can get the old value accessing the metadata argument. This is useful to mantain the pre-selected filters.

```suggestion
    clearAllFilters: filter(clearAllFiltersWire, ({ metadata }) => !!metadata.oldValue)
```

ajperezbau (changes requested):
Good job 👌

packages/x-components/src/x-modules/facets/wiring.ts
tiborux: good catch! I think that it is a better solution

CachedaCodes (changes requested):

packages/x-components/src/x-modules/facets/store/module.ts
```diff
@@ -12,6 +12,8 @@ import { FacetGroupEntry, FacetsXStoreModule } from './types';
  */
 export const facetsXStoreModule: FacetsXStoreModule = {
   state: () => ({
+    query: '',
+    status: 'initial',
```
ajperezbau, line 16: I think this status can be removed

packages/x-components/src/x-modules/facets/store/module.ts
```diff
@@ -47,6 +49,9 @@ export const facetsXStoreModule: FacetsXStoreModule = {
     },
     setFacet(state, facet: Facet) {
       Vue.set(state.facets, facet.id, facet);
+    },
+    setQuery(state, query: string) {
```
ajperezbau, line 53: I think the type here can be omitted, because it is inferred from the `FacetsMutations` interface.

```suggestion
    setQuery(state, query) {
```

ajperezbau (changes requested):
Nice 👌

CachedaCodes (dismissed):

packages/x-components/tests/e2e/search-box/search-box.spec.ts
```diff
@@ -5,7 +5,7 @@ let resultsCount = 0;
 let resultsList: string[] = [];
 const compoundResultsList: string[] = [];
 let startQuery = 0;
-let startSecondQuery = 0;
+const startSecondQuery = 0;
```
mnavarroespinosa, line 8: I think this variable can be removed along with its usage.

mnavarroespinosa (commented):
gj!

mnavarroespinosa (approved):
gj!

CachedaCodes (approved):

PR #725, opened 2022-09-19
feat: Improve tailwindcss types.

tajespasarela:
EX-7041


packages/x-components/src/components/sliding-panel.vue
```diff
@@ -173,11 +173,13 @@
      * @internal
      */
     protected updateScrollPosition(): void {
-      const { scrollLeft, clientWidth, scrollWidth } = this.$refs.scrollContainer;
-      this.isScrollAtStart = !scrollLeft;
-      /* The 2 px extra is to fix some cases in some resolutions where the scroll + client size is
-       *  less than the scroll width even when the scroll is at the end */
-      this.isScrollAtEnd = scrollLeft + clientWidth + 2 >= scrollWidth;
+      if (this.$refs.scrollContainer !== undefined) {
```
tiborux, line 176: was it added on purpose?

tiborux (approved):

packages/x-tailwindcss/src/types.ts
```diff
@@ -90,20 +91,7 @@ export type DynamicCssStylesOptions = Dictionary<{
...
-    components: Dictionary<(value: any) => Dictionary>,
-    options: { values: any }
-  ) => void;
-  addUtilities: (cssStyleOptions: CssStyleOptions) => void;
-  matchUtilities: (
-    utilities: Dictionary<(value: any) => Dictionary>,
-    options: { values: any }
-  ) => void;
-};
+export type TailwindHelpers = PluginAPI;
```
LuisMartinez15, line 94: I know that preserving the alias results in less changes, but should we just use `PluginAPI` directly?

LuisMartinez15 (approved):
good job!

packages/x-tailwindcss/src/types.ts
tajespasarela: Because it is not a "Public" type. Tailwind is not exporting it directly. So if they change this type or its location, with this alias, we only have to change in one place

tiborux (approved):

PR #730, opened 2022-09-21
feat: Improve tailwindcss types.

tajespasarela:
EX-7041

Little things forgotten in the other PR 🙏🏻

### DIFF
--- a/packages/x-tailwindcss/src/types.ts
+++ b/packages/x-tailwindcss/src/types.ts
@@ -1,9 +1,11 @@
 import { DeepPartial, Dictionary } from '@empathyco/x-utils';
 import { PluginAPI } from 'tailwindcss/types/config';
+import { Config } from 'tailwindcss';
 import { ReturnOfComponents } from './x-tailwind-plugin/components';
 import { ReturnOfDynamicComponents } from './x-tailwind-plugin/dynamic-components';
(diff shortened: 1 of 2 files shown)


packages/x-tailwindcss/src/types.ts
```diff
@@ -1,9 +1,11 @@
 import { DeepPartial, Dictionary } from '@empathyco/x-utils';
 import { PluginAPI } from 'tailwindcss/types/config';
+import { Config } from 'tailwindcss';
```
mnavarroespinosa, line 3: We could import both from `tailwindcss/types/config`, or it was intended to import it separately? 
```suggestion
import { PluginAPI, Config } from 'tailwindcss/types/config';
```

LuisMartinez15 (approved):

packages/x-tailwindcss/src/types.ts
tajespasarela: the `PluginAPI` is not "publicly" exported by tailwind, I prefer to keep it separately if the path or type change in the future. And as the Config is publicly exported, we can use it from the default barrel, and we know it will not change, unless it is a breaking change.

mnavarroespinosa (approved):

LuisMartinez15 (approved):

PR #738, opened 2022-09-26
feat(design-system): Add `button` component to the Design System.

tajespasarela:
[EX-6797]

This PR adds the `button` component and its variants to the new Design System.
At the moment, It is not used anywhere else but in the demo of the `x-tailwindcss` package.

[EX-6797]: https://searchbroker.atlassian.net/browse/EX-6797?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/x-tailwindcss/demo/options.ts
```diff
@@ -61,12 +62,7 @@ export const options: PluginOptions = {
     return {};
   },
   theme: {
-    colors: {
-      neutral: {
-        10: 'red',
-        20: 'orange'
-      }
-    },
+    colors: {},
```
tajespasarela, line 65: @javieri-empathy if you can take a look into this types. If you add:
```js
   '.button': {
      '&.outlined': {
        '&.primary': {}
      }
    }
```
to this options, `.button` is autocomplete, `&.outlined`, is autocomplete but `&.primary` not, event the function `mapColors` has the right type.

I think it is because the `DeepPartial` in the type `ComponentsDefinition`, but I'm not sure.

tajespasarela, line 65: Here you can test to override css like this:
```js
      '.button': {
        '&.outlined': {
          '&.primary': {}
        }
      }
```

@javieri-empathy if you can take a look into the type, the `&.primary` is not inferred, like it does with the `&.outlined`, despite of the `mapColors` function has the right type. I think is because the `DeepPartial` in the `ComponentsDefinition` type, but I'm not sure

packages/x-tailwindcss/demo/options.ts
```diff
@@ -7,7 +7,8 @@ import {
 } from '../src/types';
 
 export const options: PluginOptions = {
-  components({ theme }): ComponentsDefinition {
+  components(helpers): ComponentsDefinition {
+    const { theme } = helpers;
     return {
       '.icon': {
         display: 'flex',
```
tajespasarela, line 14: Here you can test to override css like this:
```js
      '.button': {
        '&.outlined': {
          '&.primary': {}
        }
      }
```

@javieri-empathy if you can take a look into the type, the `&.primary` is not inferred, like it does with the `&.outlined`, despite of the `mapColors` function has the right type. I think is because the `DeepPartial` in the `ComponentsDefinition` type, but I'm not sure

annacv (approved):
Although it's not time to focus on this now, note that in the squared and circle versions of the button the "heart no-icon" would be visually more vertically centered by using a line-height of 1. Let's keep this in mind when adding the icon component to see if the same thing happens (& see if it's the desired behavior).

packages/x-tailwindcss/demo/options.ts
javieri-empathy: And just by doing 2 small apparently harmless changes, the autocomplete suggestions work properly:

Use a bracketed access to define `ThemeColors`:

```ts
export type ThemeColors = typeof pluginTheme['colors'];
```

Use `Object.assign` to define `buttonOutlined`:

```ts
export function buttonOutlined(helpers: TailwindHelpers) {
  const { theme } = helpers;
  return {
    '&.outlined': Object.assign(
      {
        borderColor: theme('colors.neutral.50'),
        backgroundColor: theme('colors.neutral.0'),
        color: theme('colors.neutral.50'),

        '&:hover': {
          borderColor: theme('colors.neutral.50'),
          backgroundColor: theme('colors.neutral.50'),
          color: theme('colors.neutral.0')
        },

        '&:active': {
          borderColor: theme('colors.neutral.50'),
          backgroundColor: theme('colors.neutral.50'),
          color: theme('colors.neutral.0')
        },

        '&:disabled': {
          borderColor: theme('colors.neutral.25'),
          backgroundColor: theme('colors.neutral.10'),
          color: theme('colors.neutral.25')
        }
      },
      mapColors(
        color => ({
          borderColor: color['50'],
          backgroundColor: theme('colors.neutral.0'),
          color: color['50'],

          '&:hover': {
            backgroundColor: color['50'],
            borderColor: color['50'],
            color: theme('colors.neutral.0')
          },

          '&:active': {
            borderColor: color['50'],
            backgroundColor: color['50'],
            color: theme('colors.neutral.0')
          },

          '&:disabled': {
            borderColor: theme('colors.neutral.25'),
            backgroundColor: theme('colors.neutral.10'),
            color: theme('colors.neutral.25')
          }
        }),
        helpers
      )
    )
  };
}
```

I couldn't find any difference between the types using these new options, so my guess is that either Webstorm is doing something weird under the hood, or TS has a separate process for suggestions that does not follow the types one
image
image

packages/x-tailwindcss/src/utils/map-colors.ts
```diff
@@ -0,0 +1,66 @@
...
+  10?: string;
+  25: string;
+  50: string;
+  75: string;
+  90?: string;
+  100?: string;
+}
+
+/**
+ * This functions takes all the Theme colors and maps them to custom CSS defined byt the mapperFn.
```
CachedaCodes, line 24:
```suggestion
 * This functions takes all the Theme colors and maps them to custom CSS defined by the mapperFn.
```

diegopf, line 24:
```suggestion
 * This function takes all the Theme colors and maps them to custom CSS defined by the mapperFn.
```

packages/x-tailwindcss/src/x-tailwind-plugin/components/button/default.ts
```diff
@@ -0,0 +1,56 @@
...
+    borderColor: theme('colors.neutral.50'),
+    backgroundColor: theme('colors.neutral.50'),
+    color: theme('colors.neutral.0'),
+
+    fontFamily: theme('fontFamily.primary'),
+    fontSize: theme('fontSize.md'),
+    fontWeight: theme('fontWeight.bold'),
+    letterSpacing: theme('letterSpacing.md'),
+    lineHeight: theme('lineHeight.sm'),
+
```
diegopf, line 36: I'm wondering if we should add the `cursor: default` as it is specified on figma. I think during the pairing we discussed about it and agreed on writing down every attribute from figma?

diegopf (commented):
@tajespasarela What about the button link appearance from figma? are we skipping it for the moment?

packages/x-tailwindcss/src/utils/map-colors.ts
tajespasarela: done

tajespasarela: done

packages/x-tailwindcss/src/x-tailwind-plugin/components/button/default.ts
tajespasarela: ok, added

tajespasarela:
> @tajespasarela What about the button link appearance from figma? are we skipping it for the moment?

The link button will "depend" on how the typography components will be. Then we wait until we have that to have all the information

annacv (approved):
🚀

packages/x-tailwindcss/src/x-tailwind-plugin/components/button/default.ts
diegopf: @tajespasarela I think you forgot to push the change? at least I can't see it 😅

tajespasarela: I do see it up:
image

annacv (approved):

packages/x-tailwindcss/src/x-tailwind-plugin/plugin.ts
```diff
@@ -33,7 +33,11 @@ export default plugin.withOptions(
...
   function (options) {
     return {
-      theme: deepMerge({}, theme, options?.theme)
+      theme: deepMerge(
+        { colors: { current: 'currentColor', transparent: 'transparent' } },
+        theme,
+        options?.theme
+      )
     } as Config;
   }
```
diegopf, line 42: Sooner or later I think we should document this file, I'm afraid that I don't completely understand everything going on here.

diegopf (approved):

packages/x-tailwindcss/src/x-tailwind-plugin/plugin.ts
tajespasarela: created subtask for that: https://searchbroker.atlassian.net/browse/EX-7129

PR #731, opened 2022-09-22
Update cypress

JaviIG:

PR #739, opened 2022-09-26
feat: replace `parcel` with `vite`

JaviIG:
`parcel` was giving some issues recently, autoinstalling polyfills such as `process`. As this is only used to run test, and we are already using Vite for the tailwind plugin package, I prpose to add it also there. As you can see in the package-lock, this removes a big chunk of other dependencies, which normally is a good sign.

There is a breaking change in this PR , and it is that it also changes the compilation to ES6. If you find this an issue, we can compile to ES6 only for vite, and ES5 for the build output. But to match the rest of the mono-repo, I would leave it. Note that the customer using this package does not need support anymore for older browsers with ES5.

This is part of a series of PRs to achieve the update of Cypress to 10.8.0. The [global result can be seen here](https://github.com/empathyco/x/pull/731).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 .eslintcache
 .DS_Store
-.parcel-cache
 **/node_modules
 **/dist
 **/docs
--- a/package-lock.json
+++ b/package-lock.json
@@ -76,7 +76,7 @@
         "tslib": "~2.3.0",
         "tslint": "~5.16.0",
         "typescript": "~4.6.2",
-        "vite": "~3.1.3",
+        "vite": "~2.9.0",
         "vue": "~2.6.14",
         "vue-class-component": "~7.1.0",
         "vue-docgen-cli": "~4.13.1",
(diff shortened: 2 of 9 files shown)


annacv (commented):
The build process failed at this point: 
Captura de pantalla 2022-10-21 a les 13 59 21

I could fix it by: 
- Going to `ts.config.json` and enabling library 'dom': `"lib": ["esnext", "dom" ],`

Additionally I resolved conflicts which would update to ` @empathyco/x-tailwindcss@0.2.0-alpha.22` and check it still worked fine.

I didn't commit this as several errors raised in some files  and I didn't feel confident to solve them 🙏

ajperezbau:
> The build process failed at this point: Captura de pantalla 2022-10-21 a les 13 59 21
> 
> I could fix it by:
> 
> * Going to `ts.config.json` and enabling library 'dom': `"lib": ["esnext", "dom" ],`
> 
> Additionally I resolved conflicts which would update to ` @empathyco/x-tailwindcss@0.2.0-alpha.22` and check it still worked fine.
> 
> I didn't commit this as several errors raised in some files and I didn't feel confident to solve them 🙏

These problems seem to be present only when updating `vite` in the `tailwindcss` package.
Should we keep it with `vite 2` for the moment and check this in another task?

ajperezbau (commented):
It is working well, good job 👌 
I don't see any problem in compiling to ES6 if ES5 support is not needed.
Just a question, should the type definitions be included in the dist folder?

javieri-empathy:
> It is working well, good job 👌 I don't see any problem in compiling to ES6 if ES5 support is not needed. Just a question, should the type definitions be included in the dist folder?

In the package.json we specify both the `types` and `dist` folders to be included through the `files` property, so it is not an issue. We do this different in every library and it should be standardised.

LuisMartinez15 (commented):
Good job! We can also remove the `.parcel-cache` entry from the `.gitignore` too if you want.

LuisMartinez15 (approved):

LuisMartinez15 (approved):
good job!

PR #740, opened 2022-09-26
Feature/upgrade ci runners

JaviIG:
This PR updates the CI runners to node 16 and fixes a failing test in node16

EX-6907

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,7 +6,9 @@ concurrency:
 jobs:
   build:
     runs-on: ubuntu-latest
-    container: cypress/browsers:node16.16.0-chrome105-ff104-edge
+    container:
+      image: cypress/browsers:node16.16.0-chrome105-ff104-edge
+      options: --user 1001
     steps:
       - uses: actions/checkout@v3
         with:
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,7 +6,7 @@ concurrency:
 jobs:
   build:
     runs-on: ubuntu-latest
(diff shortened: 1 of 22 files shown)


tajespasarela (changes requested):

.github/workflows/build.yml
```diff
@@ -6,7 +6,9 @@ concurrency:
 jobs:
   build:
     runs-on: ubuntu-latest
-    container: cypress/browsers:node14.17.0-chrome91-ff89
+    container:
+      image: cypress/browsers:node16.16.0-chrome105-ff104-edge
+      options: --user 1001
```
LuisMartinez15, line 11: Why do we need to add this `options: --user 1001` parameter? I didn't find anything related to this in the cypress documentation 😅

LuisMartinez15 (approved):
good job! thanks!

.github/workflows/build.yml
LuisMartinez15: Nvm, I've found it! In case anyone also wonders why, [read](https://github.com/cypress-io/github-action#firefox).

JaviIG: I still don't understand why it is necessary. That's the only reference I found in their docs, asides from some issues[ like this one](https://github.com/cypress-io/cypress-docker-images/issues/560). We are not using Firefox yet, but seems that the image lacks certain permissions when installing?

[This is the error](https://github.com/empathyco/x/actions/runs/3131223359/jobs/5082443562) that gave without that user.
image

LuisMartinez15 (approved):

PR #747, opened 2022-09-27
test: replace `cypress-cucumber-preprocessor` with `@badeball/cypress-cucumber-preprocessor`

JaviIG:
Our current cucumber preprocessor has been abandoned. It has been replaced by [@badeball/cypress-cucumber-preprocessor](https://github.com/badeball/cypress-cucumber-preprocessor/issues/689). This PR is another step [in the Cypress 10.8.0 upgrade](https://github.com/empathyco/x/pull/731).

There are a few important changes
- [`.features` files are not supported](https://github.com/badeball/cypress-cucumber-preprocessor/tree/7031d0283330bca814d6923d35d984224622b4cf#bundled-features-files). This was a performance improvement in previous versions that avoided closing/opening the browser. We will have to run the CI a few times to get the average test time to know if it is affecting us. If it is, knowing how ESBuild handles plugins I think creating the plugin ourselves should have a mid complexity.
- This new preprocessor does not support `And` test implementations. That's why you will see lots of them replaced by (hopefully) their correct version (Given, When, Then).
- The blocks implementation has its own `this` type now, which is `Mocha.Context`. I would still need more time to dig more into this types.
- Now it uses ESBuild as a preprocessor. Previously we used either webpack or browserify. As ESBuild is used by Vite and we want to get rid of webpack, I think it is great news.

Applying to EX-6240


packages/x-components/tests/support/index.ts
```diff
@@ -195,25 +198,6 @@ const customCommands: CustomCommands = {
     cy.getByDataTest('loading').should('exist');
     cy.getByDataTest('loading').should('not.exist').then(noOp);
   },
-  fakeSearchResponse: searchResponse => {
```
JaviIG, line 198: This was legacy non used code

packages/x-components/tests/support/index.ts
```diff
@@ -1,5 +1,3 @@
-import 'cypress-plugin-tab';
```
JaviIG, line 1: Tab plugin was not used.

JaviIG:
I believe this right now is failing due to parcel trying to install the process polyfill. This is solved by the [replace parcel by vite PR](https://github.com/empathyco/x/pull/739)

herrardo:
Applying to this task: EX-6240

mnavarroespinosa (commented):
Looks good to me! 

Just one question, we could remove the `All.features` file as it is not used anymore, isn't it?

ajperezbau (commented):
Looks good to me 👌

tiborux (dismissed):
gj

LuisMartinez15 (approved):
amazing job!

javieri-empathy:
I've just updated the branch to be merged in an aux branch. Apparently the Cypress version we are using does not have support for Apple silicon architecture and is run inside of Rosetta. As node and npm do support Apple Silicon processors, esbuild installs the binary for them, but when run together with Cypress, it is run inside of Rosetta, what makes it try to use the darwin version and fails.
This will be fixed by upgrading cypress in the next PR.

ajperezbau (approved):
Nice, it should be a good performance boost using the native ARM version of Cypress 👀

PR #748, opened 2022-09-28
fix(base-modal): add prop to allow the modal to move the focus to an element within it on open

CachedaCodes:
EX-6755

## Motivation and context
When opening modals (namely the `MyHistory` one) the focus could still be behind the modal, which is weird.

## Type of change

- [X] `Main`
- [ ] Other. Specify:

## How has this been tested?
There are new unit tests for the new prop in the `base-modal` component.

### DIFF
--- a/packages/x-components/src/components/modals/__tests__/base-modal.spec.ts
+++ b/packages/x-components/src/components/modals/__tests__/base-modal.spec.ts
@@ -96,8 +96,10 @@ describe('testing Base Modal  component', () => {
 
   it('changes the focus to the correct element when the modal opens', async () => {
     let { wrapper, setOpen, appendToBody } = mountBaseModal({
-      // eslint-disable-next-line max-len
-      defaultSlot: `<div><button data-test="expected-focus">First button</button><button>Second button</button></div>`,
+      defaultSlot: `<div>
+          <button data-test="expected-focus">First button</button>
+          <button>Second button</button>
+        </div>`,
       open: false
     });
 
@@ -109,8 +111,10 @@ describe('testing Base Modal  component', () => {
     );
 
     ({ wrapper, setOpen, appendToBody } = mountBaseModal({
-      // eslint-disable-next-line max-len
-      defaultSlot: `<div><button>First button</button><button tabindex="1" data-test="expected-focus">Second button</button></div>`,
+      defaultSlot: `<div>
+          <button>First button</button>
+          <button tabindex="1" data-test="expected-focus">Second button</button>
+        </div>`,
       open: false
     }));
 
@@ -122,9 +126,12 @@ describe('testing Base Modal  component', () => {
     );
   });
 
-  it("doesn't changes the focus if the focusOnOpen prop is false", async () => {
+  it("doesn't change the focus if the focusOnOpen prop is false", async () => {
     const { setOpen, appendToBody } = mountBaseModal({
-      defaultSlot: `<div><button>First button</button><button>Second button</button></div>`,
+      defaultSlot: `<div>
+          <button tabindex="1">First button</button>
+          <button>Second button</button>
+        </div>`,
       open: false,
       focusOnOpen: false
     });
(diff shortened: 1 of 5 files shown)


packages/x-components/src/components/modals/__tests__/base-modal.spec.ts
```diff
@@ -87,13 +93,57 @@ describe('testing Base Modal  component', () => {
...
+    appendToBody();
+    await setOpen(true);
+
+    expect(wrapper.find(getDataTestSelector('expected-focus')).element).toBe(
+      document.activeElement
+    );
+  });
+
+  it("doesn't changes the focus if the focusOnOpen prop is false", async () => {
+    const { setOpen, appendToBody } = mountBaseModal({
```
annacv, line 126:
```suggestion
    const { wrapper, setOpen, appendToBody } = mountBaseModal({
```

packages/x-components/src/components/modals/__tests__/base-modal.spec.ts
```diff
@@ -87,13 +93,57 @@ describe('testing Base Modal  component', () => {
...
+      defaultSlot: `<div><button>First button</button><button>Second button</button></div>`,
```
annacv, line 127: (idem, below)
```suggestion
      defaultSlot: `<div><button tabindex="1" data-test="expected-focus">First button</button><button>Second button</button></div>`,
```

packages/x-components/src/components/modals/__tests__/base-modal.spec.ts
```diff
@@ -87,13 +93,57 @@ describe('testing Base Modal  component', () => {
...
+      open: false,
+      focusOnOpen: false
+    });
+
+    appendToBody();
+    const focusedElementBeforeOpen = document.activeElement;
+    await setOpen(true);
+
```
annacv, line 135: I think this would reinforce the test, as makes clear that setting `focusOnOpen: false` will prevail over the element with `tabindex="1"`
```suggestion
expect(wrapper.find(getDataTestSelector('expected-focus')).element).not.toBe(
      document.activeElement
    );
```

packages/x-components/src/components/modals/__tests__/base-modal.spec.ts
```diff
@@ -87,13 +93,57 @@ describe('testing Base Modal  component', () => {
 
     expect(wrapper.find(getDataTestSelector('default-slot-overridden')).exists()).toBe(true);
   });
+
+  it('changes the focus to the correct element when the modal opens', async () => {
+    let { wrapper, setOpen, appendToBody } = mountBaseModal({
+      // eslint-disable-next-line max-len
+      defaultSlot: `<div><button data-test="expected-focus">First button</button><button>Second button</button></div>`,
```
annacv, line 100: ...a bit silly suggestion, but for newbies like me, it would be faster to read this way..., up to you if you want to change this everywhere 🙊 🙊 
```suggestion
      defaultSlot: `<div>
                               <button data-test="expected-focus">First button</button>
                               <button>Second button</button>
                            </div>`,
```

annacv (changes requested):
🚀

LuisMartinez15 (changes requested):

LuisMartinez15:
good job!

packages/x-components/src/components/modals/__tests__/base-modal.spec.ts
CachedaCodes: I can see the adding `tabindex` adds to the clarity of the test, but I don't see this assertion being much useful.

We're already asserting that the focused element before and after opening the modal is the same one, so I find that adding another assertion checking if a different element is indeed not focused is redundant.

annacv (approved):

LuisMartinez15 (approved):

packages/x-components/src/utils/focus.ts
```diff
@@ -0,0 +1,7 @@
+/**
+ * Comma separated list of common focusable selectors.
+ */
```
LuisMartinez15, line 3: The build is now failing due to this file being exported and this doc missing a `release tag`. Can you add the `@internal` tag to it? I think it should be `@internal`, but if you think otherwise feel free to say it.

```suggestion
/**
 * Comma separated list of common focusable selectors.
 *
 * @internal
 */
```

LuisMartinez15 (changes requested):

LuisMartinez15 (approved):

annacv (approved):

PR #749, opened 2022-09-28
ci: silent jest and eslint warnings

JaviIG:
Avoids jest from printing console statements (only our logs and in jest for the CI). This should avoid noise and speed up the tests as logging is suprisingly expensive.


LuisMartinez15 (approved):
good job! thanks!

tajespasarela (approved):

PR #746, opened 2022-09-27
Add origin to Next Queries Preview

annacv:
[EX-6757](https://searchbroker.atlassian.net/browse/EX-6757)

# Pull request template
 
- Modify `fetchAndSaveNextQueryPreviewWire` to send both the `query` and the `location` in the request.
- Update store `fetchAndSaveNextQueryPreview`and `fetchNextQueryPreview` actions & its types.
- Update tests.
- Create a test for an undefined location use case

## Motivation and context

Enable sending an `origin` for this feature to be tracked.

- [ ] Dependencies. If any, specify:
- [ ] Open issue. If applicable, link:

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that causes existing functionality to not work as expected)
- [ ] Change requires a documentation update

## What is the destination branch of this PR?

- [x] `Main`
- [ ] Other. Specify:

## How has this been tested?

Tests performed according to [testing guidelines](./contributing/tests.md): 

## Checklist:

- [x] My code follows the **[style guidelines](./CONTRIBUTING.md#style-guides)** of this project.
- [x] I have performed a **self-review** on my own code.
- [x] I have **commented** my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the **[documentation](./CONTRIBUTING.md#documentation-style-guide)**.
- [x] My changes generate **no new warnings**.
- [x] I have added **tests** that prove my fix is effective or that my feature works.
- [x] New and existing **unit tests pass locally** with my changes.

PR #750, opened 2022-09-29
feat: make adapter configurable in home view

diegopf:
EX-7073

The `e2eAdapter` can't be used outside cypress directly. This adapter makes requests that cypress will intercept and resolve with mocked data. In order to workaround this issue, in this PR I prevent the `e2eAdapter` from using the [mockedFetchHttpClient](https://github.com/empathyco/x/blob/feature/EX-7072-make-adapter-configurable-in-home-view/packages/x-components/src/adapter/e2e-adapter.ts#L12) if we are not in a cypress environment and return the mocked data instead.

Note: if you see an error regarding the `SlidingQueryPreview` complaining about the missing location, this will be fixed in #746

### DIFF
--- a/packages/x-components/src/adapter/e2e-adapter.ts
+++ b/packages/x-components/src/adapter/e2e-adapter.ts
@@ -47,7 +47,8 @@ export const e2eAdapter: XComponentsAdapter = {
   tagging: endpointAdapterFactory({
     endpoint: ({ url }) => url,
     requestMapper: ({ params }) => params,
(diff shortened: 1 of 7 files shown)


CachedaCodes (changes requested):

LuisMartinez15 (dismissed):
good job!

CachedaCodes (approved):

LuisMartinez15 (approved):

PR #752, opened 2022-09-30
Feature/ex 7049 debounce query preview requests

javieri-empathy:
[EX-7049](https://searchbroker.atlassian.net/browse/EX-7049)

Adds a debounce to the query preview requests. This makes the component usable for use cases such as search preview, where it would receive the input query, and trigger a small request as a preview.

### DIFF
--- a/packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
+++ b/packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
@@ -48,7 +48,6 @@
   import { createOrigin } from '../../../utils/origin';
   import { debounce } from '../../../utils/debounce';
   import { DebouncedFunction } from '../../../utils';
-  import { XEmit } from '../../../components';
 
   /**
    * Retrieves a preview of the results of a query and exposes them in the default slot,
--- a/packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
+++ b/packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
@@ -48,6 +48,7 @@
   import { createOrigin } from '../../../utils/origin';
   import { debounce } from '../../../utils/debounce';
   import { DebouncedFunction } from '../../../utils';
+  import { XEmit } from '../../../components';
 
   /**
    * Retrieves a preview of the results of a query and exposes them in the default slot,
@@ -73,8 +74,8 @@
     })
     protected query!: string;
 
-    /**.
-     *  The origin property for the request
+    /**
+     * The origin property for the request.
      *
      * @public
      */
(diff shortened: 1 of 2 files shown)


packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
```diff
@@ -153,8 +161,54 @@
...
+    }
+
+    /**
+     * Initialises watcher to emit debounced requests, and first value for the requests.
+     *
+     * @internal
+     */
+    protected created(): void {
+      this.$watch(
+        () => this.queryPreviewRequest,
```
CachedaCodes, line 183: Any reason to use the function form instead of `'queryPreviewRequest'`?

packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
```diff
@@ -153,8 +161,54 @@
...
+        request => this.emitQueryPreviewRequestChanged(request)
+      );
+      this.emitQueryPreviewRequestChanged(this.queryPreviewRequest);
+    }
+
+    /**
+     * Cancels the (remaining) requests when the component is destroyed
+     * via the debounce.cancel() method.
```
CachedaCodes, line 191: Can we change this line to wrap `debounce.cancel()` as code for the docs? The IDE doesn't like it.
```suggestion
     * via the `debounce.cancel()` method.
```

packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
javieri-empathy: Not losing type safety 🤣:

image

packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
```diff
@@ -153,8 +161,54 @@
...
+     * via the `debounce.cancel()` method.
+     *
+     * @internal
+     */
+    protected beforeDestroy(): void {
+      this.emitQueryPreviewRequestChanged.cancel();
+    }
+
+    /**
+     * Cancels the previous request when there is a new one.
```
ajperezbau, line 200: Please, modify the description of this method to make it clear that it will be executed when the debounced function changes (e.g., when the debounce time prop changes)

packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
```diff
@@ -130,16 +139,15 @@
...
      * @returns The search request object.
-     * @public
+     * @internal
      */
-    @XEmit('QueryPreviewRequestChanged', { immediate: false })
-    public get queryPreviewRequest(): SearchRequest {
+    protected get queryPreviewRequest(): SearchRequest {
       const origin = createOrigin({
         feature: this.queryFeature,
         location: this.location
```
ajperezbau, line 153: It is not related to this PR, but If a location is not provided by an ancestor, an error is logged in the console.
Also, should the location be a prop instead of an injected value, as well as with the queryFeature?

annacv: Hi! Good catch! The Injection needed a default value not to warn. On the other hand, the location is managed via the  provide/inject logic (👀 LocationProvider component)

ajperezbau (approved):
Good job 👌

CachedaCodes (approved):

PR #754, opened 2022-10-03
fix: Initial immediate requests now wait for the extra params to be p…

javieri-empathy:
…rovided.

EX-6977

Prevents initial request (Popular searches and recommendations) to be fired before the extra params have been loaded in all modules.

The issue was that we were using a `Promise.resolve` (microtask queue) for the `immediate` emitters instead of the debounced emit function, which may delay the emission (and cancel previous attempts) to try to group as maximum changes as possible  in one emission by using the macrotask queue (`setTimeout`).

### DIFF
--- a/packages/x-components/src/plugins/__tests__/x-plugin-emitters.spec.ts
+++ b/packages/x-components/src/plugins/__tests__/x-plugin-emitters.spec.ts
@@ -156,7 +156,7 @@ describe('testing X Plugin emitters', () => {
     };
 
     // eslint-disable-next-line max-len
-    it('should not execute wires with immediate `false` when the module is registered', async () => {
+    it('should not execute wires with immediate `false` when the module is registered', () => {
       const pluginOptions: XPluginOptions = {
         adapter: XComponentsAdapterDummy,
         xModules: {
@@ -180,12 +180,12 @@ describe('testing X Plugin emitters', () => {
 
       /* Emitters relies on Vue watcher that are async. We need to wait a cycle before testing if
          they have emitted or not. */
-      await Promise.resolve();
+      jest.runAllTimers();
 
       expect(testWire).not.toHaveBeenCalled();
     });
 
-    it('should execute wires with immediate `true` when the module is registered', async () => {
+    it('should execute wires with immediate `true` when the module is registered', () => {
       const pluginOptions: XPluginOptions = {
         adapter: XComponentsAdapterDummy,
         xModules: {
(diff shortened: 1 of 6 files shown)


packages/x-components/src/plugins/__tests__/x-plugin-emitters.spec.ts
```diff
@@ -211,7 +211,7 @@ describe('testing X Plugin emitters', () => {
 
       /* Emitters relies on Vue watcher that are async. We need to wait a cycle before testing if
        they have emitted or not. */
-      await Promise.resolve();
+      jest.advanceTimersByTime(0);
```
javieri-empathy, line 214: As we are testing a `XXXChanged` event, it is only debounced once. So running all the timers once is enough for the test

packages/x-components/tests/e2e/common/common-steps.spec.ts
```diff
@@ -248,19 +248,73 @@ Then(
   (key: string, value: string) => {
     cy.wait('@interceptedResults')
       .its('request.body')
-      .then(JSON.parse)
-      .should('have.property', key, value === 'default' ? '' : value);
+      .should((body: string) => {
+        expect(JSON.parse(body)).to.have.property(key, value);
```
javieri-empathy, line 252: Cypress suggests avoiding the `.then` call for transforming values due to retryability. I don't think that really matters here  as we are using a `wait` command ( I thought it did), but now that I see it I like it.

https://docs.cypress.io/guides/core-concepts/retry-ability#Incorrectly-waiting-for-values

packages/x-components/tests/e2e/extra-params.feature
```diff
@@ -1,33 +1,41 @@
...
+    And   a popular searches API with a known response
+    And   a related tags API with a known response
     And   a recommendations API with a known response
+    And   a tracking API
     And   no special config for layout view
     And   start button is clicked
 
-  Scenario Outline: 1. Search request includes extra-params from Snippet Config
-    Given a results API with a known response
+  Scenario Outline: 1. Search request includes extra-params
```
javieri-empathy, line 14: Both tests were testing the same initially, so I decided to join them and extend them to test everything we need.

packages/x-components/src/plugins/x-emitters.ts
```diff
@@ -30,26 +30,18 @@ export function registerStoreEmitters(
...
     });
 
-    store.watch(watcherSelector, debouncedEffect, options);
+    store.watch(watcherSelector, emit, options);
 
     if (immediate) {
-      Promise.resolve().then(() => {
-        emit(watcherSelector());
-      });
+      emit(watcherSelector());
```
javieri-empathy, line 44: Simply by using the same `emit` function that we use with the watcher, we can guarantee that the events are properly delayed in the "queue" system that we have here.

annacv (approved):
No cancelled requests + 
I run build + tests in my environment & everything ok.

ajperezbau (approved):
Nice 👌

PR #755, opened 2022-10-03
ci: enable code scanning

diegopf:
EX-7144

This PR enables code scanning alerts by analysing the repository code to find security vulnerabilities and coding errors. The tool used for the scanning is [CodeQL](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/about-code-scanning-with-codeql) a code analysis engine developed by GitHub.

This new workflow will run on every push to the `main` branch or any PR created against this branch. If an issue is found during a PR check, we will need to handle it within the PR, either by [fixing it or dismissing](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/triaging-code-scanning-alerts-in-pull-requests#viewing-an-alert-on-your-pull-request) the alert.

Besides, it will run periodically every Monday at 00:00 UTC. The alerts will be found in the [code scanning section](https://github.com/empathyco/x/security/code-scanning).

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -30,10 +30,6 @@ jobs:
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
       uses: github/codeql-action/init@v2
-      with:
-        languages: ${{ matrix.language }}
 
     - name: Perform CodeQL Analysis
       uses: github/codeql-action/analyze@v2
-      with:
-        category: "/language:${{matrix.language}}"
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,74 @@
+# For most projects, this workflow file will not need changing; you simply need
+# to commit it to your repository.
+#
+# You may wish to alter this file to override the set of languages analyzed,
+# or to provide custom queries or build logic.
+#
+# ******** NOTE ********
+# We have attempted to detect the languages in your repository. Please check
+# the `language` matrix defined below to confirm you have the correct set of
+# supported CodeQL languages.
+#
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ "main" ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ "main" ]
+  schedule:
+    - cron: '27 23 * * 0'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'javascript' ]
+        # CodeQL supports [ 'cpp', 'csharp', 'go', 'java', 'javascript', 'python', 'ruby' ]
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,43 @@
+# Learn more about CodeQL at https://codeql.github.com/docs/
```
javieri-empathy, line 1: After talking with @diegopf we believe current issues aren't a real problem for now:

- `interpolate` is used to craft endpoint urls. We control the template they work with.
  - If we support SSR at some point this might become a potential problem, but as long as we control the input data it shouldn't be real.
- `cleanMarkdown` is used as part of the build process. The parameter passed is  part of our source code.
- Prototype pollution in the filters/scroll shouldn't be an issue either as we own that data.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,43 @@
...
+    # Runs periodically at 00:00 UTC on Monday
+    - cron: '00 00 * * 1'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
```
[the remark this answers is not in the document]
diegopf: I guess not, they came with the default workflow template and I think I just overlooked them. I will try to remove and see if there is any problem.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,43 @@
...
+      security-events: write
+
+    strategy:
+      fail-fast: false
+      matrix:
+        language: [ 'javascript' ]
```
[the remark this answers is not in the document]
diegopf: Not really, mostly the same as before. It was part of the default configuration GitHub provided. In the docs they say that usually you don't need to edit the default template, so I just removed the parts that was clear did not applied to us (like configure compiled languages). I will remove it

.github/workflows/codeql-analysis.yml
annacv: So, regarding this, could we [exclude](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#avoiding-unnecessary-scans-of-pull-requests) these files from being scanned? And, maybe for the `interpolate` function, we could add a `TODO` for including it once we support SSR.

diegopf: I would rather dismiss the alert than exclude the file. For the `interpolate` issues we can mark them as won't fix and the `clearMarkdown` we can say it is not part of the build.

ajperezbau:
Should we make the build to analyze the external dependencies in `node_modules` folder?

javieri-empathy (approved):

diegopf:
> This is really cool. Should we exclude the tests from being scanned?

Unless we find a good reason for that, I don't think so. I don't see it at the moment.

> Should we make the build to analyze the external dependencies in `node_modules` folder?

I don't think so. We are not the owners of that code so we can't fix any potential issue that this action could find. That would be blocking us instead of helping us.

ajperezbau (approved):

PR #759, opened 2022-10-06
test(next-queries): next queries preview e2e tests

mavmaf:
Add e2e tests for next queries preview, following scenarios created in EX-6847


packages/x-components/tests/e2e/next-queries-preview.feature
```diff
@@ -0,0 +1,70 @@
+Feature: Next queries preview component
```
tiborux, line 1: Should we add a case when you click the query of the next query? Because right now the query can be clicked and set the search query.

tiborux (changes requested):
gj

CachedaCodes (changes requested):

CachedaCodes (approved):

tiborux (approved):

PR #760, opened 2022-10-06
ci: export webpack config from `x-archetype-utils`

diegopf:
EX-7221

After creating https://github.com/empathyco/x-archetype/pull/181 @tajespasarela suggested to move the webpack fix to the `x-archetype-utils` package.  This PR exports the whole configuration for webpack so we are able to update every customer setup easily.

Besides, I introduce the same fix for the `x-components` development build. See https://github.com/empathyco/x-archetype/pull/181 if you need more context.

### DIFF
--- a/packages/x-components/vue.config.js
+++ b/packages/x-components/vue.config.js
@@ -2,6 +2,18 @@ const glob = require('glob');
 
 module.exports = {
   configureWebpack: {
-    entry: [...glob.sync('src/design-system/**/*.scss').map(style => `./${style}`), './src/main.ts']
+    entry: [
+      ...glob.sync('src/design-system/**/*.scss').map(style => `./${style}`),
+      './src/main.ts'
+    ],
+    output: {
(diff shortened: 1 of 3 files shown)


packages/x-components/vue.config.js
```diff
@@ -2,6 +2,18 @@ const glob = require('glob');
 
 module.exports = {
   configureWebpack: {
-    entry: [...glob.sync('src/design-system/**/*.scss').map(style => `./${style}`), './src/main.ts']
+    entry: [
+      ...glob.sync('src/design-system/**/*.scss').map(style => `./${style}`),
+      './src/main.ts'
+    ],
+    output: {
+      devtoolModuleFilenameTemplate: ({ resourcePath, hash, query }) =>
```
diegopf, line 10: Since this is not likely to change very often here. I think it is just better to have the configuration hardcoded here.

annacv (approved):
🚀

tiborux (approved):
gj

PR #771, opened 2022-10-12
Implement animated search input placeholder component

alvarodE:
[EX-7178](https://searchbroker.atlassian.net/browse/EX-7178)

Implements https://github.com/empathyco/x/issues/385.

New SearchInputPlaceholder component added covering the requirements specified on the issue.

Details about the new component can be found directly on its docs.

The result can be tested launching the project (`npm run serve`).

https://user-images.githubusercontent.com/72568818/195317548-6b37c192-91d0-42c1-95b4-e07aee4f4f72.mov

### DIFF
--- a/packages/x-components/src/design-system/components/input-group/default.scss
+++ b/packages/x-components/src/design-system/components/input-group/default.scss
@@ -2,10 +2,12 @@
 
 .x-input-group {
   // layout
+  position: relative;
   display: flex;
   flex-flow: row nowrap;
   align-content: center;
   align-items: center;
+  overflow: hidden;
   box-sizing: border-box;
   min-width: 0;
 
@@ -74,15 +76,22 @@
     flex: 1 1 100%;
     min-width: 0;
     border: none;
+  }
 
-    &::placeholder {
-      // typography
-      color: var(--x-color-text-input-group-placeholder-default);
-      font-family: var(--x-font-family-input-group-placeholder-default);
-      font-size: var(--x-size-font-input-group-placeholder-default);
-      font-weight: var(--x-number-font-weight-input-group-placeholder-default);
-      line-height: var(--x-size-line-height-input-group-placeholder-default);
-    }
+  > .x-input-placeholder {
+    position: absolute;
+    height: 100%;
+    pointer-events: none;
+  }
+
+  > .x-input-placeholder,
+  > .x-input::placeholder {
+    // typography
+    color: var(--x-color-text-input-group-placeholder-default);
+    font-family: var(--x-font-family-input-group-placeholder-default);
+    font-size: var(--x-size-font-input-group-placeholder-default);
+    font-weight: var(--x-number-font-weight-input-group-placeholder-default);
+    line-height: var(--x-size-line-height-input-group-placeholder-default);
   }
 
   > .x-button {
--- a/packages/x-components/src/design-system/components/input-group/default.scss
+++ b/packages/x-components/src/design-system/components/input-group/default.scss
@@ -2,12 +2,10 @@
 
 .x-input-group {
   // layout
-  position: relative;
   display: flex;
   flex-flow: row nowrap;
   align-content: center;
   align-items: center;
-  overflow: hidden;
   box-sizing: border-box;
   min-width: 0;
 
@@ -76,22 +74,16 @@
     flex: 1 1 100%;
     min-width: 0;
     border: none;
-  }
 
-  > .x-input-placeholder {
-    position: absolute;
-    height: 100%;
-    pointer-events: none;
-  }
-
-  > .x-input-placeholder,
-  > .x-input::placeholder {
-    // typography
-    color: var(--x-color-text-input-group-placeholder-default);
-    font-family: var(--x-font-family-input-group-placeholder-default);
-    font-size: var(--x-size-font-input-group-placeholder-default);
-    font-weight: var(--x-number-font-weight-input-group-placeholder-default);
-    line-height: var(--x-size-line-height-input-group-placeholder-default);
+    > .x-input-placeholder,
+    &::placeholder {
+      // typography
+      color: var(--x-color-text-input-group-placeholder-default);
+      font-family: var(--x-font-family-input-group-placeholder-default);
+      font-size: var(--x-size-font-input-group-placeholder-default);
+      font-weight: var(--x-number-font-weight-input-group-placeholder-default);
+      line-height: var(--x-size-line-height-input-group-placeholder-default);
+    }
   }
 
   > .x-button {
--- a/packages/x-components/src/views/home/Home.vue
+++ b/packages/x-components/src/views/home/Home.vue
@@ -97,9 +97,7 @@
                   :instant="controls.searchInput.instant"
                   :instant-debounce-in-ms="controls.searchInput.instantDebounceInMs"
                 />
-                <ClearSearchInput v-if="$x.query.searchBox" aria-label="Clear query">
-                  Clear
-                </ClearSearchInput>
+                <ClearSearchInput aria-label="Clear query">Clear</ClearSearchInput>
                 <SearchButton aria-label="Search" class="x-input-group__action">
                   <SearchIcon />
                 </SearchButton>
--- a/packages/x-components/src/views/home/Home.vue
+++ b/packages/x-components/src/views/home/Home.vue
@@ -64,6 +64,28 @@
         </label>
       </li>
       <li class="x-test-controls__item x-list__item">
+        <label for="nextQueriesPreview.maxItemsToRender">
+          next-queries-preview - maxItemsToRender
+          <input
+            v-model="controls.nextQueriesPreview.maxItemsToRender"
+            id="nextQueriesPreview.maxItemsToRender"
+            type="number"
+            data-test="nq-preview-max-to-render"
+          />
+        </label>
+      </li>
+      <li class="x-test-controls__item x-list__item">
+        <label for="nextQueriesPreview.maxItemsToRender">
+          next-queries-list - showOnlyAfterOffset
+          <input
+            v-model="controls.nextQueriesList.showOnlyAfterOffset"
+            id="nextQueriesList.showOnlyAfterOffset"
+            type="checkbox"
+            data-test="nq-preview-show-after-offset"
+          />
+        </label>
+      </li>
+      <li class="x-test-controls__item x-list__item">
         <label for="adapter.e2eAdapter">
           Use mocked adapter
           <input
@@ -241,7 +263,9 @@
               <ResultsList v-infinite-scroll:main-scroll>
                 <BannersList>
                   <PromotedsList>
-                    <NextQueriesList :show-only-after-offset="true">
+                    <NextQueriesList
+                      :show-only-after-offset="controls.nextQueriesList.showOnlyAfterOffset"
+                    >
                       <BaseVariableColumnGrid :animation="resultsAnimation">
                         <template #result="{ item: result }">
                           <MainScrollItem :item="result">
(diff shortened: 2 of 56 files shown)


packages/x-components/src/design-system/components/input-group/default.scss
```diff
@@ -74,15 +76,22 @@
     flex: 1 1 100%;
     min-width: 0;
     border: none;
```
tiborux, line 78: I dont know why, but now we are showing the outline

image

packages/x-components/src/design-system/components/input-group/default.scss
```diff
@@ -74,15 +76,22 @@
...
-      font-family: var(--x-font-family-input-group-placeholder-default);
-      font-size: var(--x-size-font-input-group-placeholder-default);
-      font-weight: var(--x-number-font-weight-input-group-placeholder-default);
-      line-height: var(--x-size-line-height-input-group-placeholder-default);
-    }
+  > .x-input-placeholder {
+    position: absolute;
+    height: 100%;
+    pointer-events: none;
+  }
```
tiborux, line 85: Shouldn't have the same styles as x-input?

```suggestion
  > .x-input-placeholder {
    flex: 1 1 100%;
    min-width: 0;
    border: none;
    position: absolute;
    pointer-events: none;
  }
```

packages/x-components/src/design-system/components/input-group/default.scss
```diff
@@ -74,15 +76,22 @@
     flex: 1 1 100%;
```
tiborux, line 76: In this line:
```
&:first-child {
      padding-inline-start: var(--x-size-padding-left-input-group-default);
    }
    
 ```
 
   We are adding padding to the first child only. This is doing that when you click on the search input, the placeholder will be focused first, and the cursor appears at the left of the input group but some milliseconds before, the cursor will be in the right position.
   
  I think that we should add the same padding the second child or the placeholder

**Review by tajespasarela — CHANGES_REQUESTED** [review]

File: packages/x-components/src/x-modules/search-box/components/search-input-placeholder.vue [hunk 4]
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    beforeDestroy(): void {
+      this.stopAnimationInterval();
+    }
+  }
+</script>
+
+<style lang="scss" scoped>
+  .x-search-input-placeholder {
```
tiborux, line 165: I am not sure about adding these styles here, maybe I will set this values in the design system or inline like:

```
    <SearchInputPlaceholder
                  :messages="searchInputPlaceholderMessages"
                  :animate-only-on-hover="true"
                  class="x-list x-list--justify-center"
                >
 ```

File: packages/x-components/src/design-system/components/input-group/default.scss [hunk 5]
```diff
@@ -2,10 +2,12 @@
```
tiborux, line 2: In `input/default.scss` there are some styles that are being applied to a placeholder:

```suggestion
  &::placeholder {
    // typography
    color: var(--x-color-text-input-placeholder-default);
    font-family: var(--x-font-family-input-placeholder-default);
    font-size: var(--x-size-font-input-placeholder-default);
    font-weight: var(--x-number-font-weight-input-placeholder-default);
    line-height: var(--x-size-line-height-input-placeholder-default);
  }
```

I think that you should do the same for `&-placeholder` like you are doing in this file

packages/x-components/src/x-modules/search-box/components/search-input-placeholder.vue
```diff
@@ -0,0 +1,241 @@
...
+  <component
+    :is="animation"
+    v-if="isVisible"
+    class="x-input-placeholder x-search-input-placeholder"
+    mode="out-in"
+  >
+    <span v-if="isBeingAnimated" :key="animationMessage" data-test="search-input-placeholder">
+      {{ animationMessage }}
+    </span>
+    <span v-else-if="$slots.default" data-test="search-input-placeholder">
```
tiborux, line 11: Do we have to render a span by default? I'm not sure if this should be defined in the slot.

```
  <template v-else-if="$slots.default">
      <!--
        @slot The default placeholder when the `messages` animation is not active.
      -->
      <slot data-test="search-input-placeholder"></slot>
    </template>
```

packages/x-components/src/design-system/components/input-group/default.scss
alvarodE: Yes, it was already "visible" but covered by the Clear button; now that this button is hidden when there is no query the outline can be easily perceived.

packages/x-components/src/x-modules/search-box/components/search-input-placeholder.vue
```diff
@@ -0,0 +1,241 @@
...
+    onUserBlurredSearchBox(): void {
+      this.isSearchBoxFocused = false;
+    }
+
+    mounted(): void {
+      this.resetAnimation();
+
+      this.$watch('isBeingAnimated', (value: boolean) => {
+        if (!value) {
+          this.incrementAnimationMessageIndex();
```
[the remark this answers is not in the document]
alvarodE: Its purpose is to display a different message on each mousenter different than the previous one. This way the user will see always a different message from the array of `messages` until the index reaches its end.

packages/x-components/src/views/home/Home.vue
```diff
@@ -86,13 +86,20 @@
...
+                  Search
+                </SearchInputPlaceholder>
                 <SearchInput
                   aria-label="Search for products"
-                  placeholder="Search"
                   :instant="controls.searchInput.instant"
                   :instant-debounce-in-ms="controls.searchInput.instantDebounceInMs"
                 />
-                <ClearSearchInput aria-label="Clear query">Clear</ClearSearchInput>
+                <ClearSearchInput v-if="$x.query.searchBox" aria-label="Clear query">
```
tajespasarela, line 100: this `v-if` is breaking the E2E tests. Because we clear the query from some tests by clicking this button. If the button is not visible because there is no query, the test fails because it does not find the button.
Even if there is no query in the tests we are clicking it just in case to make sure there is no query in the input.

tajespasarela (changes requested):

packages/x-components/src/x-modules/search-box/components/__tests__/search-input-placeholder.spec.ts
```diff
@@ -0,0 +1,215 @@
...
+  animateOnlyOnHover?: boolean;
+}
+
+interface SearchInputProps {
+  autofocus?: boolean;
+}
+
+interface SearchInputPlaceholderSlots {
+  default?: string;
+}
```
tajespasarela, line 113: Simplify this type and params:
```suggestion
interface TestComponentParams {
  messages: Array<string>;
  animationIntervalMs?: number;
  animateOnlyOnHover?: boolean;
  autofocus?: boolean;
  slotContent?: string;
}
```

packages/x-components/src/x-modules/search-box/components/search-input-placeholder.vue
```diff
@@ -0,0 +1,241 @@
...
+    protected isSearchBoxHovered = false;
+
+    protected isSearchBoxFocused = false;
+
+    protected animationMessageIndex = 0;
+
+    protected animationInterval: number | undefined;
+
+    protected get isVisible(): boolean {
+      return !this.query && !this.isSearchBoxFocused;
```
tajespasarela, line 79: actually, the usual behaviour of an `<input>` placeholder is to show it even it has focus, and only hide when there is something typed in the input:

https://user-images.githubusercontent.com/5759712/195584978-a633f8cb-6f46-4bf4-acfb-48a7654006bc.mov

tajespasarela (changes requested):

packages/x-components/src/x-modules/search-box/components/search-input-placeholder.vue
alvarodE: Talked offline about keeping it as it is because of the issue's requirements, it can be easily changed in the future if the approach mentioned here is seen more convenient.

tiborux (changes requested):

packages/x-components/src/x-modules/search-box/components/__tests__/search-input-placeholder.spec.ts
```diff
@@ -0,0 +1,188 @@
...
+    wrapper,
+    getPlaceholderText,
+    isPlaceholderVisible,
+    hoverInput,
+    emit
+  };
+}
+
+interface MountTestComponentOptions {
+  messages?: Array<string>;
```
tiborux, line 92: Comment these properties

packages/x-components/src/x-modules/search-box/components/search-input-placeholder.vue
```diff
@@ -0,0 +1,294 @@
...
+     *
+     * @public
+     */
+    @Prop({ default: false })
+    protected animateOnlyOnHover!: boolean;
+
+    @State('searchBox', 'query')
+    public query!: string;
+
+    protected isSearchBoxHovered = false;
```
tiborux, line 64: I know that is verbose, but we should comment everything. Props, methods, variables, getters, computed...etc

packages/x-components/src/x-modules/search-box/components/search-input-placeholder.vue
```diff
@@ -0,0 +1,294 @@
...
+    <span :key="message" data-test="search-input-placeholder">
+      {{ message }}
+    </span>
+  </component>
+</template>
+
+<script lang="ts">
+  import Vue from 'vue';
+  import { Component, Prop, Watch } from 'vue-property-decorator';
+
```
tiborux, line 17: remove this white space

packages/x-components/src/x-modules/search-box/components/__tests__/search-input-placeholder.spec.ts
tajespasarela: document*

tiborux: Also, Move the interfaces below the describe, at the end of the file

packages/x-components/src/x-modules/search-box/components/__tests__/search-input-placeholder.spec.ts
```diff
@@ -0,0 +1,188 @@
...
+}
+
+describe('testing search input placeholder component', () => {
+  beforeAll(jest.useFakeTimers);
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+  afterEach(jest.clearAllTimers);
+
+  it('is an XComponent', async () => {
```
tiborux, line 114: You can unify both test:

```
  it('is an XComponent', () => {
    const { wrapper } = renderSearchInputPlaceholder();
    const placeholderWrapper = wrapper.findComponent(SearchInputPlaceholder);
    expect(isXComponent(placeholderWrapper.vm)).toEqual(true);
    expect(getXComponentXModuleName(placeholderWrapper.vm)).toBe('searchBox');
  });
  ```

**Review by tajespasarela — CHANGES_REQUESTED** [review]

File: packages/x-components/src/x-modules/search-box/components/__tests__/search-input-placeholder.spec.ts [hunk 15]
```diff
@@ -0,0 +1,188 @@
...
+    expect(getPlaceholderText()).toEqual(messages[0]);
+    await emit('UserAcceptedAQuery', 'testing');
+    expect(isPlaceholderVisible()).toEqual(false);
+    jest.advanceTimersByTime(2000);
+    await emit('UserAcceptedAQuery', undefined);
+    expect(getPlaceholderText()).toEqual(messages[1]);
+  });
+
+  it('is not visible when the search input is focused', async () => {
+    const { wrapper, isPlaceholderVisible } = await mountTestComponent({
```
tiborux, line 175: You can await the triggers:

```
    const { wrapper, isPlaceholderVisible } = renderSearchInputPlaceholder({
      autofocus: true
    });
    await wrapper.vm.$nextTick();
    expect(isPlaceholderVisible()).toEqual(false);

    const inputWrapper = wrapper.findComponent(SearchInput);

    await inputWrapper.trigger('blur');
    expect(isPlaceholderVisible()).toEqual(true);

    await inputWrapper.trigger('focus');
    expect(isPlaceholderVisible()).toEqual(false);
    ```
    
    https://v1.test-utils.vuejs.org/api/wrapper/trigger.html

File: packages/x-components/src/x-modules/search-box/components/__tests__/search-input-placeholder.spec.ts [hunk 16]
```diff
@@ -0,0 +1,188 @@
...
+
+  function hoverInput(mode: 'in' | 'out'): Promise<void> {
+    const inputWrapper = wrapper.findComponent(SearchInput);
+    const hoverEvent = {
+      in: 'mouseenter',
+      out: 'mouseleave'
+    }[mode];
+    inputWrapper.trigger(hoverEvent);
+    return wrapper.vm.$nextTick();
+  }
```
tiborux, line 74:
```suggestion
  function hoverInput(mode: 'in' | 'out'): Promise<void> | void {
    const inputWrapper = wrapper.findComponent(SearchInput);
    const hoverEvent = {
      in: 'mouseenter',
      out: 'mouseleave'
    }[mode];
    return inputWrapper.trigger(hoverEvent);
  }
```

packages/x-components/src/x-modules/search-box/components/__tests__/search-input-placeholder.spec.ts
```diff
@@ -0,0 +1,188 @@
...
+  });
+
+  it('is not visible when there is a query set', async () => {
+    const { messages, getPlaceholderText, isPlaceholderVisible, emit } = await mountTestComponent();
+
+    expect(getPlaceholderText()).toEqual(messages[0]);
+    await emit('UserAcceptedAQuery', 'testing');
+    expect(isPlaceholderVisible()).toEqual(false);
+    jest.advanceTimersByTime(2000);
+    await emit('UserAcceptedAQuery', undefined);
```
tiborux, line 170: UserAcceptedAQuery expects a string

```suggestion
    await emit('UserAcceptedAQuery', '');
```

tiborux (changes requested):
gj

tiborux (approved):

CachedaCodes (approved):

PR #772, opened 2022-10-12
Add search input hover events

alvarodE:
Related to https://github.com/empathyco/x/issues/385.

In order to implement the new SearchInputPlaceholder component the hover events (mouse enter & leave) must be emitted by the SearchInput so its state is available from any other element/component.

Note: The first approach for the SearchInputPlaceholder was to directly use its hover state but as it is conceived as a span placed in the same space than the input, that would generate conflicts between the mouse events from both elements depending on which one is placed on top of the other.

This PR is part of a specific issue and how it is used can be found [here](https://github.com/empathyco/x/pull/771/files#diff-4f94a892b7632521d53664a6cb2c3881f76f05e0bb615a6ffdf1ab29375c100aR118).

### DIFF
--- a/packages/x-components/src/x-modules/search-box/events.types.ts
+++ b/packages/x-components/src/x-modules/search-box/events.types.ts
@@ -7,67 +7,67 @@
 export interface SearchBoxXEvents {
   /**
    * The search-box query has changed
-   * Payload: The new search-box query.
+   * * Payload: The new search-box query.
    */
   SearchBoxQueryChanged: string;
   /**
    * The user hovered in the search-box.
-   * Payload: none.
+   * * Payload: none.
    */
   UserHoveredInSearchBox: void;
   /**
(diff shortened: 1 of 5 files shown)


packages/x-components/src/x-modules/search-box/events.types.ts
```diff
@@ -7,57 +7,67 @@
 export interface SearchBoxXEvents {
   /**
    * The search-box query has changed
-   * * Payload: The new search-box query.
```
alvarodE, line 10: ⚠️ The `eslint --fix` of the `lint-staged` removes these asteriks.

tiborux (approved):

tajespasarela (dismissed):

tiborux (approved):

tajespasarela (approved):

PR #773, opened 2022-10-12
Add directional origins for translate animations

alvarodE:
Related to https://github.com/empathyco/x/issues/385.

Translate animations must be extended to support a different direction for enter and leave so the following case is achievable (search input placeholder animation):

https://user-images.githubusercontent.com/72568818/195307148-8a0b52cb-6736-48f6-bf51-caf427c56cc7.mov

This PR is part of a specific issue and how it is used can be found [here](https://github.com/empathyco/x/pull/771/files#diff-4f94a892b7632521d53664a6cb2c3881f76f05e0bb615a6ffdf1ab29375c100aR48).

### DIFF
--- a/packages/x-components/src/components/animations/create-directional-animation-factory.ts
+++ b/packages/x-components/src/components/animations/create-directional-animation-factory.ts
@@ -33,4 +33,12 @@ export function createDirectionalAnimationFactory(
   };
 }
 
-export type AnimationOrigin = 'top' | 'bottom' | 'left' | 'right';
+export type AnimationOrigin =
+  | 'top'
+  | 'bottom'
+  | 'left'
+  | 'right'
+  | 'top-to-bottom'
+  | 'bottom-to-top'
+  | 'left-to-right'
+  | 'right-to-left';
--- a/packages/x-components/src/components/animations/animate-translate/animate-translate.style.scss
+++ b/packages/x-components/src/components/animations/animate-translate/animate-translate.style.scss
@@ -21,4 +21,34 @@
       transform: translateX(100%);
     }
   }
+
+  &--enter {
+    &#{$p}--top-to-bottom {
+      transform: translateY(-100%);
+    }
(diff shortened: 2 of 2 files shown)


packages/x-components/src/components/animations/create-directional-animation-factory.ts
```diff
@@ -33,4 +33,12 @@ export function createDirectionalAnimationFactory(
...
-export type AnimationOrigin = 'top' | 'bottom' | 'left' | 'right';
+export type AnimationOrigin =
+  | 'top'
+  | 'bottom'
+  | 'left'
+  | 'right'
+  | 'top-to-bottom'
+  | 'bottom-to-top'
+  | 'left-to-right'
+  | 'right-to-left';
```
tiborux, line 44: Here, We are talking about the origin of the animation, should we rename the type to AnimationDirection? because `top-to-bottom` is not an origin, I see it as a direction

if we do this change, when can rename the values to be more explicit about what they are going to do:

```suggestion
export type AnimationOrigin =
  | 'top-to-top'
  | 'bottom-to-bottom'
  | 'left-to-left'
  | 'right-to-right'
  | 'top-to-bottom'
  | 'bottom-to-top'
  | 'left-to-right'
  | 'right-to-left';
```

I know that its redundant but I think is clear to understand what they are doing.

tajespasarela (approved):

packages/x-components/src/components/animations/create-directional-animation-factory.ts
tajespasarela: I was thinking also to change the name, but `AnimationDirection` is confusing, because when is `top`, it is moving to the bottom direction on entering, then it can be confusing, and renaming all the current naming to change `top` by `bottom` and `left` to `right` it is a confusing breaking change.

So, to me is better to leave it as it is.  But up to you

tiborux (approved):
gj

PR #778, opened 2022-10-13
feat: add tailwind showcase component.

javieri-empathy:
Adds showcase components for the tailwind plugin.

- Updates vite to version 3
- Generates a new entry point for tailwind plugin: `@empathyco/x-tailwindcss/showcase` with the showcase components
- Configures build of this new entry point with rollup, as with Vite I found impossible to generate proper types in library mode.


packages/x-tailwindcss/demo/vite.config.ts
```diff
@@ -0,0 +1,47 @@
...
+                  return {};
+                },
+                dynamicUtilities() {
+                  return {};
+                },
+                dynamicComponents() {
+                  return {};
+                },
+                theme: {}
+              })
```
tajespasarela, line 36: Can we move this config back to a separate file?

tajespasarela (changes requested):

packages/x-tailwindcss/package.json
```diff
@@ -35,15 +35,39 @@
...
     "tslib": "~2.3.0"
   },
+  "peerDependencies": {
+    "vue-class-component": "^7.1.0",
+    "vue-property-decorator": "^8.3.0",
+    "vue": "^2.6.0"
+  },
+  "peerDependenciesMeta": {
+    "vue-class-component": {
+      "optional": true
```
javieri-empathy, line 45: This makes the peeer dependencies optional. If you don't use the showcase components you don't need them

packages/x-components/tailwind.config.js
```diff
@@ -3,13 +3,27 @@ const options = require('./src/tailwind/plugin-options');
 
 /** @type {import('tailwindcss').Config} */
 module.exports = {
-  content: ['./public/index.html', './src/**/*.vue'],
+  content: [
+    './public/index.html',
+    './src/**/*.vue',
+    './node_modules/@empathyco/x-tailwindcss/showcase/**/*.js'
```
javieri-empathy, line 9: We have to be very careful with this if moved to the archetype. With this line it will include the full set of components CSS. It should only be included for development purposes.

packages/x-tailwindcss/demo/src/components/xds-button.vue
```diff
@@ -0,0 +1,81 @@
...
+      </button>
+    </div>
+  </div>
+</template>
+
+<script lang="ts">
+  import { Vue, Component, Prop } from 'vue-property-decorator';
+
+  @Component
+  export default class XdsButtonShowcase extends Vue {
```
alvarodE, line 21: I'm wondering whether it makes sense or not to have a mixin covering all the logic that may be shared between different showcase components in the future. In the end I understand that sizes, colors, etc. apply to most of them, so my suggestion is something like this:
```

import { Vue, Component, Prop } from 'vue-property-decorator';

@Component
export default class XdsComponentShowcaseMixin extends Vue {
  @Prop({ required: true })
  public baseClass!: string;

  @Prop({ default: () => ['xs', 'sm', 'md', 'lg', 'xl'] })
  public sizes!: string[];

  @Prop({
    default: () => [
      'neutral',
      'lead',
      'auxiliary',
      'accent',
      'highlight',
      'success',
      'warning',
      'error'
    ]
  })
  public colors!: string[];

  @Prop()
  public combinations!: string[];

  protected get sections(): Record<string, string[]> {
    return {
      Default: [this.baseClass],
      Colors: this.colors.map(this.prefixWith(this.baseClass)),
      Sizes: this.sizes.map(this.prefixWith(this.baseClass)),
      ...(this.combinations && {
        Combinations: this.combinations.map(this.prefixWith(this.baseClass))
      })
    };
  }

  protected prefixWith(prefix: string): (value: string) => string {
    return cssClass => `${prefix} ${prefix}-${cssClass}`;
  }
}

```

packages/x-tailwindcss/tsconfig.json
```diff
@@ -11,8 +12,8 @@
     "sourceMap": true,
     "inlineSources": true,
     "declaration": true,
-    "declarationDir": "./dist/types",
-    "noEmit": true
+    "skipLibCheck": true,
+    "skipDefaultLibCheck": true
```
diegopf, line 16: do we need both skipLibChecks? According to the [docs](https://www.typescriptlang.org/tsconfig#skipDefaultLibCheck) I understand that not.

packages/x-tailwindcss/demo/src/components/xds-button.vue
```diff
@@ -0,0 +1,81 @@
...
+    protected prefixWith(...prefix: string[]): (value: string) => string {
+      return cssClass => `${prefix.join(' ')} ${cssClass}`;
+    }
+
+    protected removeBase(cssClass: string): string {
+      return cssClass.replace(new RegExp(`${this.base}-?`, 'g'), '') || 'button';
+    }
+
+    protected copyCSSClasses(event: MouseEvent): void {
+      navigator.clipboard.writeText((event.target as HTMLElement).classList.value);
```
diegopf, line 78: This is so nice! ❤️ but might be a good idea to give a hint about it.  Something like adding a `title="click me for copy the classes"` to the button so it appears on the hover.

packages/x-tailwindcss/demo/src/components/xds-button.vue
diegopf: I think @alvarodE comment makes sense. Another option is to create a generic component that receives via props the configuration. The `showcase` component could get the tailwind config via [resolveConfig](https://tailwindcss.com/docs/configuration#referencing-in-java-script) so we have the prefix and colors. Afterwards a generic `xds-component` could just paint the given config. Something like 

```
// Showcase component
    mounted(): void {
      const config = resolveConfig(this.tailwindConfigFile);
      this.colors = config.theme?.colors ?? {};
      this.prefix = config.prefix ??  '';
    }
```

```<template>
  <div>
    <h3>{{ componentConfig.tag }}</h3>
    <ul>
      <li v-for="(_, colorName) in tailwindConfig.colors" :key="colorName">
        <component
          :is="componentConfig.tag"
          :class="[
            `${tailwindConfig.prefix}${componentConfig.base}-${colorName}`,
            componentConfig.classes
          ]"
        >
          {{ colorName }}
        </component>
      </li>
    </ul>
  </div>
</template>

<script lang="ts">
  import { Component, Prop, Vue } from 'vue-property-decorator';
  @Component
  export default class XdsItem extends Vue {
    @Prop({
      required: true
    })
    public tailwindConfig!: {
      prefix: string;
      colors: Record<string, object>;
    };
    @Prop({
      required: true
    })
    public componentConfig!: {
      tag: string;
      base: string;
      classes: string;
    };
  }
</script>
```


In the example above only colors are displayed but I think the point is clear 😅

packages/x-components/tailwind.config.js
```diff
@@ -3,13 +3,27 @@ const options = require('./src/tailwind/plugin-options');
...
   prefix: 'x-',
   plugins: [xTailwindPlugin(options), oldDsPlugin],
   // In order to avoid conflicting class names with the old design system while the new XDS is being developed
-  // we are disabling all corePlugins from tailwindcss but `fontSize`, `fontWeight`, `lineHeight` and `textColor`.
+  // we are disabling all corePlugins from tailwindcss but some of them.
   // You can check that tailwindcss is still working by checking the style of the heading `Test controls`
   // below the `Start` button on the `x-components` internal demo app.
   // See https://github.com/empathyco/x/pull/655#discussion_r948923711
-  corePlugins: ['fontSize', 'fontWeight', 'lineHeight', 'textColor']
+  corePlugins: [
```
diegopf, line 18: why are we enabling more corePlugins? is it just because we missed them in the past?

packages/x-tailwindcss/tsconfig.json
```diff
@@ -1,8 +1,9 @@
 {
+  "rootDir": "src",
```
diegopf, line 2: why do we need the `rootDir`? What's the difference here? at least I don't see any testing it locally.

packages/x-tailwindcss/rollup.config.js
```diff
@@ -0,0 +1,57 @@
...
+        tsconfig: resolve(__dirname, 'demo/tsconfig.json'),
+        tsconfigOverride: {
+          include: ['src/components/**/*.{ts,vue}', 'src/shims-vue.d.ts'],
+          exclude: ['vite.config.ts', 'tailwindcss-nesting.d.ts']
+        }
+      }),
+      vue()
+    ]
+  }
+];
```
diegopf, line 57: Maybe is it better to have two different config files, one for the internal demo and another of the library and import them here? I think this way will be easier to read and come back later to locate where do we need to modify something.

packages/x-tailwindcss/package.json
diegopf: Nice! I wasn't aware of this

diegopf (commented):
Good job! As we discussed with @tajespasarela earlier I think it could be fine to merge this approach and just rethink it afterwards when we need to include another component in the showcase.

packages/x-tailwindcss/demo/src/components/xds-button.vue
javieri-empathy: I thought about 2 approaches. But it is something I decided not to do in this PR because of early abstractions. I don't have a clear visibility of what would be helpful for each component that we design. So until we have 2-3 more components I wouldn't abstract the utilities to generate this showcase.

packages/x-tailwindcss/tsconfig.json
javieri-empathy: I always forget which one is the one I should use 😅. Good catch

packages/x-components/tailwind.config.js
javieri-empathy: For the showcase view inside x-components. If I don't enable you will see them like this:
image
instead of 
image

packages/x-tailwindcss/tsconfig.json
javieri-empathy: As we are only including in this tsconfig files from `src` it is unnecesary (moreover it should be under `compilerOptions`, there it is doing nothing). I tried to use a single tsconfig instead of 2 but I found it to be more easy to mess up.

packages/x-components/tailwind.config.js
javieri-empathy: This is another proposal that I thought yesterday (yeah, I know, too late) about exposing a dev server rather than a set of Vue components.
In my mind this would allow us to remove the rollup config for the showcase components, it would facilitate using this from a custom project regardless the components library used, and also ease up this configuration (run a command instead of setting up a view and a development tailwind config).
https://searchbroker.atlassian.net/browse/EX-7329

alvarodE: I love how that sounds Javi :sisi3:

alvarodE:
With the last merge of main into this branch there are a lot of changes on the doc files that I think we should discard. Is it because of the Prettier or something? @javieri-empathy

javieri-empathy:
Fuck, I can promise you I didn't format them intentionally. I'm looking at them and I can tell you they some of them were not properly formatted, but there are also some changes that do not make sense, like moving titles to the previous line. My bet is that when merging `main` branch changes prettier decided to format them.

tajespasarela (approved):

PR #784, opened 2022-10-17
Add facets type field on Platform adapter

alvarodE:
[EX-6373](https://searchbroker.atlassian.net/browse/EX-6373)

### DIFF
--- a/packages/x-adapter-platform/src/mappers/responses/__tests__/search-response.mapper.spec.ts
+++ b/packages/x-adapter-platform/src/mappers/responses/__tests__/search-response.mapper.spec.ts
@@ -1,11 +1,8 @@
-import { platformSearchResponse } from '../../../__tests__/__fixtures__/search.response';
-import { PlatformSearchResponse } from '../../../types/responses/search-response.model';
 import { searchResponseMapper } from '../search-response.mapper';
+import { platformSearchResponse } from '../../../__tests__/__fixtures__/search.response';
 
 describe('searchResponseMapper tests', () => {
   it('should map the response', () => {
-    expect(
-      searchResponseMapper(platformSearchResponse as PlatformSearchResponse, {})
-    ).toMatchSnapshot();
+    expect(searchResponseMapper(platformSearchResponse, {})).toMatchSnapshot();
   });
 });
--- a/packages/x-adapter-platform/src/mappers/responses/__tests__/search-response.mapper.spec.ts
+++ b/packages/x-adapter-platform/src/mappers/responses/__tests__/search-response.mapper.spec.ts
@@ -1,8 +1,11 @@
-import { searchResponseMapper } from '../search-response.mapper';
 import { platformSearchResponse } from '../../../__tests__/__fixtures__/search.response';
+import { PlatformSearchResponse } from '../../../types';
+import { searchResponseMapper } from '../search-response.mapper';
 
 describe('searchResponseMapper tests', () => {
   it('should map the response', () => {
-    expect(searchResponseMapper(platformSearchResponse, {})).toMatchSnapshot();
+    expect(
+      searchResponseMapper(platformSearchResponse as PlatformSearchResponse, {})
+    ).toMatchSnapshot();
   });
 });
--- a/packages/x-adapter-platform/src/mappers/responses/__tests__/search-response.mapper.spec.ts
+++ b/packages/x-adapter-platform/src/mappers/responses/__tests__/search-response.mapper.spec.ts
@@ -1,5 +1,5 @@
 import { platformSearchResponse } from '../../../__tests__/__fixtures__/search.response';
-import { PlatformSearchResponse } from '../../../types';
+import { PlatformSearchResponse } from '../../../types/responses/search-response.model';
 import { searchResponseMapper } from '../search-response.mapper';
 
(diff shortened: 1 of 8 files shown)


packages/x-adapter-platform/src/mappers/responses/__tests__/search-response.mapper.spec.ts
```diff
@@ -1,8 +1,11 @@
-import { searchResponseMapper } from '../search-response.mapper';
 import { platformSearchResponse } from '../../../__tests__/__fixtures__/search.response';
+import { PlatformSearchResponse } from '../../../types';
+import { searchResponseMapper } from '../search-response.mapper';
 
 describe('searchResponseMapper tests', () => {
   it('should map the response', () => {
-    expect(searchResponseMapper(platformSearchResponse, {})).toMatchSnapshot();
+    expect(
+      searchResponseMapper(platformSearchResponse as PlatformSearchResponse, {})
```
alvarodE, line 8: Cast needed to avoid problems with the new `PlatformFacetType`.

packages/x-adapter-platform/src/schemas/facets/utils.ts
```diff
@@ -1,12 +1,26 @@
...
+    },
+    hierarchical: {
+      modelName: 'HierarchicalFacet',
+      schema: hierarchicalFilterSchema
+    },
+    range: {
+      modelName: 'NumberRangeFacet',
+      schema: numberFilterSchema
+    }
+  }[type] as FacetConfig;
```
alvarodE, line 25: I'm wondering whether we need a defaulf fallback here or not to cover the case where we'd get from the service an unsupported type. I don't think so because in the end it is a standard API and we should plan and coorinate back & front for a new development in case a new type is needed, but happy to read other opinions :)

packages/x-adapter-platform/src/schemas/models/facet.schema.ts
```diff
@@ -20,29 +16,12 @@ export const facetSchema = createMutableSchema<
...
     $path: 'values',
-    $subSchema: ({ facet }) => getFacetConfig(facet).schema as any,
+    $subSchema: ({ type }) => getFacetConfig(type).schema as any,
     $context: {
       facetId: 'facet'
     }
   }
 });
-
-export const facetsConfig: FacetsConfig = {
```
alvarodE, line 35: With this refactor I see unnecessary to keep this object, but I'm not sure if it could be used for some custom cases. Let me know if we should keep it or not 🙏

packages/x-adapter-platform/src/schemas/models/facet.schema.ts
```diff
@@ -20,29 +16,12 @@ export const facetSchema = createMutableSchema<
 >({
   id: 'facet',
   label: 'facet',
-  modelName: source => {
-    return getFacetConfig(source.facet).modelName as any;
-  },
+  modelName: ({ type }) => getFacetConfig(type).modelName as any,
```
alvarodE, line 19: Check if we can remove these `any` casts with the new change.

packages/x-adapter-platform/src/mappers/responses/__tests__/search-response.mapper.spec.ts
```diff
@@ -1,8 +1,11 @@
-import { searchResponseMapper } from '../search-response.mapper';
 import { platformSearchResponse } from '../../../__tests__/__fixtures__/search.response';
+import { PlatformSearchResponse } from '../../../types';
```
tiborux, line 2: avoid importing from the barrel

```suggestion
import { PlatformSearchResponse } from '../../../types/responses/search-response.model';
```

packages/x-adapter-platform/src/schemas/facets/utils.ts
```diff
@@ -1,12 +1,26 @@
-import { facetsConfig } from '../models/facet.schema';
+import { PlatformFacetType } from '../../types';
+import { hierarchicalFilterSchema, numberFilterSchema, simpleFilterSchema } from '../models';
```
tiborux, line 2: avoid importing from the barrel

packages/x-adapter-platform/src/schemas/facets/utils.ts
tiborux: I am not sure about this, I think that maybe it's a good idea but as you said if it's a standard API this should always be right.  I am not 100% sure that this is going to be right always, you know 😂

tiborux (commented):
gj

packages/x-adapter-platform/src/schemas/facets/utils.ts
alvarodE: Added just in case.

packages/x-adapter-platform/src/mappers/responses/__tests__/search-response.mapper.spec.ts
mnavarroespinosa: What do you think of doing this assertion directly on the `__tests__/__fixtures__/search.response` file?

mnavarroespinosa (commented):
Good job!

mnavarroespinosa (approved):

tiborux (approved):

PR #788, opened 2022-10-19
feat(types): add autocomplete to `theme` function in `TailwindHelpers`

LuisMartinez15:
EX-7085

Adds autocomplete functionality to the `theme` function in `TailwindHelpers` so it is easier to use it. It doesn't restrict values, as those can be extended for customers and would result in errors when doing so.

Regarding modifying the `ExtractPath` type to allow `numeric` keys, it wasn't needed, as those keys can also be used as `strings`.

In order to use this autocomplete feature in `x-components`, you can't destructure the TailwindHelpers parameter.

For example, this allows autocompletion:
image

But this doesn't:
image


CachedaCodes (approved):

tiborux (approved):
gj

PR #798, opened 2022-10-25
feat: mixin to generate props dynamically

diegopf:
EX-7220


packages/x-components/src/components/__tests__/dynamic-props.mixin.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+};
+
+describe('dynamicPropsMixin', () => {
+  it('expects to have the defined props from the mixin', () => {
+    const { wrapper } = renderComponent();
+    const props = Object.keys(wrapper.props());
+    expect(props).toHaveLength(3);
+    expect(props.includes('list')).toBeTruthy();
+    expect(props.includes('button')).toBeTruthy();
+    expect(props.includes('data')).toBeTruthy();
```
alvarodE, line 24: Does it make sense to "simplify" this assertions like this?
```suggestion
    expect(props).toEqual(['list', 'button', 'data']);
```

packages/x-components/src/components/dynamic-props.mixin.ts
```diff
@@ -0,0 +1,37 @@
...
+export function dynamicPropsMixin<T extends string>(
+  propNames: readonly string[]
+): ExtendedVue<Vue, unknown, unknown, unknown, Record<T, string>> {
+  return Vue.extend({
+    props: propNames.reduce(
+      (a, v) => ({ ...a, [v]: { type: String } }),
+      {} as Record<T, PropValidator<string>>
+    )
+  });
+}
```
javieri-empathy, line 37: A couple of suggestions after talking with you:

- Just by linking the generic parameter with the function parameter you will gain type inference, and you would avoid forcing TS users to pass both the generic and the standard parameter.
- Our naming convention discourages using abbreviations or single letters for everything. There are severa places where this is wrong, so feel free to change it if you find one of those cases.
- I guess the final type should be optional. I suggested below  using an inline type, but feel free to change it to a `Partial<Record<PropNames, string>>` if you find it easier to read.

```suggestion
export function dynamicPropsMixin<PropNames extends string>(
  propNames: PropNames[]
): ExtendedVue<Vue, unknown, unknown, unknown, { [Key in PropNames]?: string }> {
  return Vue.extend({
    props: propNames.reduce(
      (props, propName) => ({ ...props, [propName]: { type: String } }),
      {} as Record<PropNames, PropValidator<string>>
    )
  });
}
```

packages/x-components/src/components/__tests__/dynamic-props.mixin.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+  });
+});
+
+/**
+ * The options for the `renderComponent` function.
+ */
+interface ComponentOptions {
+  props?: readonly string[];
+}
```
javieri-empathy, line 33: After talking with you I realised that complicating code here to have types makes no sense. In the end you have no types as the testing wrapper has the `Wrapper<Vue>` type. We could complicate it and return an `ExtendedVue` like you did for the mixin factory, but (for one time) I think it is complicating things unnecessarily. So you could simply remove the readonly stuff if you want and forget about literal types here.

packages/x-components/src/components/__tests__/dynamic-props.mixin.spec.ts
```diff
@@ -0,0 +1,41 @@
+import Vue from 'vue';
+import { mount, Wrapper } from '@vue/test-utils';
+import { ExtractArrayItems } from '@empathyco/x-utils';
+import { dynamicPropsMixin } from '../dynamic-props.mixin';
+
+const renderComponent = ({
+  props = ['list', 'button'] as const
+}: ComponentOptions = {}): ComponentAPI => {
+  const wrapper = mount({
+    mixins: [dynamicPropsMixin<ExtractArrayItems<typeof props>>(props)],
```
javieri-empathy, line 10:
```suggestion
    mixins: [dynamicPropsMixin(props)],
```

javieri-empathy (changes requested):

packages/x-components/src/components/__tests__/dynamic-props.mixin.spec.ts
javieri-empathy: There is a reason to do it the other way. With `.toEqual` you are also implicitly testing the order, which some times like this one we do not control. I don't think this will change in the short term, so we can leave it like you suggested.

javieri-empathy (approved):

alvarodE (approved):

PR #801, opened 2022-10-27
feat: Use Promotion position to sort grid

alvarodE:
[EX-7295](https://searchbroker.atlassian.net/browse/EX-7295)

### DIFF
--- a/packages/x-components/src/x-modules/search/components/__tests__/promoteds-list.spec.ts
+++ b/packages/x-components/src/x-modules/search/components/__tests__/promoteds-list.spec.ts
@@ -118,13 +118,13 @@ describe('testing PromotedsList component', () => {
   });
 
   it('provides the result of concatenating ancestor injected items with the promoteds', () => {
-    const resultStub = getResultsStub().slice(0, 1);
-    const promotedStub = getPromotedsStub().slice(0, 1);
+    const resultStub = getResultsStub();
+    const promotedStub = getPromotedsStub();
     const localVue = createLocalVue();
     localVue.use(Vuex);
     const store = new Store<DeepPartial<RootXStoreState>>({});
     installNewXPlugin({ store }, localVue);
-    resetXSearchStateWith(store, { promoteds: promotedStub });
+    resetXSearchStateWith(store, { promoteds: promotedStub, totalResults: resultStub.length * 2 });
 
     /* It provides an array with one result */
     @Component({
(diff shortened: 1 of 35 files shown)


ajperezbau (commented):
Good job 👌

packages/x-components/src/x-modules/search/components/promoteds-list.vue
```diff
@@ -74,9 +74,23 @@
...
+      const items = [...this.injectedListItems];
+      for (const item of this.stateItems) {
+        let index = item.position - 1;
+        while (items.at(index)?.modelName === 'Promoted') {
+          index++;
+        }
+        const isIndexInLoadedPages = index <= items.length;
+        const areAllPagesLoaded = this.$x.results.length === this.$x.totalResults;
+        if (!isIndexInLoadedPages && !areAllPagesLoaded) {
+          break;
```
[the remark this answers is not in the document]
alvarodE: Yes, the Search service sorts promoteds & banners taking into account their position and, in case of collission, the creation date.

ajperezbau (approved):

packages/x-components/src/x-modules/search/components/__tests__/promoteds-list.spec.ts
```diff
@@ -118,13 +118,13 @@ describe('testing PromotedsList component', () => {
...
-    const resultStub = getResultsStub().slice(0, 1);
-    const promotedStub = getPromotedsStub().slice(0, 1);
+    const resultStub = getResultsStub();
+    const promotedStub = getPromotedsStub();
     const localVue = createLocalVue();
     localVue.use(Vuex);
     const store = new Store<DeepPartial<RootXStoreState>>({});
     installNewXPlugin({ store }, localVue);
-    resetXSearchStateWith(store, { promoteds: promotedStub });
+    resetXSearchStateWith(store, { promoteds: promotedStub, totalResults: resultStub.length * 2 });
```
diegopf, line 127: I see we have 4 results and 6 promoted, is it okay for the test to only load 9 items? didn't we talk about to concatenate all the promoted that are left behind if there are no more pages? or is it just true for the first one?

diegopf (changes requested):

packages/x-components/src/x-modules/search/components/__tests__/promoteds-list.spec.ts
alvarodE: For the case of the test the situation is that there are 4 results loaded but the total is 8 (4 * 2), so it is simulating the case that not all the results are already fetched. 

In that case, promoteds with a position out of the current page won't be displayed, and that's the case of the last promoted in the test (the one with position = 11).

diegopf (approved):

ajperezbau (approved):

PR #814, opened 2022-10-30
ci: add cache in github actions

tiborux:
# Pull request template

Some steps in the CI can use a cache, like install and lint steps.

## Motivation and context

This PR is done because right now our CI is really slow, with the cache enabled for the install and the eslint steps, we are able to reduce 4-5min the time of the actions. 

[Github actions cache](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows)

I read this post: [Speed your CI](https://jongleberry.medium.com/speed-up-your-ci-and-dx-with-node-modules-cache-ac8df82b7bb0). It said that maybe its better to cache `node_module/.cache`instead of the whole `node_modules` but I am not sure. 

## Type of change

- [X] `Main`
- [ ] Other. Specify:

## How has this been tested?

After updating the changes, you can check the different times between the 2 actions. The first time that you upload the code, it took:
image

 if you relaunch the same action, the time will be decreased by 4-5 min (it changes between relaunch)
 
 
image

## Checklist:

- [X] I have made corresponding changes to the **[documentation](./CONTRIBUTING.md#documentation-style-guide)**.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,9 +15,33 @@ jobs:
         with:
           fetch-depth: 0
           persist-credentials: false
+      - name: Cache node modules
+        id: cache-node-modules
+        uses: actions/cache@v3
+        with:
+          path: |
+            **/node_modules
+            ~/.cache/Cypress
+          key: node-modules-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            node-modules-
+      - name: Cache eslint
+        id: cache-eslint
+        uses: actions/cache@v3
+        with:
+          path: |
+            .eslintcache
+          key: eslint-${{github.sha}}
+          restore-keys: |
+            eslint-
       - name: Install lerna and all packages
+        if: steps.cache-node-modules.outputs.cache-hit != 'true'
         run: npm ci
+      - name: Build
+        if: steps.cache-node-modules.outputs.cache-hit == 'true'
+        run: npm run build
       - name: Run linter in all packages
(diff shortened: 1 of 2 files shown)


ajperezbau (changes requested):
Good job, this is nice 👌 

To future reviewers, keep in mind that for what I've seen, it is often recommended to not cache the `node_modules` directory, and cache instead the `npm` cache directory.

This doesn't seem to be the case when using `lerna`, as stated [here](https://github.com/actions/cache/blob/main/examples.md#node---npm).

Also, we don't run `npm ci` when a `node_modules` cache is found, and the Node version to use is strict, so I think that caching the `node_modules` is a good solution 👌

tiborux:
Check jest cache

.github/workflows/build.yml
```diff
@@ -15,8 +15,30 @@ jobs:
         with:
           fetch-depth: 0
           persist-credentials: false
+      - name: Cache node modules
```
LuisMartinez15, line 18: I'd add a comment here pointing out what @ajperezbau and the github actions docs say regarding caching `node_modules`. I know right now we are not using the `setup-node` action, as the docker image provided by `cypress` that we are using already has node preinstalled in it. Nonetheless, since we might totally change how we manage and/or hoist our dependencies (maybe start using `pnpm` instead of `npm` + `lerna bootstrap --hoist`) and might also end up replacing `cypress` in favour of `playwright`, I'd state the good practice just for our future selves.

LuisMartinez15 (approved):
good job!

PR #815, opened 2022-11-01
feat: apply dynamic css mixin to sliding panel and column picker list buttons

ajperezbau:
The `base-dropdown` component is not being used in the `archetype`, so we can add the prop there when we need it.

EX-7224

### DIFF
--- a/packages/x-components/src/components/sliding-panel.vue
+++ b/packages/x-components/src/components/sliding-panel.vue
@@ -40,7 +40,7 @@
   import { Debounce } from './decorators/debounce.decorators';
   import { dynamicPropsMixin } from './dynamic-props.mixin';
 
-  /**.
+  /**
    * This component allows for any other component or element inside it to be horizontally
    * navigable. It also implements customizable buttons as well as other minor customizations to its
    * general behavior.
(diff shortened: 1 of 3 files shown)


packages/x-components/src/components/sliding-panel.vue
```diff
@@ -35,20 +35,23 @@
...
 <script lang="ts">
-  import Vue from 'vue';
-  import { Component, Prop } from 'vue-property-decorator';
+  import { Component, Mixins, Prop } from 'vue-property-decorator';
   import { VueCSSClasses } from '../utils/types';
   import { Debounce } from './decorators/debounce.decorators';
+  import { dynamicPropsMixin } from './dynamic-props.mixin';
 
-  /**
+  /**.
```
mnavarroespinosa, line 43:
```suggestion
  /**
```

mnavarroespinosa (commented):
good job!

packages/x-components/src/components/sliding-panel.vue
alvarodE, line 43:
```suggestion
  /**
```

alvarodE (dismissed):

alvarodE:
Good job! I'd say that it could be worth to add this functionality to the base-dropdown as we have it already identified but also understand the point of leaving it for the future 👍

alvarodE (approved):

mnavarroespinosa (approved):

PR #816, opened 2022-11-01
feat: wai-aria for BaseEventButton based components

mavmaf:
The first component to analyse was the `BaseEventButton`. However, due to its versatile nature, it is not possible to predict which behaviour is going to be given to the component, so the focus was then set to the component using this `BaseEventButton`:
- `BaseIdTogglePanelButton`
- `ScrollToTop`
- `AllFilter`
- `ClearHistoryQueries`
- `RemoveHistoryQuery`
- `BaseResultAddToCart`
- `ClearFilters`

You'll see that depending on the component, `aria-label` or `aria-labelledby` was added:
- ~~`aria-labelledby ` for still quite configurable components, such as `BaseIdTogglePanelButton` (in case of visually impaired users, they will need to know what's the panel for, so we force that indication with aria-labelledby), `BaseResultAddToCart` (clients can use different namings: card, bag, basket... with this label we are giving the option to indicate how the screen reader will refer to it), or `AllFilter` (we'll need to indicate the facet we are referring to).~~
**UPDATE:**
- `aria-labelledby ` labels have been removed from the component. Two reason behind this correction:
      1. We're assuming no one will include this components with no text or no icon at all indicating to a non visually impaired 
      which is the function of that button.
      2. In case an icon/image is added, the alt attribute of that image will be used as the accessible name. Moreover, "image 
      buttons use the alt attribute as the label. The alt attribute value must be provided, and it must be clear and concise 
      and representative of the action performed when the button is activated by the user (not a description of the image 
      itself)" ([source](https://dequeuniversity.com/rules/axe/4.5/input-image-alt))

- `aria-label` for components with predictable behaviour: `ScrollToTop`, `ClearHistoryQueries`, `RemoveHistoryQuery`. No matter how the button has in it, they will be used for a specific purpose.
- No label was added to `ClearFilters` component, as its slot describes the behaviour per se.

To test the changes you have available `accessibility-check` view. You can play around with Screen Reader (Chrome extension) and LightHouse: test that buttons' functionality are understandable and Lighthouse accessibility report shows no errors even if there is no text or icon on those components (you won't see 100% accomplishment in Lighthouse report due to color contrast issues, but that's out of the scope of this PR.

Components should also keep to these guidelines: https://searchbroker.atlassian.net/wiki/spaces/~5fae2ee07cfc240071ff281b/pages/3864527026/Button+ordinary+toggle+and+menu+buttons

## Type of change

- [x] `Main`
- [ ] Other. Specify:

### DIFF
--- a/packages/x-components/src/components/panels/base-id-toggle-panel-button.vue
+++ b/packages/x-components/src/components/panels/base-id-toggle-panel-button.vue
@@ -5,7 +5,7 @@
     class="x-button x-base-id-toggle-panel-button"
     data-test="base-id-toggle-button"
     :aria-pressed="isPanelOpen"
-    aria-labelledby="toggle-panel"
+    aria-labelledby="base-id-toggle-button"
   >
     <!-- @slot (Required) Button content with a text, an icon or both -->
     <slot :isPanelOpen="isPanelOpen" />
--- a/packages/x-components/src/components/panels/base-id-toggle-panel-button.vue
+++ b/packages/x-components/src/components/panels/base-id-toggle-panel-button.vue
@@ -4,6 +4,8 @@
     :events="events"
     class="x-button x-base-id-toggle-panel-button"
     data-test="base-id-toggle-button"
+    :aria-pressed="isPanelOpen"
+    aria-labelledby="toggle-panel"
   >
     <!-- @slot (Required) Button content with a text, an icon or both -->
     <slot :isPanelOpen="isPanelOpen" />
--- a/packages/x-components/src/components/panels/base-id-toggle-panel-button.vue
+++ b/packages/x-components/src/components/panels/base-id-toggle-panel-button.vue
@@ -5,7 +5,7 @@
     class="x-button x-base-id-toggle-panel-button"
     data-test="base-id-toggle-button"
     :aria-pressed="isPanelOpen"
-    aria-labelledby="base-id-toggle-button"
+    :aria-labelledby="`base-id-toggle-button-${panelId}`"
   >
     <!-- @slot (Required) Button content with a text, an icon or both -->
     <slot :isPanelOpen="isPanelOpen" />
--- a/packages/x-components/src/components/panels/base-id-toggle-panel-button.vue
+++ b/packages/x-components/src/components/panels/base-id-toggle-panel-button.vue
@@ -5,7 +5,6 @@
     class="x-button x-base-id-toggle-panel-button"
     data-test="base-id-toggle-button"
     :aria-pressed="isPanelOpen"
-    :aria-labelledby="`base-id-toggle-button-${panelId}`"
   >
     <!-- @slot (Required) Button content with a text, an icon or both -->
     <slot :isPanelOpen="isPanelOpen" />
--- a/packages/x-components/src/views/accessibility-check.vue
+++ b/packages/x-components/src/views/accessibility-check.vue
@@ -35,7 +35,9 @@
         <FacetsProvider :facets="[facet]" />
         <Facets>
           <template #default="{ facet }">
-            <AllFilter :facet="facet" />
+            <AllFilter :id="facet.label" :facet="facet">
+              {{ facet.label }}
+            </AllFilter>
             <Filters v-slot="{ filter }" :filters="facet.filters">
               <SimpleFilter :filter="filter" />
             </Filters>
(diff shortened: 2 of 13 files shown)


packages/x-components/src/x-modules/history-queries/components/history-query.vue
```diff
@@ -21,6 +21,7 @@
       class="x-history-query__remove"
       :historyQuery="suggestion"
       data-test="remove-history-query"
+      aria-label="remove"
```
alvarodE, line 24: I'm wondering whether this `aria-label` should go here or directly inside `remove-history-query.vue` 🤔 How do you make that decission? (Sorry if I'm missing something, I don't have any expierence on this topic so first I would like to understand how do you face this kind of issues).

packages/x-components/src/x-modules/history-queries/components/clear-history-queries.vue
```diff
@@ -5,6 +5,7 @@
     :disabled="isHistoryQueriesEmpty"
     :events="clearHistoryQueriesEvents"
     data-test="clear-history-queries"
+    aria-label="Clear all"
```
alvarodE, line 8: In case a setup gives a meaningful content to the slot below, which element has a preference? The `aria-label` or that content? If it is the `aria-label` it may force us to also override this "Clear all" with another value for the `aria-label` more related to the setup itself 🤔

packages/x-components/src/x-modules/facets/components/filters/all-filter.vue
```diff
@@ -3,9 +3,10 @@
     class="x-filter x-all-filter"
     data-test="all-filter"
     role="checkbox"
-    :aria-checked="facet.label.toString()"
+    :aria-checked="isSelected"
     :events="clickEvent"
     :class="cssClasses"
+    :aria-labelledby="facet.label"
```
alvarodE, line 9: As this value will be mapped into an HTML ID from another part of the application, should if have a prefix/sufix to "join" both elements without any other component using the same ID? Maybe something like:
```suggestion
    :aria-labelledby="`all-filter-${facet.label}`"
```

packages/x-components/src/components/panels/base-id-toggle-panel-button.vue
```diff
@@ -4,6 +4,8 @@
     :events="events"
     class="x-button x-base-id-toggle-panel-button"
     data-test="base-id-toggle-button"
+    :aria-pressed="isPanelOpen"
+    aria-labelledby="base-id-toggle-button"
```
alvarodE, line 8: I put a similar comment on other file, should we use here the panel ID to make sure we don't generate any conflict when using more than one panel at the same time?
```suggestion
    :aria-labelledby="`base-id-toggle-button-${panelId}`"
```

packages/x-components/src/x-modules/history-queries/components/history-query.vue
mavmaf: Due to the HistoryQuery component structure I would say there's no difference between putting it inside `remove-history-query.vue` or `history-query.vue` but yeah, you're right, I'll change it even if it is just for consistency purposes.

packages/x-components/src/x-modules/history-queries/components/clear-history-queries.vue
mavmaf: In case you need to substitute it, you can use `aria-labelledby`, placing it together with the id:
```
      <div>
        <h1>ClearHistoryQueries and RemoveHistoryQuery</h1>
        <SearchInput />
        <ClearHistoryQueries aria-labelledby="here" id="here">Not reading this since there is an aria-label</ClearHistoryQueries>
        <HistoryQueries :maxItemsToRender="3" />
      </div>
```
This way we ensure the button will be accessible by default but we can also change the content if needed.

packages/x-components/src/views/accessibility-check.vue
```diff
@@ -0,0 +1,116 @@
...
+      <div>
+        <h1>BaseIdTogglePanelButton</h1>
+        <BaseIdTogglePanelButton
+          :id="`base-id-toggle-button-${panelId}`"
+          class="x-button x-button--ghost"
+          :panelId="panelId"
+        >
+          Panel
+        </BaseIdTogglePanelButton>
+        <BaseIdTogglePanel :startOpen="true" panelId="aside-panel">
```
alvarodE, line 13:
```suggestion
        <BaseIdTogglePanel :startOpen="true" :panelId="panelId">
```

alvarodE (approved):

ajperezbau (approved):

alvarodE (approved):

PR #819, opened 2022-11-02
Feature/update cypress 10

JaviIG:
Upgrades Cypress to version 12. This enables us to test on Safari, and should improve Cypress performance on Silicon macbooks.

- Ignore `cypres.config.ts` in eslint from now. This messes us types with `jest` if included in the src `tsconfig`. PR is already big enough to do another change here.
- Component testing:
  - Corrects imports
  - Loading CSS from the `mount` function has been depracted, so I added an small utility
- E2e testing
  - Fixes issue with sort test: Cypress couldn't locate the element to select using the `children` call
  - Fixes issue with keyboard navigation. It wasn't able to retrieve the old focused element correctly. Now it does.


packages/x-components/cypress.config.ts
```diff
@@ -0,0 +1,45 @@
+import { addCucumberPreprocessorPlugin } from '@badeball/cypress-cucumber-preprocessor';
+import { createEsbuildPlugin } from '@badeball/cypress-cucumber-preprocessor/esbuild';
+// @ts-expect-error TS definitions are wrong
```
mnavarroespinosa, line 3: I don't see any wrong TS definition, where was it failing for you?

mnavarroespinosa (commented):
good job!

packages/react-wrapper/package.json
```diff
@@ -48,7 +48,7 @@
     "@types/jest": "~27.0.3",
     "@types/react": "~16.9.0",
     "@types/react-dom": "~16.9.0",
-    "cypress": "~8.4.0",
+    "cypress": "~12.0.0",
```
tiborux, line 51: there is new version, should we update to it? 12.1.0. Maybe we can use the minor version instead of the patch?

packages/x-components/cypress.config.ts
```diff
@@ -0,0 +1,45 @@
...
+        'file:preprocessor',
+        createBundler({
+          plugins: [createEsbuildPlugin(config)]
+        })
+      );
+      return config;
+    },
+    specPattern: 'tests/e2e/**/*.feature'
+  },
+  component: {
```
tiborux, line 34: I was watching the migration video that they have, and they suggest renaming all the tests to `*.cy.ts`, but as we are using the specPattern, I'm not sure if this is necessary.

image

Also, they renamed the support file to be `e2e.js` or `component.js`. I suppose that all these changes makes sense if you are not setting the configuration as we do.

image

packages/x-components/package.json
```diff
@@ -53,14 +53,20 @@
...
+    "test:e2e:firefox": "vue-cli-service test:e2e --e2e --browser firefox",
     "test:e2e:ci": "vue-cli-service test:e2e --headless --browser chrome",
     "test": "npm run test:unit && npm run test:e2e:ci",
-    "cypress:open": "cypress open",
-    "cypress:open:component": "cypress open-ct"
+    "cypress:open": "cypress open --e2e --browser chrome",
+    "cypress:open:safari": "cypress open --e2e --browser safari",
+    "cypress:open:firefox": "cypress open--e2e --browser firefox",
+    "cypress:open:component": "cypress open --component --browser chrome",
+    "cypress:open:component:safari": "cypress open --component --browser safari",
```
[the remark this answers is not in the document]
javieri-empathy: Yes, I forgot to remove safari ones which are wrong btw. I want to run tests in both firefox & safari once a day, but as safari requires more steps to configure it, I would do it in another PR.

packages/x-components/cypress.config.ts
javieri-empathy: I would leave it as it is for now. Enough noise in this PR to start renaming more files. We can do that in the future if we need it.
About the support file, we are sharing the same support file from e2e and component. This right now only makes sense to have the same `getByDataTest` command and avoid creating more support and shared files.

mnavarroespinosa (approved):
good job!

tiborux (approved):

PR #820, opened 2022-11-03
Feature/Not show Empathize with no content

annacv:
[EX-7169](https://searchbroker.atlassian.net/browse/EX-7169)

## Motivation and context
No more need to check if there is Empathize content to show it or not, It will do it by itself.
This was a recurrent problem in the setups.

## Type of change

- [x] `Main`
- [ ] Other. Specify:

## How has this been tested?

- Test adding a fake button inside the `predictive-layer` component to remove empathize content and check that it doesn't get opened if the slot is empty while the `isOpen` state also remains `false` for this use case.

- Tests performed according to [testing guidelines](./contributing/tests.md): 

## Checklist:

### DIFF
--- a/packages/x-components/src/x-modules/empathize/components/__tests__/empathize.spec.ts
+++ b/packages/x-components/src/x-modules/empathize/components/__tests__/empathize.spec.ts
@@ -1,53 +1,104 @@
-import { mount } from '@vue/test-utils';
+import { mount, Wrapper } from '@vue/test-utils';
 import Vue from 'vue';
 import BaseEventButton from '../../../../components/base-event-button.vue';
 import { getXComponentXModuleName, isXComponent } from '../../../../components/x-component.utils';
 import { getDataTestSelector, installNewXPlugin } from '../../../../__tests__/utils';
 import Empathize from '../empathize.vue';
+import { empathizeXModule } from '../../x-module';
+import { XPlugin } from '../../../../plugins/index';
+import { XEvent } from '../../../../wiring/index';
 
 describe('testing empathize component', () => {
-  let localVue: typeof Vue;
+  // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+  // @ts-ignore
+  function renderEmpathize({
+    eventsToOpenEmpathize,
+    eventsToCloseEmpathize,
+    template = `<Empathize
+                  :events-to-open-empathize="['TestOpenEvent']"
+                  :events-to-close-empathize="['TestCloseEvent']"
+                >
+                  <template #default>
+                    <span data-test="empathize-content">Empathize</span>
+                  </template>
+                </Empathize>`,
+    eventToSpy
+  }: RenderEmpathizeOptions = {}): RenderEmpathizeAPI {
+    const [, localVue] = installNewXPlugin();
+    XPlugin.registerXModule(empathizeXModule);
 
-  beforeEach(() => {
-    [, localVue] = installNewXPlugin({});
-  });
+    let eventSpy;
+    if (eventToSpy) {
+      eventSpy = jest.fn();
+      XPlugin.bus.on(eventToSpy).subscribe(eventSpy);
+    }
+
+    const wrapper = mount(
+      {
+        props: ['eventsToOpenEmpathize', 'eventsToCloseEmpathize'],
+        components: { Empathize, BaseEventButton },
+        template
+      },
+      {
+        localVue,
+        propsData: {
+          eventsToOpenEmpathize,
+          eventsToCloseEmpathize
+        }
+      }
+    ).findComponent(Empathize);
+
+    return {
+      wrapper,
+      eventSpy
+    };
+  }
 
   it('is an XComponent which has an XModule', () => {
-    const empathize = mount(Empathize, { localVue });
-    expect(isXComponent(empathize.vm)).toEqual(true);
-    expect(getXComponentXModuleName(empathize.vm)).toEqual('empathize');
+    const { wrapper } = renderEmpathize();
+    expect(isXComponent(wrapper.vm)).toEqual(true);
+    expect(getXComponentXModuleName(wrapper.vm)).toEqual('empathize');
   });
 
   it('listens to UserOpenedEmpathize and UserClosedEmpathize by default', async () => {
-    const component = {
-      template: `
-        <Empathize :events-to-open-empathize="['TestOpenEvent']"
-                   :events-to-close-empathize="['TestCloseEvent']">
-          <template #default>
-            <span data-test="empathize-content">Empathize</span>
-          </template>
-        </Empathize>
-      `,
-      components: {
-        BaseEventButton,
-        Empathize
-      },
-      localVue
-    };
-    const componentWrapper = mount(component, {
-      localVue
-    });
+    const { wrapper } = renderEmpathize();
 
-    componentWrapper.vm.$x.emit('TestOpenEvent' as any);
+    wrapper.vm.$x.emit('TestOpenEvent' as any);
     await new Promise(resolve => setTimeout(resolve));
     // await localVue.nextTick();
-    expect(componentWrapper.find('.x-empathize').exists()).toBe(true);
-    expect(componentWrapper.find(getDataTestSelector('empathize-content')).exists()).toBe(true);
+    expect(wrapper.find('.x-empathize').exists()).toBe(true);
+    expect(wrapper.find(getDataTestSelector('empathize-content')).exists()).toBe(true);
 
-    componentWrapper.vm.$x.emit('TestCloseEvent' as any);
+    wrapper.vm.$x.emit('TestCloseEvent' as any);
     await new Promise(resolve => setTimeout(resolve));
     // await localVue.nextTick();
-    expect(componentWrapper.find('.x-empathize').exists()).toBe(false);
-    expect(componentWrapper.find(getDataTestSelector('empathize-content')).exists()).toBe(false);
+    expect(wrapper.find('.x-empathize').exists()).toBe(false);
+    expect(wrapper.find(getDataTestSelector('empathize-content')).exists()).toBe(false);
   });
 });
+
+interface RenderEmpathizeOptions {
+  /**
+   * Array of {@link XEvent | xEvents} to open the empathize.
+   */
+  eventsToOpenEmpathize: XEvent[];
+  /**
+   * Array of {@link XEvent | xEvents} to close the empathize.
+   */
+  eventsToCloseEmpathize: XEvent[];
+  /**
+   * The template to render {@link Empathize} component.
+   */
+  template?: string;
+  /**
+   * An event to spy on.
+   */
+  eventToSpy?: XEvent;
+}
+
+interface RenderEmpathizeAPI {
+  /** The Vue testing utils wrapper for the {@link Empathize} component. */
+  wrapper: Wrapper<Vue>;
+  /** A Jest spy set in the {@link XPlugin} `on` function. */
+  eventSpy?: jest.Mock;
+}
--- a/packages/x-components/src/x-modules/empathize/components/__tests__/empathize.spec.ts
+++ b/packages/x-components/src/x-modules/empathize/components/__tests__/empathize.spec.ts
@@ -22,6 +22,9 @@ describe('testing empathize component', () => {
     const [, localVue] = installNewXPlugin();
     XPlugin.registerXModule(empathizeXModule);
 
+    const parent = document.createElement('div');
+    document.body.appendChild(parent);
+
     const wrapper = mount(
       {
         components: { Empathize },
@@ -29,6 +32,7 @@ describe('testing empathize component', () => {
       },
       {
         localVue,
+        attachTo: parent,
         propsData: {
           eventsToOpenEmpathize,
           eventsToCloseEmpathize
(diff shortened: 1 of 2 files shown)


packages/x-components/src/x-modules/empathize/components/empathize.vue
```diff
@@ -73,6 +76,23 @@
...
+    protected isEmptySlot: boolean | Element = true;
+
+    /**
+     * Detects if the default slot has been replaced.
+     *
+     * @internal
+     * @returns True if the slot is empty, the slot element if it has content.
+     */
+    updated(): boolean | Element {
+      return (this.isEmptySlot = !!this.$refs.isEmptySlot);
```
alvarodE, line 93: Probably you've already thought about it and you know why it wouldn't work but, just in case: Couldn't we check the content of the default slot instead of adding a span and check whether is it visible or not?
```suggestion
      return (this.isEmptySlot = !this.$slots.default?.length);
```

packages/x-components/src/x-modules/empathize/components/empathize.vue
```diff
@@ -73,6 +76,23 @@
...
+     */
+    protected isEmptySlot: boolean | Element = true;
+
+    /**
+     * Detects if the default slot has been replaced.
+     *
+     * @internal
+     * @returns True if the slot is empty, the slot element if it has content.
+     */
+    updated(): boolean | Element {
```
alvarodE, line 92: I think we don't need to return anything on the hook.
```suggestion
    updated(): void {
```

packages/x-components/src/x-modules/empathize/components/empathize.vue
```diff
@@ -83,9 +103,10 @@
      * @internal
      */
     @XOn(component => (component as Empathize).eventsToOpenEmpathize)
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
     open(payload: unknown, metadata: WireMetadata): void {
-      this.changeOpenState(true, metadata);
+      if (!this.$refs.isEmptySlot) {
```
alvarodE, line 107: Is there a reason to not use the calculated `this.isEmptySlot` instead of using a ref?
```suggestion
      if (!this.isEmptySlot) {
```

packages/x-components/src/x-modules/empathize/components/empathize.vue
```diff
@@ -73,6 +76,23 @@
      */
     protected isOpen = false;
 
+    /**
+     * The default slot is empty to avoid opening empathize with no content.
+     *
+     * @internal
+     */
+    protected isEmptySlot: boolean | Element = true;
```
alvarodE, line 84: I'd say that this variable will always be a boolean, and in that case you don't need to type it if in the declaration itself is assigned to `true`:
```suggestion
    protected isEmptySlot = true;
```

packages/x-components/src/x-modules/empathize/components/empathize.vue
annacv: Hi @alvarodE it was the first approach, but the tests failed. I thought that it was because the content didn't get updated (as v-show only was hiding it) and then the condition was always true. It makes sense with the comment you added below about the `isEmptySlot` type. Does it makes sense to you, or I'm losing something?

LuisMartinez15: Using the calculated `isEmptySlot` property over the `ref` results in a buggy behaviour. The `@XOn` listener is triggered before the `updated` lifecycle hook (it is bound to the `created` one), so you end up having to, for example, click twice the `search box`, in order for the `empathize` to finally open.

```suggestion
      if (this.$refs.noContent) {
```

packages/x-components/src/x-modules/empathize/components/empathize.vue
```diff
@@ -119,6 +139,7 @@
     changeOpenState(newOpenState: boolean, metadata: WireMetadata): void {
       if (this.isOpen !== newOpenState) {
         this.isOpen = newOpenState;
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-call
```
LuisMartinez15, line 142: Is this needed? I've tried removing it and `eslint` didn't blame.

```suggestion
```

packages/x-components/src/x-modules/empathize/components/empathize.vue
LuisMartinez15: I think we should take advantage of `vue's` features, in this case of its `$refs` one, instead of relying in `CSS` or query selectors.

Regarding the actual code, you don't need to `return` anything here and a simple `assignment` should be enough.

```suggestion
      this.hasContent = !this.$refs.noContent;
```

packages/x-components/src/x-modules/empathize/components/empathize.vue
```diff
@@ -1,15 +1,18 @@
 <template>
   <component :is="animation">
     <div
-      v-if="isOpen"
+      v-show="isOpen && !isEmptySlot"
```
LuisMartinez15, line 4:
```suggestion
      v-show="isOpen && hasContent"
```

packages/x-components/src/x-modules/empathize/components/empathize.vue
```diff
@@ -73,6 +76,23 @@
      */
     protected isOpen = false;
```
LuisMartinez15, line 78: I'll put it here, as you haven't added the new $ref to the $refs object.

```suggestion

    public $refs!: {
      /** Hidden span as default slot content to check if the empathize has content or not. */
      noContent?: HTMLSpanElement;
    };

```

packages/x-components/src/x-modules/empathize/components/__tests__/empathize.spec.ts
```diff
@@ -1,53 +1,110 @@
-import { mount } from '@vue/test-utils';
+import { mount, Wrapper } from '@vue/test-utils';
 import Vue from 'vue';
-import BaseEventButton from '../../../../components/base-event-button.vue';
 import { getXComponentXModuleName, isXComponent } from '../../../../components/x-component.utils';
 import { getDataTestSelector, installNewXPlugin } from '../../../../__tests__/utils';
 import Empathize from '../empathize.vue';
+import { empathizeXModule } from '../../x-module';
+import { XPlugin } from '../../../../plugins/index';
```
LuisMartinez15, line 7: Avoid importing from barrel files

```suggestion
import { XPlugin } from '../../../../plugins/x-plugin';
```

packages/x-components/src/x-modules/empathize/components/__tests__/empathize.spec.ts
```diff
@@ -1,53 +1,110 @@
...
+import { XEvent } from '../../../../wiring/index';
```
LuisMartinez15, line 8: Same here

```suggestion
import { XEvent } from '../../../../wiring/events.types';
```

packages/x-components/src/x-modules/empathize/components/__tests__/empathize.spec.ts
```diff
@@ -1,53 +1,110 @@
...
+  it('will not open empathize if there is no content to render', async () => {
+    const template = `<Empathize v-bind="$attrs"/>`;
+    const { wrapper } = renderEmpathize({ template });
+
+    wrapper.vm.$x.emit('UserClickedSearchBox');
+    // Fast-forward until all timers have been executed
+    jest.runAllTimers();
+    await wrapper.vm.$nextTick();
+
+    expect(wrapper.find('.x-empathize').exists()).toBe(true);
```
LuisMartinez15, line 63: Why are we mixing finding by `CSS` class and `data-test`? Am I missing something?

```suggestion
    expect(wrapper.find(getDataTestSelector('empathize')).exists()).toBe(true);
```

packages/x-components/src/x-modules/empathize/components/__tests__/empathize.spec.ts
```diff
@@ -1,53 +1,110 @@
...
-    expect(componentWrapper.find(getDataTestSelector('empathize-content')).exists()).toBe(false);
+    const { wrapper } = renderEmpathize();
+
+    wrapper.vm.$x.emit('UserClickedSearchBox');
+    jest.runAllTimers();
+    await wrapper.vm.$nextTick();
+
+    // Both should exist and be visible
+    expect(wrapper.find('.x-empathize').exists()).toBe(true);
+    expect(wrapper.find('.x-empathize').element).toBeVisible();
```
LuisMartinez15, line 76: Same here

```suggestion
    expect(wrapper.find(getDataTestSelector('empathize')).exists()).toBe(true);
    expect(wrapper.find(getDataTestSelector('empathize')).element).toBeVisible();
```

packages/x-components/src/x-modules/empathize/components/__tests__/empathize.spec.ts
```diff
@@ -1,53 +1,110 @@
...
+    expect(wrapper.find(getDataTestSelector('empathize-content')).exists()).toBe(true);
+    expect(wrapper.find(getDataTestSelector('empathize-content')).element).toBeVisible();
+
+    wrapper.vm.$x.emit('UserClosedEmpathize');
+    jest.runAllTimers();
+    await wrapper.vm.$nextTick();
+
+    // Both should exist, as v-show doesn't remove the elements in the DOM, and not be visible
+    expect(wrapper.find('.x-empathize').exists()).toBe(true);
+    expect(wrapper.find('.x-empathize').element).not.toBeVisible();
```
LuisMartinez15, line 86: Same here

```suggestion
    expect(wrapper.find(getDataTestSelector('empathize')).exists()).toBe(true);
    expect(wrapper.find(getDataTestSelector('empathize')).element).not.toBeVisible();
```

packages/x-components/src/x-modules/empathize/components/__tests__/empathize.spec.ts
```diff
@@ -1,53 +1,110 @@
...
   it('is an XComponent which has an XModule', () => {
-    const empathize = mount(Empathize, { localVue });
-    expect(isXComponent(empathize.vm)).toEqual(true);
-    expect(getXComponentXModuleName(empathize.vm)).toEqual('empathize');
+    const { wrapper } = renderEmpathize();
+    expect(isXComponent(wrapper.vm)).toEqual(true);
+    expect(getXComponentXModuleName(wrapper.vm)).toEqual('empathize');
+  });
+
+  it('will not open empathize if there is no content to render', async () => {
```
LuisMartinez15, line 54: You can safely remove the `async` modifier as it is not needed for this test

```suggestion
  it('will not open empathize if there is no content to render', () => {
```

packages/x-components/src/x-modules/empathize/components/__tests__/empathize.spec.ts
```diff
@@ -1,53 +1,110 @@
...
+    const template = `<Empathize v-bind="$attrs"/>`;
+    const { wrapper } = renderEmpathize({ template });
+
+    wrapper.vm.$x.emit('UserClickedSearchBox');
+    // Fast-forward until all timers have been executed
+    jest.runAllTimers();
+    await wrapper.vm.$nextTick();
```
LuisMartinez15, line 61: You can safely remove these lines as they're not needed.

```suggestion
```

packages/x-components/src/x-modules/empathize/components/__tests__/empathize.spec.ts
```diff
@@ -1,53 +1,110 @@
...
+
+    expect(wrapper.find('.x-empathize').exists()).toBe(true);
+    expect(wrapper.find(getDataTestSelector('empathize-content')).exists()).toBe(false);
   });
 
   it('listens to UserOpenedEmpathize and UserClosedEmpathize by default', async () => {
```
LuisMartinez15, line 67: Maybe we can merge the previous test with this one? Also, on a side note, I don't know why we need to use "the hacky" approach of using `jest timers`, but I think this component would fit better in the `cypress component testing` environment than here, as we are checking the existence of `HTML` elements and their visibility. You don't have to change it, but just thinking out loud.

LuisMartinez15 (changes requested):
good job!

packages/x-components/src/x-modules/empathize/components/__tests__/empathize.spec.ts
annacv: Hi @LuisMartinez15, when `changeOpenState ` is fired, there's a Debounce applied (although the debounce time is set to 0). The event is not fired when `Empathize` has no content, so you are right, we didn't need to use `jest timers` in that case. But in this situation I thought it was necessary, otherwise tests failed. Is that the right approach, or is there something wrong I missed?

LuisMartinez15: I didn't notice the `debounce`. I guess that's why we need to run the timers.

packages/x-components/src/x-modules/empathize/components/empathize.vue
LuisMartinez15: I don't see this being committed, did you forget about it? 😅

LuisMartinez15 (changes requested):

packages/x-components/src/x-modules/empathize/components/empathize.vue
annacv: 🙈 Thks @LuisMartinez15 !

LuisMartinez15 (approved):
good job!

tiborux (approved):

PR #822, opened 2022-11-03
feat(tagging): track banner clicks

diegopf:
EX-7339 

This change is a request from motive where the banners have tagging information. Currently in platform we don't have tagging for banners, so if you wan to test this locally you can by overriding the bannerSchema and mocking the tagging:

```
bannerSchema.$override<PlatformBanner, Partial<Banner>>({
  tagging: {
    click: {
      url: () => 'https://localhost',
      params: () => Object.assign({})
    }
  }
});
```

### DIFF
--- a/packages/x-components/src/x-modules/search/components/__tests__/banner.spec.ts
+++ b/packages/x-components/src/x-modules/search/components/__tests__/banner.spec.ts
@@ -61,6 +61,25 @@ describe('testing Banner component', () => {
 
     expect(wrapper.get(getDataTestSelector('banner')).text()).toEqual('Search UIs');
   });
+
+  // eslint-disable-next-line max-len
+  it('emits UserClickedABanner when the user clicks in the left, middle or right button on the component', () => {
+    const listener = jest.fn();
+    const banner = createBannerStub('banner');
+    const { wrapper } = renderBanner({ banner });
+    wrapper.vm.$x.on('UserClickedABanner').subscribe(listener);
+
+    wrapper.trigger('click');
+    expect(listener).toHaveBeenNthCalledWith(1, banner);
+
+    wrapper.trigger('click', { button: 1 });
+    expect(listener).toHaveBeenNthCalledWith(2, banner);
+
+    wrapper.trigger('click', { button: 2 });
+    expect(listener).toHaveBeenNthCalledWith(3, banner);
+
+    expect(listener).toHaveBeenCalledTimes(3);
+  });
 });
 
 interface RenderBannerOptions {
--- a/packages/x-components/src/x-modules/tagging/wiring.ts
+++ b/packages/x-components/src/x-modules/tagging/wiring.ts
@@ -147,10 +147,10 @@ export const trackBannerClickedWire = createTrackWire('click');
 export const trackAddToCartWire = createTrackWire('add2cart');
 
 /**
- * Factory helper to create a wire for the track of a taggable object.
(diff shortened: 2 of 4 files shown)


packages/x-components/src/x-modules/tagging/wiring.ts
```diff
@@ -130,24 +130,31 @@ export const setQueryTaggingInfo = moduleDebounce(
...
  */
-export const trackAddToCartWire = createTrackResultWire('add2cart');
+export const trackAddToCartWire = createTrackWire('add2cart');
 
 /**
- * Factory helper to create a wire for the track of a result.
+ * Factory helper to create a wire for the track of a taggable object.
  *
  * @param property - Key of the tagging object to track.
  * @returns A new wire for the given property of the result tagging.
```
AdrianArenal, line 153: One detail just for consistency
```suggestion
 * @returns A new wire for the given property of the taggable object.
```

btw, in our side we were calling it 'taggable element' instead of 'taggable object', but we are fine with both 👍🏽

packages/x-components/src/x-modules/search/components/__tests__/banner.spec.ts
```diff
@@ -61,6 +61,25 @@ describe('testing Banner component', () => {
...
   });
+
+  // eslint-disable-next-line max-len
+  it('emits UserClickedABanner when the user clicks in the left, middle or right button on the component', () => {
+    const listener = jest.fn();
+    const banner = createBannerStub('banner');
+    const { wrapper } = renderBanner({ banner });
+    wrapper.vm.$x.on('UserClickedABanner').subscribe(listener);
+
+    wrapper.trigger('click');
```
AdrianArenal, line 72: quick question, why we don't need an `await` before trigger here? We tend to use it and checking the [docs](https://v1.test-utils.vuejs.org/guides/dom-events.html) they're using it also since its returning a Promise 🤔   
Maybe this can lead to unexpected async errors

javieri-empathy: The click event is emitted immediately. We don't need to wait for another re-render which is what the await. If it changes and is emitted asynchronously somehow I would like to be informed.

javieri-empathy: Salu2

AdrianArenal (dismissed):

packages/x-components/src/x-modules/tagging/wiring.ts
```diff
@@ -130,24 +130,31 @@ export const setQueryTaggingInfo = moduleDebounce(
...
+ * Factory helper to create a wire for the track of a taggable element.
  *
  * @param property - Key of the tagging object to track.
- * @returns A new wire for the given property of the result tagging.
+ * @returns A new wire for the given property of the taggable element.
  *
  * @internal
  */
-function createTrackResultWire(property: keyof Tagging): Wire<Result> {
+function createTrackWire(property: keyof Tagging): Wire<Taggable> {
```
joseacabaneros, line 157: Could we export this function to be used outside X? At Motive, we also track things like clicks in suggestions or when the user opens the search-layer. We have a copy of this function to create wires for tacking that. It would be awesome if we could use this instead of a copy.

AdrianArenal (approved):

joseacabaneros (approved):

ajperezbau (approved):

PR #823, opened 2022-11-03
feat(banners): add tagging info banners in platform adapter

diegopf:
EX-7355


mnavarroespinosa (commented):
What would you think of updating the mocked search response (`packages/x-adapter-platform/src/__tests__/__fixtures__/search.response.ts`) with the tagging property? This way we could check that the mapper is working properly.

ajperezbau:
> What would you think of updating the mocked search response (`packages/x-adapter-platform/src/__tests__/__fixtures__/search.response.ts`) with the tagging property for the banners? This way we could check that the mapper is working properly.

Yep, I agree with that, but, do we know where will the API return the tagging for the banners?
I suppose that they will do the same as for the result, and return it in each banner element inside `banner.content`, right?

ajperezbau (commented):
Good job 👌

mnavarroespinosa (approved):
good job!

ajperezbau (approved):
Good job 👌

PR #829, opened 2022-11-07
feat: place modal under selected element

tiborux:
EX-5023

To test it you can a fake element in the App.vue like `<div class="header"></div>` and then add this selector in the main-modal in Home.vue `<MainModal :animation="modalAnimation" clientHeaderSelector=".header">`

You can do the same by installing locally this build in the archetype.


packages/x-components/src/components/modals/base-events-modal.vue
```diff
@@ -4,6 +4,7 @@
     @focusin:body="emitBodyClickEvent"
     :animation="animation"
     :open="isOpen"
+    v-bind="$attrs"
```
tiborux, line 7: I am not sure if this is right, I don't like to pass the same prop between multiple components to send it to the nested one.

mnavarroespinosa: I agree but I don't see another way apart from injecting it.

mnavarroespinosa (commented):
good job!

packages/x-components/src/components/modals/base-modal.vue
```diff
@@ -230,7 +277,9 @@
 ## Examples
 
 The `BaseModal` is a simple component that serves to create complex modals. Its open state has to be
-passed via prop. It also accepts an animation to use for opening & closing.
+passed via prop. There is a prop, `clientHeaderSelector`, used to place the modal under some element
```
LuisMartinez15, line 280:
```suggestion
passed via prop. There is a prop, `referenceSelector`, used to place the modal under some element
```

packages/x-components/src/components/modals/base-modal.vue
```diff
@@ -87,6 +100,40 @@
...
+     * @internal
+     */
+    @Debounce(100, { leading: true })
+    updatePosition(): void {
+      this.$refs.modal.style.top = this.clientHeaderSelector
+        ? `${
+            this.selectedElement.getBoundingClientRect().height +
+            this.selectedElement.getBoundingClientRect().y
+          }px`
+        : '0';
```
LuisMartinez15, line 136: Regarding this implementation. We are only allowing to position an element below another one. Should we make this positioning also configurable? I don't know what exactly the purpose of the task was, but it could be something to consider. On a side note, I think we should check if the `referenceElement` exists, rather than if a `referenceSelector` has been defined. Maybe there's a `referenceSelector` defined but no element matches it in the DOM, resulting in a `Reference error` due to `referenceSelector` not being defined.

Looking at the code, calling `getBoundingClientRect()` method causes a layout `reflow`. We are calling twice consecutively, which I assume are `batched` and result in a single call. However, there are other approaches to avoid calling `getBoundingClientRect()` or to reduce its `reflow` penalty. For further information regarding alternatives, you can check this [post](https://toruskit.com/blog/how-to-get-element-bounds-without-reflow/). I don't think we have to refactor it now, but it might be worth to consider it, at least for the future.

```suggestion
      const { height, y } = this.selectedElement?.getBoundingClientRect() ?? { height: 0, y: 0 };
      this.$refs.modal.style.top = `${height + y}px`;
```

packages/x-components/src/components/modals/base-modal.vue
```diff
@@ -239,7 +288,12 @@ is open.
 <template>
   <div>
     <button @click="open = true">Open modal</button>
-    <BaseModal animation="fadeAndSlide" :open="open" @click:overlay="open = false">
+    <BaseModal
+      animation="fadeAndSlide"
```
LuisMartinez15, line 292:
```suggestion
      :animation="fadeAndSlide"
```

packages/x-components/src/components/modals/base-modal.vue
```diff
@@ -239,7 +288,12 @@ is open.
...
+      :open="open"
+      @click:overlay="open = false"
+      clientHeaderSelector=".header"
```
LuisMartinez15, line 295:
```suggestion
      referenceSelector=".header"
```

packages/x-components/tests/unit/base-modal.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  return {
+    getBaseModal() {
+      return cy.getByDataTest('modal');
+    }
+  };
+}
+
+describe('Testing base modal', () => {
+  it('places the modal at top if there is not a selector', () => {
```
LuisMartinez15, line 70:
```suggestion
  it('places the modal at the top of the page if referenceSelector is not defined', () => {
```

packages/x-components/tests/unit/base-modal.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+    const { getBaseModal } = renderBaseModal();
+    getBaseModal().should($modal => expect($modal.position().top).to.be.eq(0));
+  });
+
+  it('places the modal under the selector passed by prop', () => {
```
LuisMartinez15, line 75:
```suggestion
  it('places the modal under the defined referenceSelector', () => {
```

packages/x-components/tests/unit/base-modal.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+
+  // eslint-disable-next-line max-len
+  it('calculates where to place the modal even if there are more elements on the selected element', () => {
+    const preHeaderHeight = 300;
+    const options = {
+      clientHeaderHeight: { desktop: 100 },
+      clientHeaderSelector: '.header',
+      template: `
+                  <div class="wrapper">
+                    <div class="pre-header" style="height: 300px"></div>
```
LuisMartinez15, line 122:
```suggestion
                    <div class="pre-header" style="height: ${preHeaderHeight}px"></div>
```

packages/x-components/tests/unit/base-modal.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * Renders an {@link BaseModal} component with the provided options.
+ *
+ * @param  options - Options to render {@link BaseModal} with.
+ * @returns Helper methods for the rendered {@link BaseModal}.
+ */
+function renderBaseModal({
+  clientHeaderSelector = '',
+  clientHeaderHeight = {},
```
LuisMartinez15, line 16:
```suggestion
  clientHeaderHeight = {
    desktop: 0,
    tablet: 0,
    mobile: 0
  },
```

packages/x-components/tests/unit/base-modal.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+
+interface RenderBaseModalOptions {
+  /**
+   * The selector under which the modal is to be placed.
+   */
+  clientHeaderSelector?: string;
+  /**
+   * The client header height.
+   */
+  clientHeaderHeight?: Record<string, number>;
```
LuisMartinez15, line 143:
```suggestion
  clientHeaderHeight?: {
    desktop: number;
    tablet: number;
    mobile: number;
  };
```

packages/x-components/tests/unit/base-modal.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+  });
+});
+
+interface RenderBaseModalOptions {
+  /**
+   * The selector under which the modal is to be placed.
+   */
+  clientHeaderSelector?: string;
+  /**
+   * The client header height.
```
LuisMartinez15, line 141:
```suggestion
   * The client header height for each device.
```

packages/x-components/tests/unit/base-modal.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+    const options = {
+      clientHeaderHeight: {
+        mobile: 100,
+        tablet: 400,
+        desktop: 200
+      },
+      clientHeaderSelector: '.header'
+    };
+    const { getBaseModal } = renderBaseModal(options);
+    // Starts in desktop
```
LuisMartinez15, line 96: I'd prefer to change this code comments in favour of logs, so we are also aware of this changes when seeing the test in action.

```suggestion
    cy.log('Viewport in desktop resolution');
```

packages/x-components/tests/unit/base-modal.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+    getBaseModal().should($modal =>
+      expect($modal.position().top).to.be.eq(options.clientHeaderHeight.desktop)
+    );
+
+    // Change to mobile
```
LuisMartinez15, line 101:
```suggestion
    cy.log('Resizing viewport to mobile resolution');
```

packages/x-components/tests/unit/base-modal.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+    cy.viewport('iphone-8');
+    getBaseModal().should($modal =>
+      expect($modal.position().top).to.be.eq(options.clientHeaderHeight.mobile)
+    );
+
+    // Change to tablet
```
LuisMartinez15, line 107:
```suggestion
    cy.log('Resizing viewport to tablet resolution');
```

packages/x-components/tests/unit/base-modal.spec.ts
```diff
@@ -0,0 +1,153 @@
...
+ * @param  options - Options to render {@link BaseModal} with.
+ * @returns Helper methods for the rendered {@link BaseModal}.
+ */
+function renderBaseModal({
+  clientHeaderSelector = '',
+  clientHeaderHeight = {},
+  template = `
+      <div class="wrapper">
+        <div class="header"></div>
+        <BaseModal :clientHeaderSelector="clientHeaderSelector" :open="open" />
```
LuisMartinez15, line 20: As you are not testing different values of the `open` property, shouldn't we just pass `true` here and remove it from the `props` section? We either test it or remove it, as you prefer.

LuisMartinez15 (changes requested):
good job!

packages/x-components/src/components/modals/base-modal.vue
```diff
@@ -87,6 +100,40 @@
...
+          if (this.clientHeaderSelector) {
+            const element = document.querySelector(this.clientHeaderSelector);
+            if (element) {
+              this.selectedElement = element;
+              resizeObserver.observe(element);
+            }
+          }
+        },
+        { immediate: true }
+      );
```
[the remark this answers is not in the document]
tiborux: `.disconnect()` i was 10 minutes to find this mistake 😂

packages/x-components/src/components/modals/base-modal.vue
tiborux: At this point we want to position the modal under the selected element, maybe in the future, we can add the other logic.

About the other point, I am not sure if its something that we have to be worried about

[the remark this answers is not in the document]
tiborux: yep, you are right

LuisMartinez15 (approved):
👏🏿

mnavarroespinosa (approved):
good job!

PR #830, opened 2022-11-08
feat(design-system): deprecate old design system button styles

ajperezbau:
Keep in mind these changes from the task refinement:

- This PR only deprecates the `button` styles, the related styles aren't deprecated because we need them if we import and use the deprecated styles file.
- The deprecated styles will not be split by component, the `deprecated-full-theme.css` file will contain all the styles.

BREAKING CHANGE: the old design system button styles are now located in the `deprecated-full-theme.css`, if a setup is not using the new design system, it must import this file.

EX-7140


mnavarroespinosa (approved):
good job!

PR #832, opened 2022-11-08
fix(design-system): Change the xds button size variants.

tajespasarela:
EX-7369

This PR:
- Changes the naming of the XDS Button from [`xs`, `sm`, `md`, `lg`] to: [ `sm`, `md`, `lg`, `xl`] to make it coherent with the next icon XDS component.
- The default size would be the new `md` so the buttons will be smaller by default.
- Also refactored the way default styles reusing the corresponding variant.
- Fixed the colors in the button showcase to use the new colors naming.


tiborux (approved):

annacv (approved):
I only have one doubt: for rounded button, will we create another layout?

PR #834, opened 2022-11-09
refactor(design-system): Refactor showcase component to make it reusable.

tajespasarela:
EX-7368

This PR:
- Extracts the common part of the `XdsButton` showcase component to a new `XdsBaseShowcase` component to reuse with new components like the icon.

### DIFF
--- a/packages/x-tailwindcss/demo/src/components/xds-base-showcase.vue
+++ b/packages/x-tailwindcss/demo/src/components/xds-base-showcase.vue
@@ -1,0 +1,58 @@
+<template>
+  <div class="x-flex x-flex-col x-gap-32">
+    <h1 class="x-text-lg">{{ title }}</h1>
+    <div v-for="(classes, section) in sections" :key="section" class="x-flex x-flex-row x-gap-16">
+      <h2 class="x-text-md x-w-128 x-text-right x-flex-none">{{ section }}</h2>
+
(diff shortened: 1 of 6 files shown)


packages/x-tailwindcss/demo/src/components/xds-base-showcase.vue
```diff
@@ -0,0 +1,58 @@
...
+     */
+    @Prop({ required: true })
+    public title!: string;
+
+    /**
+     * Copies the classList of an HTML Element to the clipboard.
+     *
+     * @param event - The MouseEvent to get the HTML Element from.
+     */
+    protected copyCssClassesToClipboard(event: MouseEvent): void {
```
tiborux, line 43: Can we add a tooltip or something to know that the styles are being copied to the clipboard?

annacv (commented):
I would only add spacing to the whole `xds-showcase` container or to the `xds-base-showcase` section.

tiborux (approved):
gj

packages/x-tailwindcss/demo/src/components/xds-base-showcase.vue
tajespasarela: done

annacv (approved):
Spacing is not core, so approved 🚀

tajespasarela:
> Spacing is not core, so approved 🚀

That can be done in the project where used (X-Components, X-Archetype, etc.)

PR #836, opened 2022-11-09
feat(x-installer): allow retrieving the target element for application mounting based on the snippet configuration

joseacabaneros:
[MOTPRD-9239](https://searchbroker.atlassian.net/browse/MOTPRD-9239)

`domElement` in `InstallXOptions` now allows a function in addition to an element selector or a DOM element. With this change, the target element where the Vue application is gonna mount, can be a dynamic element based on the `SnippetConfig`.

At Motive, we are working on encapsulating motive-x under a shadowDOM only if an `isolated` property is activated through the `SnippetConfig`. So at runtime, we want to decide if we have to create a DOM tree for the `shadowRoot` or, on the other hand, we have to create a simple `div` based in the snippet configuration. Something like that: 

```TS
function getDomElement({ isolated }: SnippetConfig): Element {
  const xDomElement = document.createElement('div');

  if (isolated) {
    const container = document.createElement('div');
    const shadowRoot = container.attachShadow({ mode: 'closed' });
    shadowRoot.appendChild(xDomElement);
    document.body.appendChild(container);
    setStyleHost(shadowRoot);
  } else {
    document.body.appendChild(xDomElement);
    setStyleHost(document.head);
  }

  return xDomElement;
}
```

I also improved a bit the coverage percentage of the `x-installer` file 😉 
From:
Screenshot 2022-11-09 at 2 24 17 PM

To:
Screenshot 2022-11-09 at 2 22 29 PM

### DIFF
--- a/packages/x-components/src/x-installer/x-installer/x-installer.ts
+++ b/packages/x-components/src/x-installer/x-installer/x-installer.ts
@@ -1,4 +1,4 @@
-import { forEach } from '@empathyco/x-utils';
+import { forEach, isFunction } from '@empathyco/x-utils';
 import Vue, { PluginObject, VueConstructor } from 'vue';
 import { BaseXBus } from '../../plugins/x-bus';
 import { XBus } from '../../plugins/x-bus.types';
@@ -319,7 +319,7 @@ export class XInstaller {
    * @internal
    */
   protected getMountingTarget(domElement?: InstallXOptions['domElement']): Element {
-    if (typeof domElement === 'function') {
+    if (isFunction(domElement)) {
       domElement = domElement(this.snippetConfig!);
     }
     if (typeof domElement === 'string') {
@@ -331,10 +331,7 @@ export class XInstaller {
       }
       return target;
     }
-    if (domElement !== undefined) {
-      return domElement;
-    }
-    return document.body.appendChild(document.createElement('div'));
+    return domElement ?? document.body.appendChild(document.createElement('div'));
   }
 
   /**
--- a/packages/x-components/src/x-installer/x-installer/x-installer.ts
+++ b/packages/x-components/src/x-installer/x-installer/x-installer.ts
@@ -306,29 +306,35 @@ export class XInstaller {
   }
 
   /**
-   * It returns the HTML element to mount the Vue Application. If the `domElement` parameter in the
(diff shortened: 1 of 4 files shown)


packages/x-components/src/x-installer/x-installer/x-installer.ts
```diff
@@ -306,29 +306,35 @@ export class XInstaller {
...
+          `XComponents app couldn't be mounted: Element "${domElement}" couldn't be found`
         );
       }
       return target;
-    } else if (elementOrSelector !== undefined) {
-      return elementOrSelector;
-    } else {
-      return document.body.appendChild(document.createElement('div'));
     }
+    if (domElement !== undefined) {
```
LuisMartinez15, line 334: We can remove the `!== undefined`, as it can only be either an `Element` or `undefined`.

```suggestion
    if (domElement) {
```

packages/x-components/src/x-installer/x-installer/x-installer.ts
```diff
@@ -306,29 +306,35 @@ export class XInstaller {
...
    *
-   * @returns The Element to use as mounting point for the Vue App.
+   * @returns The Element to use as mounting target for the Vue Application.
    * @internal
    */
-  protected getMountingTarget(elementOrSelector?: string | Element): Element {
-    if (typeof elementOrSelector === 'string') {
-      const target = document.querySelector(elementOrSelector);
+  protected getMountingTarget(domElement?: InstallXOptions['domElement']): Element {
+    if (typeof domElement === 'function') {
```
LuisMartinez15, line 322: It is not necessary and you probably weren't aware of it, but we have an `isFunction` `typeguard` helper 😅 

I don't leave a code suggestion cause you will have to import it anyways, but feel free to do it or take a look at it in case you are curious (reinventing the wheel 💪🏿 ).

LuisMartinez15 (approved):
good job and thanks for your contribution! 👏🏿 👏🏿

packages/x-components/src/x-installer/x-installer/x-installer.ts
joseacabaneros: Overengineering is always in my team 😄 
Changed for the typeguard.

tajespasarela (approved):

tajespasarela:
Good work! 
image

PR #837, opened 2022-11-10
build(dev-deps): sync `vite` version to 3.2.0

javieri-empathy:
Syncs the `vite` version to 3.2.0 version in every package. The other way we might end up with 2 vite installations with is not ideal.


herrardo (approved):
Thanks mate!

mnavarroespinosa (approved):

PR #838, opened 2022-11-10
feat(design-system): add `suggestion` component to the XDS

mnavarroespinosa:
[EX-7254](https://searchbroker.atlassian.net/browse/EX-7254)

- Adds the `suggestion` component to the XDS.
- Adds the component usage to the XDS Showcase.

There are some differences with the [design](https://www.figma.com/file/9a1DMgqIHt5ZZBvS2waaOs/X-DS?node-id=2118%3A39371) as I could not find a solution that was not overcomplicating things (any idea is welcome :bow2: )

- The `default` and `tag` variants are both aligned to the center vertically. I could not find a way to align the `default` variant as specified in the design keeping the icon at the top when the text is wrapped:

image

- The `tag` variant won't have text ellipsis by default as it is not compatible with the flexbox usage and it is preferable to have the advantages of using flexbox.

To test it run: `vite serve demo` from `x-tailwindcss` package dir. If you check it from the XDS view in `x-components` demo, you'll see it messed up as styles are mixed due to both old DS and XDS being imported.

### DIFF
--- a/packages/x-tailwindcss/demo/src/components/xds-suggestion.vue
+++ b/packages/x-tailwindcss/demo/src/components/xds-suggestion.vue
@@ -75,9 +75,9 @@
 
     @Prop({
       default: () => [
-        'x-suggestion x-suggestion-accent x-suggestion-md',
+        'x-suggestion x-suggestion-success x-suggestion-md',
         'x-suggestion x-suggestion-matching x-suggestion-auxiliary x-suggestion-md',
-        'x-suggestion-tag x-suggestion-success x-suggestion-lg'
+        'x-suggestion-tag x-suggestion-error-50 x-suggestion-lg'
       ]
     })
     public combinations!: string[];
--- a/packages/x-tailwindcss/src/x-tailwind-plugin/components/suggestion/default.ts
+++ b/packages/x-tailwindcss/src/x-tailwind-plugin/components/suggestion/default.ts
@@ -39,9 +39,23 @@ export function suggestionDefault(helpers: TailwindHelpers) {
       fontWeight: theme('fontWeight.regular')
     },
 
-    // Enable Align icon
+    /**
+     * Enable `enableIconOffset` icon alignment flag.
+     *
+     * Suggestions should have the icon aligned with the first text
+     * line.
+     *
+     * As the suggestion is a flex container with the items aligned
+     * to the baseline, we have to vertically move the icon.
+     *
+     * The `--iconVerticalOffset` variable will contain the offset which we have
+     * to use to translate the icon.
+     *
+     */
     '& *': {
       '--enableIconOffset': 'var(--ON)',
+      // To half the icon's height, we subtract half the font's x-height (1ex).
+      // For some reason, we have to adjust it by subtracting one pixel additionally.
       '--iconVerticalOffset': 'calc(50% - 0.5ex - 1px)'
     }
   };
--- a/packages/x-tailwindcss/src/x-tailwind-plugin/components/variables/index.ts
+++ b/packages/x-tailwindcss/src/x-tailwind-plugin/components/variables/index.ts
@@ -3,6 +3,20 @@ import { TailwindHelpers } from '../../../types';
 /**
  * Returns the CSS global variables to be used across the components.
  *
+ * --ON and --OFF variables are available for feature state switching.
+ *
+ * @example
+ * - Define a CSS variable for the feature flag with a default value (--ON or --OFF).
+ *  `--myFeature: var(--OFF);`
+ *
+ * - Two possible usages:
+ *
+ *  1. Use the feature flag using var() function providing a fallback as second argument.
+ *  `background: var(--myFeature, red);`: Fallback value won't be applied if `--myFeature` is off.
+ *
+ *  2. Use the feature flag using var() function followed by a fallback value.
+ *  `background: var(--myFeature) red;`: Fallback value will be applied if `--myFeature` is off.
+ *
  * @param helpers - The {@link TailwindHelpers} to generate CSS.
  * @returns The {@link CssStyleOptions} for the component.
  */
--- a/packages/x-tailwindcss/src/x-tailwind-plugin/components/icon/default.ts
+++ b/packages/x-tailwindcss/src/x-tailwind-plugin/components/icon/default.ts
@@ -19,9 +19,14 @@ export function iconDefault(helpers: TailwindHelpers) {
     aspectRatio: '1/1',
     vectorEffect: 'non-scaling-stroke',
     flex: '0 0 auto',
+
+    // Disable flag by default, it can be enabled from other components.
     '--enableIconOffset': 'var(--OFF)',
+    // When flag is enabled, `fontSize` won't be changed.
     fontSize: 'var(--enableIconOffset) var(--fontSize)',
+    // When flag is enabled, `translateY` will be applied.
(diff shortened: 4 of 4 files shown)


packages/x-tailwindcss/demo/src/components/xds-suggestion.vue
```diff
@@ -0,0 +1,83 @@
+<template>
+  <XdsBaseShowcase
+    #default="{ cssClass, section, copyCssClassesToClipboard, removeClassPrefix }"
+    title="Suggestion"
+    :sections="sections"
+  >
+    <div
```
ajperezbau, line 7: Should we change that `div` to be a `button` element?

ajperezbau (changes requested):
Nice 👌

packages/x-tailwindcss/demo/src/components/xds-suggestion.vue
```diff
@@ -0,0 +1,83 @@
...
+    public base!: string;
+
+    @Prop({
+      default: () => ['x-suggestion-md', 'x-suggestion-lg']
+    })
+    public sizes!: string[];
+
+    @Prop({
+      default: () => [
+        '',
```
tiborux, line 48: why '' as value?

packages/x-tailwindcss/demo/src/components/xds-suggestion.vue
```diff
@@ -0,0 +1,83 @@
...
+    public colors!: string[];
+
+    @Prop({ default: () => ['x-suggestion-tag'] })
+    public tag!: string;
+
+    @Prop({ default: () => ['x-suggestion-matching'] })
+    public matching!: string[];
+
+    @Prop({
+      default: () => ['x-suggestion-tag x-suggestion-md', 'x-suggestion-tag x-suggestion-lg']
```
tiborux, line 68: change one combination to use different colors

packages/x-tailwindcss/demo/src/components/xds-suggestion.vue
```diff
@@ -0,0 +1,83 @@
...
+    public combinations!: string[];
+
+    protected get sections(): ShowcaseSections {
+      return {
+        Default: [this.base],
+        Colors: this.colors.map(addParentClasses(this.base)),
+        Sizes: this.sizes.map(addParentClasses(this.base)),
+        Tag: this.colors.map(addParentClasses(this.base, this.tag)),
+        Combinations: this.combinations.map(addParentClasses(this.base)),
+        Matching: this.matching.map(addParentClasses(this.base))
```
tiborux, line 79:
```suggestion
        Matching: this.matching.map(addParentClasses(this.base))
        Combinations: this.combinations.map(addParentClasses(this.base)),
```

packages/x-tailwindcss/src/x-tailwind-plugin/components/suggestion/default.ts
```diff
@@ -0,0 +1,41 @@
...
+// eslint-disable-next-line  @typescript-eslint/explicit-function-return-type
+export function suggestionDefault(helpers: TailwindHelpers) {
+  const { theme } = helpers;
+  return {
+    fontFamily: theme('fontFamily.primary'),
+    fontWeight: theme('fontWeight.regular'),
+
+    display: 'flex',
+    justifyContent: 'center',
+    alignItems: 'center',
```
tiborux, line 19:
```suggestion
    alignItems: 'flex-start',
```

packages/x-tailwindcss/demo/src/components/xds-suggestion.vue
mnavarroespinosa: good point! I've added it

packages/x-tailwindcss/demo/src/components/xds-suggestion.vue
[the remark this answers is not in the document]
mnavarroespinosa: Sure, I've changed it!

packages/x-tailwindcss/src/x-tailwind-plugin/components/suggestion/default.ts
```diff
@@ -0,0 +1,41 @@
...
+ * Returns the default styles for component `suggestion`.
+ *
+ * @param helpers - The {@link TailwindHelpers} to generate CSS.
+ * @returns The {@link CssStyleOptions} for the component.
+ */
+// eslint-disable-next-line  @typescript-eslint/explicit-function-return-type
+export function suggestionDefault(helpers: TailwindHelpers) {
+  const { theme } = helpers;
+  return {
+    fontFamily: theme('fontFamily.primary'),
```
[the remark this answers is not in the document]
mnavarroespinosa: Good catch! I've updated it and also other references to `fontFamily.primary`

packages/x-tailwindcss/demo/src/components/xds-suggestion.vue
mnavarroespinosa: I added it so that an example with no color applied is displayed for the other sections whose examples iterate over the colors.

packages/x-tailwindcss/src/x-tailwind-plugin/components/suggestion/default.ts
mnavarroespinosa: In the end, this has been changed to `baseline` for the icon to be aligned.

packages/x-tailwindcss/src/x-tailwind-plugin/components/suggestion/tag.ts
```diff
@@ -0,0 +1,48 @@
...
+      backgroundColor: theme('colors.neutral.0'),
+      borderWidth: theme('borderWidth.1'),
+      borderColor: 'currentColor',
+
+      '&:hover': {
+        backgroundColor: theme('colors.neutral.10'),
+        textDecoration: 'none'
+      },
+
+      '&.suggestion-md': {
```
tiborux, line 30: I think that this will cause the problem that Ivan said, if you use sugggestion-md, the styles related with tag will be loaded even if you are not using them

tiborux: @tajespasarela said that this could be solved using variables or something. For the moment, we will leave it as it is.

packages/x-tailwindcss/src/x-tailwind-plugin/components/variables/index.ts
```diff
@@ -0,0 +1,18 @@
+import { TailwindHelpers } from '../../../types';
+
+/**
+ * Returns the component `icon` CSS.
```
tiborux, line 4: Update this comment

tiborux (changes requested):
gj

ajperezbau:
Should we add some documentation to the fix to align the icons?

ajperezbau (approved):
Good job! 👌

packages/x-tailwindcss/demo/src/components/xds-suggestion.vue
```diff
@@ -0,0 +1,97 @@
+<template>
```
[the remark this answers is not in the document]
mnavarroespinosa: Good point! I've modified a combination for it.

tiborux (approved):

ajperezbau (approved):

PR #846, opened 2022-11-11
feat: make base dropdown based components accesible

mavmaf:
Accessibility Check view has `BaseColumnPickerDropdown` and `SortDropdown`.
~~Check they fulfil accessibility requirements, as per [this](https://searchbroker.atlassian.net/wiki/spaces/~5fae2ee07cfc240071ff281b/pages/4041867471/Menu+Button) and [this](https://searchbroker.atlassian.net/wiki/spaces/~5fae2ee07cfc240071ff281b/pages/4042653758/Menu) documents.~~

Please check:
- https://www.w3.org/WAI/ARIA/apg/patterns/combobox/
- https://www.w3.org/WAI/ARIA/apg/patterns/listbox/

You'll see a label for the BaseColumnPickerDropdown. It's there just to show how `aria-label` can be replaced


packages/x-components/src/components/base-dropdown.vue
```diff
@@ -31,10 +32,11 @@
         @keydown.end="highlightLastItem"
         @keydown.esc="close"
         @keydown.home="highlightFirstItem"
-        :aria-expanded="isOpen.toString()"
         class="x-dropdown__items-list"
-        role="listbox"
+        role="menu"
         tabIndex="0"
+        :aria-activedescendant="value"
```
javieri-empathy, line 38: [Shouldn't the `aria-activedescendant` receive  an `id`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-activedescendant#values)? 
If that's the case we should build a way to auto-generate global unique ids for each one of the elements.

packages/x-components/src/views/accessibility-check.vue
```diff
@@ -1,118 +1,46 @@
 <template>
```
javieri-empathy, line 1: Don't you want to keep the previous elements? We can build a sub-router if you want to have your own a11y playgrounds. It can be helpful to have safe places to check all of these behaviors.

javieri-empathy (changes requested):

packages/x-components/src/components/base-dropdown.vue
```diff
@@ -45,7 +47,7 @@
             :class="itemsCSSClasses[index]"
             class="x-dropdown__item"
             data-test="dropdown-item"
-            role="option"
+            role="menuitem"
```
LuisMartinez15, line 50: I guess that if we end up changing the top button `role` to `listbox` this should be reverted to `option`

```suggestion
            role="option"
```

LuisMartinez15 (changes requested):
good job!

packages/x-components/src/components/base-dropdown.vue
```diff
@@ -31,10 +32,11 @@
         @keydown.end="highlightLastItem"
         @keydown.esc="close"
         @keydown.home="highlightFirstItem"
-        :aria-expanded="isOpen.toString()"
         class="x-dropdown__items-list"
-        role="listbox"
+        role="menu"
```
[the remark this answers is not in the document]
mavmaf: As agreed via Slack, we'll go for `combobox`+ `listbox`

packages/x-components/src/components/base-dropdown.vue
```diff
@@ -31,10 +32,11 @@
...
         tabIndex="0"
+        :aria-activedescendant="value"
+        :aria-label="value"
```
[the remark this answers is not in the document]
mavmaf: let me know how it looks with the new patterns

packages/x-components/src/components/base-dropdown.vue
mavmaf: changed to `:aria-activedescendant="`dropdown-option-${value}`"`

packages/x-components/src/views/accessibility-check.vue
mavmaf: sounds good

LuisMartinez15 (approved):
good job!

packages/x-components/src/components/base-dropdown.vue
```diff
@@ -31,22 +38,23 @@
...
             @click="emitSelectedItemChanged(item)"
+            id="`dropdown-option-${value}`"
             :aria-selected="(index === highlightedItemIndex).toString()"
             :aria-current="(item === value).toString()"
             :class="itemsCSSClasses[index]"
             class="x-dropdown__item"
             data-test="dropdown-item"
             role="option"
-            tabindex="-1"
+            tabindex="0"
```
javieri-empathy, line 57: Because this is a button, it is automatically focusable in the document order. `tabindex` should not be needed.

packages/x-components/src/components/base-dropdown.vue
```diff
@@ -5,11 +5,18 @@
     @keydown.up.prevent="highlightPreviousItem"
     :class="dropdownCSSClasses"
     class="x-dropdown"
+    role="combobox"
+    aria-haspopup="listbox"
+    tabIndex="0"
+    :aria-expanded="isOpen"
+    :aria-label="value"
```
javieri-empathy, line 12: The aria label shouldn't be the selected value. It should be a description of what the field does. Moreover like I said below, the value can be an object. So adding this you are just making the component more inaccessible.

javieri-empathy (changes requested):

packages/x-components/src/components/base-dropdown.vue
mavmaf: Yep, my intention was to use it as fallback since there's no a common name descriptive enough I can come with both for the `base-column-picker-dropdown` and `sort-dropdown`. Any thoughts? I wouldn't trust on devs to remember to add it when creating their views ^^'

javieri-empathy: I wouldn't force it here. In the end this is more like an "abstract" component, we don't have that info at this level. We can add it for the "concrete" implementation of this component (the column picker and the sort) like you did with the switch.

javieri-empathy (approved):

LuisMartinez15 (approved):
good job!

support-empathy:
 [Check out EPDocs preview](https://docs-dev.empathy.co/preview/x-846/index.html)

javieri-empathy (approved):

PR #847, opened 2022-11-14
feat: history-queries-switch wai aria adjustments

mavmaf:
## Motivation and context
Ensure the component fulfil the following requirements:

**Keyboard Interaction**

- Space: When focus is on the switch, changes the state of the switch.
- Enter (Optional): When focus is on the switch, changes the state of the switch.

**WAI-ARIA Roles, States, and Properties**

- The switch has role switch.
- The switch has an accessible label provided by one of the following:

     1. Visible text content contained within the element with role switch.
     2. A visible label referenced by the value of aria-labelledby set on the element with role switch.
     3. aria-label set on the element with role switch.

- When on, the switch element has state aria-checked set to true.
- When off, the switch element has state aria-checked set to false.
- If the presentation includes additional descriptive static text relevant to a switch or switch group, the switch or switch group has the property aria-describedby set to the ID of the element containing the description.

## Type of change

- [x] `Main`
- [ ] Other. Specify:

## How has this been tested?

## Checklist:

### DIFF
--- a/packages/x-components/src/x-modules/history-queries/components/history-queries-switch.vue
+++ b/packages/x-components/src/x-modules/history-queries/components/history-queries-switch.vue
@@ -1,5 +1,5 @@
 <template>
-  <BaseSwitch @change="toggle" :value="isEnabled" />
+  <BaseSwitch @change="toggle" :value="isEnabled" aria-label="Queries' history" />
 </template>
 
 <script lang="ts">
--- a/packages/x-components/src/views/accessibility-check.vue
+++ b/packages/x-components/src/views/accessibility-check.vue
@@ -1,118 +1,55 @@
 <template>
-  <Scroll id="accessibility-scroll">
-    <section class="x-list x-padding--07 x-list--gap-05">
-      <div>
-        <h1>BaseIdTogglePanelButton</h1>
-        <BaseIdTogglePanelButton class="x-button x-button--ghost" :panelId="panelId">
-          Panel
-        </BaseIdTogglePanelButton>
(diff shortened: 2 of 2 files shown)


annacv (approved):

packages/x-components/src/x-modules/history-queries/components/history-queries-switch.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <BaseSwitch @change="toggle" :value="isEnabled" />
+  <BaseSwitch @change="toggle" :value="isEnabled" aria-label="Queries' history" />
```
javieri-empathy, line 2: Is this aria-label necessary? In our base use case we have a label for this switch that tells more about the action.
image

As I'm reading [here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby#benefits_and_drawbacks), `aria-labelledby` has highest precedence, so we might be able to keep this `aria-label` by default and add another `aria-labelledby` in the archetype. Is that correct?

mavmaf: Hey! As I see it now in the archetype, using a screen reader (I use Screen Reader Chrome extension instead of VoiceOver for sanity purposes xD) will only say "button ticked / button not ticked".
As you said, `aria-labelledby` has highest precedence, but in fact even a second `aria-label `will replace the message on this first one. So I think we're save using this label as 'fallback'.

javieri-empathy (approved):

PR #854, opened 2022-11-17
chore: create empty `x-bus` package

diegopf:
EX-7277

### DIFF
--- a/packages/x-bus/package.json
+++ b/packages/x-bus/package.json
@@ -1,12 +1,13 @@
 {
   "name": "x-bus",
-  "version": "1.0.0-alpha.00",
+  "version": "0.1.0-alpha.0",
   "description": "Event bus to help with events orchestration",
   "author": "Empathy Systems Corporation S.L.",
   "homepage": "https://github.com/empathyco/x/tree/main/packages/x-bus#readme",
   "license": "Apache-2.0",
   "main": "dist/cjs/index.js",
   "keywords": [
+    "bus",
     "events",
     "orchestration"
   ],
--- a/packages/x-bus/package.json
+++ b/packages/x-bus/package.json
@@ -1,0 +1,50 @@
+{
+  "name": "x-bus",
+  "version": "1.0.0-alpha.00",
+  "description": "Event bus to help with events orchestration",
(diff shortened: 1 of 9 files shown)


packages/x-bus/tsconfig.esm.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "module": "esnext",
+    "outDir": "./dist/esm",
+    "declaration": false,
+    "declarationDir": null,
+    "noEmit": false
```
LuisMartinez15, line 8: I think this should inherit the value of the extended parent `tsconfig.json`, as we don't want to emit declaration files twice (here and in the `cjs` build).

```suggestion
```

LuisMartinez15 (changes requested):
good job!

packages/x-bus/tsconfig.esm.json
diegopf: The `noEmit` is not only related to types but to [all compiler output files](https://www.typescriptlang.org/tsconfig#noEmit) so if we enable the `noEmit` flag we will not get the esm build 😅

packages/x-bus/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "build:cjs": "tsc --project tsconfig.cjs.json",
+    "build:esm": "tsc --project tsconfig.esm.json",
+    "postbuild": "npm pack",
+    "test": "jest",
+    "prepublishOnly": "npm run build"
+  },
+  "bugs": {
+    "url": "https://github.com/empathyco/x/issues"
+  },
+  "devDependencies": {
```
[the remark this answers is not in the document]
diegopf: Good catch! I think currently is working because `rimraf` is listed as dep in other packages but we shouldn't rely on that. I've added it to the devDeps

packages/x-bus/tsconfig.esm.json
LuisMartinez15: you're right, I always mix it with `types` emission 🤦🏿

PR #861, opened 2022-11-18
fix: number filter schema to map the right values

tiborux:
EX-7442

### DIFF
--- a/packages/x-adapter-platform/src/schemas/models/filters/__tests__/number-filter.spec.ts
+++ b/packages/x-adapter-platform/src/schemas/models/filters/__tests__/number-filter.spec.ts
@@ -29,7 +29,7 @@ describe('tests', () => {
     expect(filter.range).toEqual({ min: null, max: 0 });
   });
 
-  it('should map to null the max value even if the max value is 0', () => {
+  it('should map to null the max value even if the min value is 0', () => {
     const filter = mapper(createFilter('0-*'), {});
     expect(filter.range).toEqual({ min: 0, max: null });
   });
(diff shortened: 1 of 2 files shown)


packages/x-adapter-platform/src/schemas/models/filters/number-filter.schema.ts
```diff
@@ -10,7 +10,13 @@ export const numberFilterSchema = createMutableSchema<Schema<PlatformFilter, Num
   selected: () => false,
   modelName: () => 'NumberRangeFilter',
   range: {
-    min: ({ value }) => Number(value.split('-')[0]),
-    max: ({ value }) => Number(value.split('-')[1])
+    min: ({ value }) => {
+      const min = Number(value.split('-')[0]);
```
javieri-empathy, line 14: Where is the test for this?

javieri-empathy (changes requested):

packages/x-adapter-platform/src/schemas/models/filters/number-filter.schema.ts
tiborux: ufff dura esa

packages/x-adapter-platform/src/schemas/models/filters/__tests__/number-filter.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+    const filter = mapper(createFilter('10-*'), {});
+    expect(filter.range).toEqual({ min: 10, max: null });
+  });
+
+  it('should map to null the min value even if the max value is 0', () => {
+    const filter = mapper(createFilter('*-0'), {});
+    expect(filter.range).toEqual({ min: null, max: 0 });
+  });
+
+  it('should map to null the max value even if the max value is 0', () => {
```
annacv, line 32:
```suggestion
  it('should map to null the max value even if the min value is 0', () => {
```

javieri-empathy (approved):

annacv (approved):

PR #862, opened 2022-11-21
fix: make appear attribute configurable on animations that have it set to true by default

ajperezbau:
This change is needed to be able to change the animation of the image on hover without applying the new transition before the image is loaded.

Related to [this](https://github.com/empathyco/x/pull/863) PR.

EX-5970

### DIFF
--- a/packages/x-components/src/components/animations/collapse-height.vue
+++ b/packages/x-components/src/components/animations/collapse-height.vue
@@ -4,9 +4,9 @@
     @enter="expand"
     @after-enter="cleanUpAnimationStyles"
     @leave="collapse"
-    appear
     name="x-collapse-height-"
     v-bind="$attrs"
+    :appear="appear"
   >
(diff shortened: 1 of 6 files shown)


alvarodE (approved):

packages/x-components/src/components/animations/collapse-height.vue
```diff
@@ -29,6 +29,14 @@
     inheritAttrs: false
   })
   export default class CollapseHeight extends Vue {
+    /**
+     * Indicates if the transition must be applied on the initial render of the node.
+     */
+    @Prop({
+      type: Boolean,
```
mnavarroespinosa, line 36: Do we need to specify the type here? I think that it looks redundant as it is already typed in line 39. What do you think? @ajperezbau

mnavarroespinosa (approved):

PR #863, opened 2022-11-21
feat: show next result image on hover

ajperezbau:
To test this, the `x-components/src/views/home/result.vue` component can be updated to something like [this](https://gist.github.com/ajperezbau/3a28d62a40e687a7dc8e2f1b556d846e).

Keep in mind that the animation will not be applied correctly the first time, because the `appear` attribute of the animations is not configurable right now.

[In this PR](https://github.com/empathyco/x/pull/862) the animations are updated making it configurable.

In the meantime, the `appear` attribute of the animations used to test this can be set to false.

Also, in this PR I change the `picture` element of the component for a `div`. This is needed because `Firefox` and `Safari` browsers fire the `load` event of the images twice if they are inside a `picture` element. 

Even if we don't consider this problem, I am not sure that a `picture` element should be used there, because the purpose of that element is to provide multiple sources to the same image, and let the browser decide which one to use depending on the format, the viewport, and so on.

BREAKING CHANGE: The `base-result-image` `animation` prop is now named `loadAnimation`.

EX-5970


packages/x-components/src/components/result/base-result-image.vue
```diff
@@ -61,19 +71,51 @@
      */
     @Prop({ default: () => NoElement })
     public animation!: string | typeof Vue;
```
alvarodE, line 73: I imagine that it would be a breaking change and we prefer to avoid that issue but now that we have three different animation variables (`animation`, `hoverAnimation`, `imageAnimation`) I suggest a refactor on namings, something like:
1. `loadAnimation`, the prop to specify how to animate between placeholder & result image & fallback image,
2. `hoverAnimation`, the prop to specify how to change between images when hovering the result.
3. `animation` or `currentAnimation`, the getter that calculates the specific animation to apply at each moment.

packages/x-components/src/components/result/base-result-image.vue
```diff
@@ -61,19 +71,51 @@
...
+
+    /**
+     * Indicates if the next valid image should be displayed on hover.
+     *
+     * @public
+     */
+    @Prop({ type: Boolean, default: () => false })
```
alvarodE, line 80: We can simplify a bit this prop's definition:
```suggestion
    @Prop({ default: false })
```

CachedaCodes (changes requested):

packages/x-components/src/components/result/base-result-image.vue
ajperezbau: I would leave it as:

```
@Prop({ type: Boolean, default: false })
```

The `Boolean` type is needed there because of a bug in `vue-property-decorator`.

If you don't set it, when the component is used, and you set a boolean like `<component booleanProp />`, the value of `booleanProp` will be an empty string, instead of `true`.

packages/x-components/src/components/result/base-result-image.vue
```diff
@@ -1,32 +1,42 @@
 <template>
-  <picture ref="image" class="x-picture x-result-picture" data-test="result-picture">
+  <!-- This is a div because using a picture causes the onload event of the image to fire twice. -->
```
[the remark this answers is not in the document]
ajperezbau: Yep, I think that this comment is useful, just in case anyone wonders why it is not a `picture` anymore.

CachedaCodes (approved):

packages/x-components/src/components/result/base-result-image.vue
```diff
@@ -171,4 +240,12 @@ displayed while the real one is loaded.
   </template>
 </BaseResultImage>
 ```
+
+### Showing the next image on hover
+
+If a result has multiple images, it can show the next one on hover.
```
alvarodE, line 246: Does it make sense to include here also how to set an animation? So we document the two new props in the same example 🤔

packages/x-components/src/components/result/base-result-image.vue
```diff
@@ -53,27 +62,60 @@
...
+
+    /**
+     * Copy of the images of the result.
+     *
+     * It is used as a queue of images to load, once an image loads/fails to load, it is removed
+     * from this array.
+     *
+     * @internal
+     */
+    protected images: string[] = this.result.images ?? [];
```
alvarodE, line 91: As this array changes over time and in the end stores the images pending to be loaded, should we name it `pendingImages` or something similar?

packages/x-components/src/components/result/base-result-image.vue
```diff
@@ -98,8 +153,21 @@
      * @internal
      */
     protected get imageSrc(): string {
-      const image = this.result.images?.find(image => !this.failedImages.includes(image));
-      return image ?? '';
+      return this.loadedImages[
+        !this.showNextImageOnHover || !this.isHovering ? 0 : this.loadedImages.length - 1
```
alvarodE, line 157: I don't see the point on using `this.loadedImages.length - 1` and not just `1`, but probably I'm missing something I think.

packages/x-components/src/components/result/base-result-image.vue
```diff
@@ -98,8 +153,21 @@
...
+
+    /**
+     * Indicates if the loader should try to load the next image.
+     *
+     * @returns True if it should try to load the next image.
+     *
+     * @internal
+     */
+    protected get shouldLoadNextImage(): boolean {
+      const nextImgIndex = this.showNextImageOnHover && this.userHasHoveredImage ? 2 : 1;
```
alvarodE, line 169: If talking about indexes these number should be `1` and `0`, ins't it?

packages/x-components/src/components/result/base-result-image.vue
ajperezbau: Yep, it is `this.loadedImages.length - 1` for two reasons:
- The second image loads when the user hovers the picture the first time, so accessing the index 1 before the second image is loaded would cause an undefined index.
- The second image can fail to load. In that case, the first image should remain visible.

packages/x-components/src/components/result/base-result-image.vue
ajperezbau: Nope, the name of the variable is wrong.
They are not indexes, they are the number of images to load.
I will change that name 👍

packages/x-components/src/components/result/base-result-image.vue
ajperezbau: Sure!

alvarodE (approved):

alvarodE:
Nice job!

CachedaCodes (approved):

PR #865, opened 2022-11-22
feat: Hide history queries with no results

alvarodE:
[EX-7447](https://searchbroker.atlassian.net/browse/EX-7447)

### DIFF
--- a/packages/x-components/src/x-modules/history-queries/store/__tests__/actions.spec.ts
+++ b/packages/x-components/src/x-modules/history-queries/store/__tests__/actions.spec.ts
@@ -1,4 +1,4 @@
-import { HistoryQuery } from '@empathyco/x-types';
+import { HistoryQuery, Result } from '@empathyco/x-types';
 import { DeepPartial } from '@empathyco/x-utils';
 import Vue from 'vue';
 import Vuex, { Store } from 'vuex';
@@ -267,6 +267,55 @@ describe('testing history queries module actions', () => {
 
       await store.dispatch('toggleHistoryQueries', true);
       expect(localStorageService.getItem<boolean>(HISTORY_QUERIES_ENABLED_KEY)).toBe(true);
+    });
+  });
+
+  describe('updateHistoryQueriesWithSearchResponse', () => {
+    const results: Result[] = [],
+      totalResults = 50;
+    let gato: HistoryQuery, perro: HistoryQuery;
+    beforeEach(() => {
+      [gato, perro] = createHistoryQueries('gato', 'perro');
+      resetStateWith({ historyQueries: [gato, perro] });
+    });
+
+    it('updates a history query if its response is a success', async () => {
+      await store.dispatch('updateHistoryQueriesWithSearchResponse', {
+        request: {
+          query: 'gato',
+          page: 1
+        },
+        status: 'success',
+        results,
+        totalResults
+      });
+      expectHistoryQueriesToEqual([{ ...gato, totalResults }, perro]);
+    });
+
+    it('does not update a history query if its response is an error', async () => {
+      await store.dispatch('updateHistoryQueriesWithSearchResponse', {
+        request: {
+          query: 'gato',
+          page: 1
+        },
+        status: 'error',
+        results,
+        totalResults
+      });
+      expectHistoryQueriesToEqual([gato, perro]);
+    });
+
+    it('does not update a history query if it is not stored', async () => {
+      await store.dispatch('updateHistoryQueriesWithSearchResponse', {
+        request: {
+          query: 'pato',
+          page: 1
+        },
+        status: 'success',
+        results,
+        totalResults
+      });
+      expectHistoryQueriesToEqual([gato, perro]);
     });
   });
 });
--- a/packages/x-components/src/x-modules/history-queries/store/__tests__/actions.spec.ts
+++ b/packages/x-components/src/x-modules/history-queries/store/__tests__/actions.spec.ts
@@ -324,7 +324,7 @@ describe('testing history queries module actions', () => {
 
     it('updates a history query if it was made on a previous session', async () => {
       gato.totalResults = 50;
-      gato.timestamp -= store.state.sessionTimeStampInMs;
+      gato.timestamp = store.state.sessionTimeStampInMs;
       resetStateWith({ historyQueries: [gato, perro] });
       await store.dispatch('updateHistoryQueriesWithSearchResponse', {
         request: {
--- a/packages/x-components/src/x-modules/history-queries/store/__tests__/actions.spec.ts
+++ b/packages/x-components/src/x-modules/history-queries/store/__tests__/actions.spec.ts
@@ -277,6 +277,7 @@ describe('testing history queries module actions', () => {
     ];
     const totalResults = results.length;
     let gato: HistoryQuery, perro: HistoryQuery;
+
     beforeEach(() => {
       [gato, perro] = createHistoryQueries('gato', 'perro');
       resetStateWith({ historyQueries: [gato, perro] });
@@ -323,7 +324,7 @@ describe('testing history queries module actions', () => {
 
     it('updates a history query if it was made on a previous session', async () => {
       gato.totalResults = 50;
-      gato.timestamp -= 60 * 60 * 1000;
+      gato.timestamp -= store.state.sessionTimeStampInMs;
       resetStateWith({ historyQueries: [gato, perro] });
       await store.dispatch('updateHistoryQueriesWithSearchResponse', {
         request: {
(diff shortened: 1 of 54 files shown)


packages/x-components/src/x-modules/history-queries/store/actions/update-history-queries-with-search-response.action.ts
```diff
@@ -0,0 +1,29 @@
...
+ * @param searchResponse - The search response to update history queries with.
+ *
+ * @returns A `void` promise that resolves when the history query finishes updating.
+ *
+ * @public
+ */
+// eslint-disable-next-line max-len
+export const updateHistoryQueriesWithSearchResponse: HistoryQueriesXStoreModule['actions']['updateHistoryQueriesWithSearchResponse'] =
+  ({ state, dispatch }, searchResponse) => {
+    if (searchResponse.status === 'success') {
```
alvarodE, line 20: The main problem with this approach is here, and it is about when to update the total results count of a history query.

Right now every successfuly response updates it, but that means it will be updated when sorting, loading a new page of the same query, and even filtering which decreases the count and that is irrelevant for the history queries.

mnavarroespinosa (changes requested):
Good job!

packages/x-components/tests/e2e/history-queries.feature
```diff
@@ -94,3 +94,18 @@ Feature: History queries component
...
+  Scenario Outline: 7. Rendered history queries list is empty if no results are provided
+    Given following config: hide if equals query <hideIfEqualsQuery>, debounce <debounceInMs>, requested items <maxItemsToStore>, rendered <maxItemsToRender>, instant search <instant>
+    And   a results API with no results
+    And   start button is clicked
+    And   no history queries are displayed
+    When  "<query>" is searched
+    Then  the searched query is displayed in the search-box
+    When  clear search button is pressed
+    Then  no history queries are displayed
+    And   clear history queries button is enabled
```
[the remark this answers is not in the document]
alvarodE: I agree with you and I thought about it but the problem I see is that some customers don't display My history, so users wouldn't have a way to clear their history if we disable the button and only queries with no results have been performed.

I'll create a task in the backlog to document this issue: [EX-7475](https://searchbroker.atlassian.net/browse/EX-7475)

mnavarroespinosa (approved):
Looks good to me!

ajperezbau (changes requested):
Good job!

packages/x-components/src/x-modules/history-queries/store/getters/history-queries-with-results.getter.ts
```diff
@@ -0,0 +1,18 @@
...
+ * @param getters - Current {@link https://vuex.vuejs.org/guide/getters.html | getters} of the
+ * history queries module.
+ *
+ * @returns The history queries that have results at the moment they were requested.
+ *
+ * @public
+ */
+// eslint-disable-next-line max-len
+export const historyQueriesWithResults: HistoryQueriesXStoreModule['getters']['historyQueriesWithResults'] =
+  (state, { historyQueries }) => historyQueries.filter(({ totalResults }) => totalResults !== 0);
```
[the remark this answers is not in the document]
alvarodE: Yes but I put it like this intentionally, because meanwhile the request has not succeeded the value will be `undefined` and in that case I think the safest approach is to not remove it from the view. What do you think? Does it make sense or is it better the other way around?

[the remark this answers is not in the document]
ajperezbau: I would say that it is safer the other way around.
If there is some error with the request, the history query will be rendered anyway, so it will make the step of not updating the query if the request fails useless.
And even if you add something to handle the request error and hide the history query, It seems a bit odd to remove it just a bit after it was rendered.

packages/x-components/src/x-modules/history-queries/store/__tests__/actions.spec.ts
```diff
@@ -269,4 +269,88 @@ describe('testing history queries module actions', () => {
...
+        status: 'success',
+        results,
+        totalResults
+      });
+      expectHistoryQueriesToEqual([gato, perro]);
+    });
+
+    it('updates a history query if it was made on a previous session', async () => {
+      gato.totalResults = 50;
+      gato.timestamp -= store.state.sessionTimeStampInMs;
```
ajperezbau, line 327: I would set it to `store.state.sessionTimeStampinMs - 1` because in the tests `sessionTimeStampinMs` seems to be 0.

```suggestion
      gato.timestamp = store.state.sessionTimeStampInMs - 1;
```

The test is passing right now, but I think it should be failing, I will let you another comment in the action.

packages/x-components/src/x-modules/history-queries/store/actions/update-history-queries-with-search-response.action.ts
```diff
@@ -0,0 +1,46 @@
...
+      const indexOfHistoryQuery = state.historyQueries.findIndex(
+        ({ query }) => query === searchResponse.request.query
+      );
+      if (indexOfHistoryQuery >= 0) {
+        const historyQuery = state.historyQueries[indexOfHistoryQuery];
+        const isCurrentSessionHistoryQuery = historyQuery.timestamp > state.sessionTimeStampInMs;
```
ajperezbau, line 31: Now that I think about this, shouldn't this be an equal or greater condition?

```suggestion
        const isCurrentSessionHistoryQuery = historyQuery.timestamp >= state.sessionTimeStampInMs;
```

The `sessionTimeStampInMs` contains the timestamp when the current session started, so If the query timestamp is equal it was done in the current session.

alvarodE: Mmm I took the `sessionHistoryQueries` as a reference, there the condition is `>`, don't you think here it should be the same? @ajperezbau

ajperezbau: You are right!
It should be the same, so it is ok as it is 👌

packages/x-components/src/x-modules/history-queries/store/__tests__/actions.spec.ts
ajperezbau: Ok, since the `sessionTimeStampInMs` is not considered to be in the current session, what do you think about setting the query timestamp to the same value?
So it is clear that this behavior is intentional.

```suggestion
      gato.timestamp = store.state.sessionTimeStampInMs;
```

packages/x-components/src/x-modules/history-queries/store/__tests__/actions.spec.ts
```diff
@@ -269,4 +269,88 @@ describe('testing history queries module actions', () => {
...
+        },
+        status: 'success',
+        results,
+        totalResults
+      });
+      expectHistoryQueriesToEqual([{ ...gato, totalResults }, perro]);
+    });
+
+    it('updates a history query if the new totalResults is higher', async () => {
+      gato.totalResults = 1;
```
mnavarroespinosa, line 342: The test pass even if the new `totalResults` is lower, I guess it is due to both the HQ and session timestamps being 0 causing `isCurrentSessionHistoryQuery` in the action to be false. 

```suggestion
      gato.totalResults = 1;
      gato.timestamp = store.state.sessionTimeStampInMs + 100; // Make sure HQ is in current session
```

Could you also add a test to check that the HQ is not updated when the new `totalResults` is lower?

mnavarroespinosa (changes requested):

packages/x-components/tests/e2e/history-queries/history-queries.spec.ts
```diff
@@ -46,6 +46,7 @@ And('a {string} of queries already searched', (list: string) => {
 
 When('history query number {int} is clicked', (historyQueryItem: number) => {
   cy.getByDataTest('history-query').eq(historyQueryItem).click().invoke('text').as('searchedQuery');
+  cy.waitForResultsToRender();
```
ajperezbau, line 49: Here I would say the same that I commented about the `query-suggestions` test.
I would add the step to wait for the results after the `    And   history query number <historyQueryItem> is clicked` of the first scenario:

```suggestion
```
```
```

ajperezbau (changes requested):
Keep in mind that I will create a task to create a step similar to the `related results are displayed`, but removing the extra logic, only waiting for the results to load.
That is because the step is being used in a lot of sites, but saving the results in the list and so on is only needed in a few that check if the results have changed later.

Good job! 👌

mnavarroespinosa (approved):
Good job!

ajperezbau (approved):
Nice! 👌

PR #869, opened 2022-11-22
fix: Avoid creating npmrc file on Github Actions

alvarodE:
The previous implementation of the Github Actions to release a version included creating a new .npmrc file. Now that this file already exists, there is no way to edit it without creating a change in the repo that Lerna will complain about.

The solution is to perform the same operation, which is configure the npm auth, without modifying the .npmrc file. `npm config set` changes the root .npmrc file, so it shouldn't be a problem.

https://docs.npmjs.com/cli/v8/commands/npm-config
https://stackoverflow.com/a/43012838

A different approach would be to ignore the .npmrc file on Lerna checking for pending changes, [here](https://github.com/empathyco/x/blob/main/lerna.json#L4):
Screenshot 2022-11-22 at 15 18 36

### DIFF
--- a/.github/actions/release-alpha/action.yml
+++ b/.github/actions/release-alpha/action.yml
@@ -15,8 +15,10 @@ runs:
         git config --local user.email "x@empathy.co"
         git config --local user.name "empathy/x"
       shell: bash
-    - name: Create .npmrc file with authorization token
-      run: echo //registry.npmjs.org/:_authToken=${NPM_TOKEN} > .npmrc
+    - name: Configure npm auth
+      run: |
+        npm config set registry="http://registry.npmjs.org/"
+        npm config set _authToken=${NPM_TOKEN}
       shell: bash
       env:
         NPM_TOKEN: ${{ inputs.npm_token }}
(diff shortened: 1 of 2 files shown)


.github/actions/release-alpha/action.yml
```diff
@@ -15,8 +15,10 @@ runs:
         git config --local user.email "x@empathy.co"
         git config --local user.name "empathy/x"
       shell: bash
-    - name: Create .npmrc file with authorization token
-      run: echo //registry.npmjs.org/:_authToken=${NPM_TOKEN} > .npmrc
+    - name: Configure npm auth
+      run: |
+        npm config set registry="http://registry.npmjs.org/"
```
diegopf, line 20:
```suggestion
        npm config set registry="//registry.npmjs.org/"
```

diegopf (changes requested):
I think we shouldn't add the protocol here, specially if it is http and not https

javieri-empathy (dismissed):

.github/actions/release-alpha/action.yml
alvarodE: It is needed to specify the registry, anyway I'll use https :sisi3:

diegopf (approved):

javieri-empathy (approved):

PR #870, opened 2022-11-23
Use `event.composedPath()[0]` instead of `event.target`

joseacabaneros:
Two different things were done here:
- Create `getTargetElement` utility function to retrieve the event target through `event.composedPath()[0]` instead of `event.target`. The reason is `event.target` in a Shadow DOM context, points to the Shadow DOM element, not the element that triggered the event. You can check real examples inside unit testing added. 
Components, which used `event.target` to check if a click were triggered outside component, were replaced for this utility function.
More info: https://tinytip.co/tips/js-event-composed-path/
- Overlapping `move` transition in `staggered-fade-and-slide` animation with modals due to `z-index: 1`. 
Screenshot 2022-11-22 at 5 35 10 PM


packages/x-components/src/components/animations/staggered-fade-and-slide.vue
```diff
@@ -48,22 +48,26 @@
...
-    }
+    &::v-deep .x-staggered-fade-and-slide {
+      &--enter-active,
+      &--leave-active {
+        transition: $transition-duration ease-out;
+        transition-property: opacity, transform;
+      }
+
+      &--move {
+        transition: transform $transition-duration ease-out;
```
joseacabaneros, line 62: Actually, we can maintain the `z-index: 1;` if the container has `z-index: 0;`. Whatever you prefer

packages/x-components/src/utils/html.ts
```diff
@@ -10,3 +10,21 @@
...
+ * @remarks In a shadow DOM context, this function only works if the Shadow DOM uses `open`
+ * encapsulation mode.
+ *
+ * @param event - Event which takes place in the DOM.
+ * @returns Target Element of the event.
+ *
+ * @public
+ */
+export function getTargetElement(event: Event): Element {
+  return (event.composedPath()[0] ?? event.target) as Element;
```
joseacabaneros, line 29: I kept `event.target` as fallback for retrocompatibility, but it wouldn't be needed because `event.composedPath()[0]` should always have the target element. WDTY? Can we get rid of it?

packages/x-components/src/utils/html.ts
```diff
@@ -10,3 +10,21 @@
 export function isElementEqualOrContained(a: Element, b: Element): boolean {
   return a === b || a.contains(b);
 }
+
+/**
+ * Returns the target element for a given event. The target element is got form `composedPath` Event
```
joseacabaneros, line 15: `is got from`

packages/x-components/src/utils/html.ts
javieri-empathy: AFAIK there is no possible way that the `composedPath` method returns an empty array, so I would get rid of the `event.target`.

packages/x-components/src/utils/__tests__/html.spec.ts
```diff
@@ -24,3 +24,78 @@ describe(`testing ${isElementEqualOrContained.name} utility method`, () => {
     expect(isElementEqualOrContained(a, b)).toBe(false);
   });
 });
+
+describe(`testing ${getTargetElement.name} utility method`, () => {
```
javieri-empathy, line 28: Because this is a browser behavior,  I would move this test to Cypress Component testing, rather than using `jest` and `jsdom`. Then we can add more browsers and test that this works properly in all of them.

javieri-empathy: I tried this in Cypress and it gives me much more confidence than `jsdom` 🤣.
```
import { getTargetElement } from '../../src/utils/html';

describe(`testing ${getTargetElement.name} utility method`, () => {
  beforeEach(() => {
    document.body.innerHTML = '';
  });

  function createBasicDOMTree(): [HTMLDivElement, HTMLButtonElement] {
    const container = document.createElement('div');
    const button = container.appendChild(document.createElement('button'));
    button.textContent = 'Click';

    return [container, button];
  }

  function createShadowDOMTree(mode: 'open' | 'closed'): [HTMLDivElement, HTMLButtonElement] {
    const [container, button] = createBasicDOMTree();

    const shadowContainer = document.createElement('div');
    const shadowRoot = shadowContainer.attachShadow({ mode });
    shadowRoot.appendChild(container);

    return [shadowContainer, button];
  }

  it('returns the element that triggered the event', () => {
    const [container, button] = createBasicDOMTree();
    document.body.appendChild(container);
    const spy = cy.spy(getTargetElement);
    container.addEventListener('click', spy);
    button.click();
    cy.wrap(spy).should('have.callCount', 1);
    cy.wrap(spy).should('have.returned', button);
  });

  it('returns the element that triggered the event in a opened shadow DOM context', () => {
    const [shadowContainer, button] = createShadowDOMTree('open');
    const spy = cy.spy(getTargetElement);
    shadowContainer.addEventListener('click', spy);
    document.body.appendChild(shadowContainer);
    button.click();
    cy.wrap(spy).should('have.callCount', 1);
    cy.wrap(spy).should('have.returned', button);
  });

  it("doesn't return the element that triggered the event in a closed shadow DOM context", () => {
    const [shadowContainer, button] = createShadowDOMTree('closed');
    const spy = cy.spy(getTargetElement);
    shadowContainer.addEventListener('click', spy);
    document.body.appendChild(shadowContainer);
    button.click();
    cy.wrap(spy).should('have.callCount', 1);
    cy.wrap(spy).should('have.returned', shadowContainer);
  });
});
```

packages/x-components/src/components/animations/staggered-fade-and-slide.vue
javieri-empathy: I buy the change like this.
Is this to remove the `1001` index that you added in motive?

javieri-empathy (changes requested):

packages/x-components/src/components/animations/staggered-fade-and-slide.vue
joseacabaneros: Nope, it is to live together with the `z-index` of the `base-id-modal` component. https://github.com/empathyco/x/blob/main/packages/x-components/src/components/modals/base-modal.vue#L277

packages/x-components/src/utils/html.ts
joseacabaneros: Removed then!

packages/x-components/src/components/base-dropdown.vue
```diff
@@ -78,6 +78,7 @@
   import { Identifiable } from '@empathyco/x-types';
   import { Component, Prop, Watch } from 'vue-property-decorator';
   import Vue from 'vue';
+  import { getTargetElement } from '../utils/index';
```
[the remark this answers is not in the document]
joseacabaneros: Changed

packages/x-components/src/utils/__tests__/html.spec.ts
joseacabaneros: I agree with you. Migrated unit test to Cypress

javieri-empathy (approved):

LuisMartinez15 (approved):
good job!

PR #871, opened 2022-11-24
feat(tabs-panel): add base tabs panel component

mnavarroespinosa:
EX-6142


packages/x-components/src/components/panels/base-tabs-panel.vue
```diff
@@ -0,0 +1,331 @@
+<template>
+  <section v-if="getTabs().length > 0" class="x-tabs-panel" data-test="base-tabs-panel">
+    <component
+      :is="animation"
```
alvarodE, line 4: I'm wondering wether we should apply an animation to the content instead of the tabs list. In the end I think the part which will change the most with users interactions will be that content, isn't it? Each time the user clicks a different tab.

packages/x-components/src/components/panels/base-tabs-panel.vue
```diff
@@ -0,0 +1,331 @@
...
+      <li v-for="tab in getTabs()" :key="tab">
+        <!--
+          @slot Slot that is be used for replacing the whole tab.
+            @binding {tab} string - The tab name.
+            @binding {isSelected} boolean - Indicates if the tab is selected.
+            @binding {selectTab} {() => void} selectTab - Function to select the tab.
+         -->
+        <slot
+          name="tab"
+          v-bind="{ tab, isSelected: tabIsSelected(tab), selectTab: () => selectTab(tab) }"
```
alvarodE, line 19: In the same way `isSelected` is a simplified name from `tabIsSelected` I'd say we can simplify `selectTab`:
```suggestion
          v-bind="{ tab, isSelected: tabIsSelected(tab), select: () => selectTab(tab) }"
```

packages/x-components/src/components/panels/base-tabs-panel.vue
```diff
@@ -0,0 +1,331 @@
...
+    protected selectedTab: string = this.initialTab;
+
+    /**
+     * Extracts the tab from the slots.
+     *
+     * @returns The list of tabs.
+     *
+     * @internal
+     */
+    protected getTabs(): string[] {
```
alvarodE, line 98: Could this be a getter?
```suggestion
    protected get tabs(): string[] {
```

packages/x-components/src/components/panels/base-tabs-panel.vue
```diff
@@ -0,0 +1,331 @@
...
+      return Object.keys(this.$scopedSlots).filter(tab => tab !== 'tab-content' && tab !== 'tab');
```
alvarodE, line 99: A different implementation that for me it seems easier to "expand" and add more slots in the future would be:
```suggestion
      return Object.keys(this.$scopedSlots).filter(slotName => !['tab', 'tab-content'].includes(slotName));
```

packages/x-components/src/components/panels/base-tabs-panel.vue
```diff
@@ -0,0 +1,331 @@
...
+         -->
+        <slot
+          name="tab"
+          v-bind="{ tab, isSelected: tabIsSelected(tab), selectTab: () => selectTab(tab) }"
+        >
+          <button
+            :key="tab"
+            @click="selectTab(tab)"
+            class="x-button x-tabs-panel__button"
+            :class="tabIsSelected(tab) ? activeTabClass : tabClass"
```
alvarodE, line 25: Should we apply the `tabClass` always, and the `activeTabClass` only when selected?
```suggestion
            :class="[tabClass, { [activeTabClass]: tabIsSelected(tab) }]"
```

packages/x-components/src/components/panels/base-tabs-panel.vue
```diff
@@ -0,0 +1,331 @@
...
+          </button>
+        </slot>
+      </li>
+    </component>
+
+    <div
+      v-if="selectedTab && $scopedSlots[selectedTab]"
+      :key="selectedTab"
+      :class="panelClass"
+      data-test="base-tabs-panel-content"
```
alvarodE, line 46: I find these twoo lines a little bit confusing as we are naming as "panel" this element when talking about classes, but as "content" when talking about tests. As the component itself is named "tabs panel", I'd go with "content" always.
```suggestion
      :class="contentClass"
      data-test="base-tabs-panel-content"
```

packages/x-components/src/components/panels/base-tabs-panel.vue
mnavarroespinosa: I was not completely sure about this, but I thought about adding the animation to the list (to animate how the list is presented, or new tabs could appear) and not to the tab itself as this last can be added anyway through the `tab` slot.  This way both animations can be added in one way or another.

packages/x-components/src/components/panels/base-tabs-panel.vue
mnavarroespinosa: Sure, good point!

packages/x-components/src/components/panels/base-tabs-panel.vue
```diff
@@ -0,0 +1,331 @@
...
+            @binding {tab} string - The tab name.
+            @binding {isSelected} boolean - Indicates if the tab is selected.
+            @binding {selectTab} {() => void} selectTab - Function to select the tab.
+         -->
+        <slot
+          name="tab"
+          v-bind="{ tab, isSelected: tabIsSelected(tab), selectTab: () => selectTab(tab) }"
+        >
+          <button
+            :key="tab"
```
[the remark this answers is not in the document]
mnavarroespinosa: You're totally right, it's not needed, it's collateral damage from refactoring 🤣

packages/x-components/src/components/panels/base-tabs-panel.vue
mnavarroespinosa: `$scopedSlots` is not reactive, so I can't use a computed property here.

packages/x-components/src/components/panels/base-tabs-panel.vue
mnavarroespinosa: It's a good question, my first approach was to always add `tabClass` but I think it can be a problem under some scenarios. 

For example, having the following values:
```
<BaseTabsPanel
  activeTabClass="x-button-auxiliary x-button-lg" <!-- Larger button with auxiliary style -->
  tabClass="x-button-ghost x-button-sm" <!-- Smaller button with ghost style -->
>
```
This would lead to the active tab having the following classes: `x-button-ghost x-button-sm x-button-auxiliary x-button-lg`, where some styles for the active tab can overlap with the tab styles, to overcome this we would need to maybe use the `!important` flag.

packages/x-components/src/components/panels/base-tabs-panel.vue
mnavarroespinosa: Sounds good to me, I'll change it, thanks!

packages/x-components/src/components/panels/base-tabs-panel.vue
ajperezbau: I think that it is better to apply only one of them to avoid having the situations that @mnavarroespinosa describes.

packages/x-components/src/components/panels/base-tabs-panel.vue
```diff
@@ -0,0 +1,332 @@
...
+    <component
+      :is="animation"
+      tag="ul"
+      class="x-list"
+      :class="tabsListClass"
+      data-test="base-tabs-panel-list"
+    >
+      <li v-for="tab in getTabs()" :key="tab">
+        <!--
+          @slot Slot that is be used for replacing the whole tab.
```
ajperezbau, line 12:
```suggestion
          @slot Slot used to replace the whole tab.
```

packages/x-components/src/components/panels/base-tabs-panel.vue
```diff
@@ -0,0 +1,332 @@
+<template>
```
ajperezbau, line 1: Just a question, should we add the `tablist` role to the list component, and the `tab` role to the `button`?
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tab_role

packages/x-components/src/components/panels/base-tabs-panel.vue
```diff
@@ -0,0 +1,332 @@
...
+      </li>
+    </component>
+
+    <div
+      v-if="selectedTab && $scopedSlots[selectedTab]"
+      :key="selectedTab"
+      :class="contentClass"
+      data-test="base-tabs-panel-content"
+    >
+      <slot :name="selectedTab" />
```
ajperezbau, line 47: Can you document this slot?

ajperezbau (changes requested):
Nice job!

packages/x-components/src/components/panels/__tests__/base-tabs-panel.spec.ts
```diff
@@ -0,0 +1,208 @@
...
+/**
+ * Function that returns a BaseTabsPanel wrapper. The animation prop is not
+ * tested because it is part of Vue 'component'.
+ *
+ * @param params - Wrapper options.
+ *
+ * @returns BaseTabsPanel vue-test-utils wrapper.
+ */
+function renderBaseTabsPanel({
+  scopedSlots,
```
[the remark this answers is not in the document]
mnavarroespinosa: Sure! I've added another prop for it.

packages/x-components/src/components/panels/base-tabs-panel.vue
mnavarroespinosa: Good point! I've added the `tablist`, `tab`, and also `tabpanel` roles.

packages/x-components/src/components/panels/base-tabs-panel.vue
mnavarroespinosa: Sure thing!

packages/x-components/src/components/panels/base-tabs-panel.vue
mnavarroespinosa: I've added a new prop to have an animation for the tab content and left another animation for the tabs list.

packages/x-components/src/components/panels/__tests__/base-tabs-panel.spec.ts
```diff
@@ -0,0 +1,218 @@
...
+
+    expect(getTabsButtons()).toHaveLength(1);
+    expect(getTabsButtons().at(0).text()).toBe('custom summer tab content');
+
+    // Select first tab
+    await clickNthTab(0);
+    expect(getTabsButtons().at(0).text()).toBe('custom summer tab content ✅');
+  });
+
+  it('changes the selected tab on click', async () => {
```
ajperezbau, line 146: Can you add to this test some assertions to check if the `aria-selected` attribute is being updated?

ajperezbau (changes requested):
Nice 👌

packages/x-components/src/components/panels/base-tabs-panel.vue
```diff
@@ -0,0 +1,385 @@
...
+      BaseTabsPanel
+    }
+  };
+</script>
+```
+
+#### Customizing the content with classes
+
+- The `activeTabClass` prop can be used to add classes to the active tab button.
+- The `contentClass` prop can be used to add classes to the panel.
```
alvarodE, line 226:
```suggestion
- The `contentClass` prop can be used to add classes to the content.
```

alvarodE (dismissed):

alvarodE:
Nice job!

packages/x-components/src/components/panels/__tests__/base-tabs-panel.spec.ts
mnavarroespinosa: Done!

alvarodE (dismissed):

ajperezbau (commented):
Good job!

packages/x-components/src/components/panels/base-tabs-panel.vue
```diff
@@ -63,14 +64,19 @@
...
   import { Component, Prop } from 'vue-property-decorator';
+  import { NoElement } from '../no-element';
   import { dynamicPropsMixin } from '../dynamic-props.mixin';
 
   /**
    * Base Tabs Panel.
    *
    * @public
    */
-  @Component
```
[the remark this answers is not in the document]
mnavarroespinosa: Great! Looks better this way

ajperezbau (approved):
Looks good to me!

packages/x-components/src/components/panels/base-tabs-panel.vue
```diff
@@ -0,0 +1,470 @@
...
+
+    /**
+     * Changes the current selected tab. If the tab is already selected
+     * and `allowTabDeselect` is `true`, the tab will be unselected.
+     *
+     * @param tab - The tab to be selected.
+     *
+     * @internal
+     */
+    protected selectTab(tab: string): void {
```
alvarodE, line 143: As now we can select and desect I'd say that something like `toggleTab` would be easier to understand, maybe?

alvarodE (approved):

packages/x-components/src/components/panels/base-tabs-panel.vue
mnavarroespinosa: It's a good point, I did not choose `toggleTab` as it is not always toggling the selection, it will only do it when `allowTabDeselect` is `true`.

PR #872, opened 2022-11-24
feat: add `Highlight` component

javieri-empathy:
Adds a `Highlight` component (both Vue and XDS). This component takes some text and highlights some parts.

image


packages/x-components/src/components/highlight.vue
```diff
@@ -0,0 +1,217 @@
...
+     * property will contain the whole text.
+     *
+     * @returns An object containing the different parts of the text.
+     * @internal
+     */
+    protected get matchParts(): HighlightMatch {
+      const matcherIndex = normalizeString(this.text).indexOf(normalizeString(this.highlight));
+      return matcherIndex !== -1 && this.highlight
+        ? this.splitAt(this.text, matcherIndex, matcherIndex + this.highlight.length)
+        : { start: this.text, match: '', end: '' };
```
mnavarroespinosa, line 112: The `matcherIndex` is based on the normalized strings of `text` and `highlight` but the `text` and `highlight` length passed to `splitAt` are not normalized so I guess it could return unexpected results in case the original variables contain spaces at start/end or diacritical marks causing the length to change.

packages/x-components/src/components/suggestions/__tests__/highlight.spec.ts
```diff
@@ -0,0 +1,162 @@
+import { mount, Wrapper } from '@vue/test-utils';
```
mnavarroespinosa, line 1: Could you move this file next to the `highlight.vue` component file? Now it is under `suggestions` folder.

packages/x-components/src/components/suggestions/__tests__/highlight.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+      return wrapper.find(getDataTestSelector('highlight-end'));
+    },
+    async setHighlight(highlight) {
+      return await wrapper.setProps({ highlight });
+    }
+  };
+}
+
+describe('testing Highlight component', () => {
+  it('highlights the rendered text when the match happens at the beginning', () => {
```
mnavarroespinosa, line 49: In relation to the comment I put in the computed `matchParts` of the Highlight component, could you add a test where the `text` contains spaces at the start or end?

packages/x-components/src/components/highlight.vue
javieri-empathy: Good catch. I think by adding a `.trim()` call we are good to go. Combining diacritical marks shouldn't make the text length to change.

tajespasarela (changes requested):

tajespasarela (approved):

mnavarroespinosa (approved):

PR #873, opened 2022-11-24
chore: Remove unused configs and docs

javieri-empathy:
This configurations now belong to the suggestion components. The recommendations docs where talking about suggestions  (copy & paste :troll:)

### DIFF
--- a/packages/x-components/src/x-modules/recommendations/components/recommendations.vue
+++ b/packages/x-components/src/x-modules/recommendations/components/recommendations.vue
@@ -161,43 +161,6 @@ recommendations.
 </Recommendations>
 ```
 
-### Play with props
-
-In this example, the suggestions has been limited to render a maximum of 3 items.
-
-_Type “puzzle” or another toy in the input field to try it out!_
-
-```vue
-<template>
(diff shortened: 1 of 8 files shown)


tajespasarela (dismissed):

packages/x-components/src/x-modules/recommendations/components/recommendations.vue
```diff
@@ -161,43 +161,6 @@ recommendations.
 </Recommendations>
 ```
 
-### Play with props
```
mnavarroespinosa, line 164: I think we could keep this example as `maxItemsToRender` is actually a prop of this component that is even being tested in the tests. 

Apart from changing `BaseSuggestions` to `Recommendations`, the prop `maxItemsToRender` needs to be corrected as it is misspelled as `maxItemToRender` in the example template. It's misspelled also in `PopularSearches`, `BaseSuggestions`, and `NextQueries` files.

mnavarroespinosa (changes requested):

tajespasarela (dismissed):

tajespasarela (approved):

PR #884, opened 2022-11-25
ci: add `combine-prs` workflow

diegopf:
EX-7430

Adding this workflow should help us to review and merge dependabot PRs by only having to review one instead of a few. It will be possible to run this workflow on demand and it will merge all PRs that match a configurable prefix and conditions, see screenshot.

image

Currently I'm adding the workflow from [combine prs repo](https://github.com/hrvey/combine-prs-workflow#get-started) as it is because I think it fits our needs out of the box but we could enhance it to adapt it to our use case if needed.

### DIFF
--- a/.github/workflows/combine-prs.yml
+++ b/.github/workflows/combine-prs.yml
@@ -1,0 +1,151 @@
+name: 'Combine PRs'
+
+# Controls when the action will run - in this case triggered manually
+on:
+  workflow_dispatch:
+    inputs:
+      branchPrefix:
+        description: 'Branch prefix to find combinable PRs based on'
+        required: true
+        default: 'dependabot'
+      mustBeGreen:
+        description: 'Only combine PRs that are green (status is success)'
+        required: true
+        default: true
+      combineBranchName:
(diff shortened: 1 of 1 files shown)


.github/workflows/combine-prs.yml
```diff
@@ -0,0 +1,151 @@
+name: 'Combine PRs'
```
alvarodE, line 1: Would it make sense to document a little bit what this workflow does? In case any developer arrives here with doubts about the process and can avoid reading all the code to understand it. 

At least topics like which is the result of executing this workflow and which conditions are applied (only dependabot, the label to skip it, etc.) I'd find it useful, but I see no other workflows are documented so let's see your opinion :)

javieri-empathy (dismissed):
OK, I would refactor this code and extract it to an action 😅, but as we discussed, I agree in trying it first before spending more effort here.

javieri-empathy (approved):

LuisMartinez15 (approved):
good job 👏🏿

PR #885, opened 2022-11-25
feat: script to import SVGs into Vue components for the XDS

CachedaCodes:
[EX-7384](https://searchbroker.atlassian.net/browse/EX-7384)

## Motivation and context
We needed an easy way to import SVGs from Figma to a Vue component we could use with the XDS. This script should with that.

## Type of change

- [X] `Main`
- [ ] Other. Specify:

## How has this been tested?
Create a folder and put some SVGs in it, the run the script as indicated in the README and see the Vue component it generates.

### DIFF
--- a/packages/x-svg-converter/package.json
+++ b/packages/x-svg-converter/package.json
@@ -27,7 +27,7 @@
   "scripts": {
     "build": "tsc",
     "postbuild": "npm pack",
-    "test": "jest -i"
+    "test": ""
   },
   "devDependencies": {
     "@types/jest": "~27.0.3",
--- a/packages/x-svg-converter/package.json
+++ b/packages/x-svg-converter/package.json
@@ -25,6 +25,7 @@
     "node": "16"
   },
   "scripts": {
+    "prebuild": "rimraf dist ./*.tgz",
     "build": "tsc",
     "postbuild": "npm pack",
     "test": "jest -i"
--- a/packages/x-svg-converter/package.json
+++ b/packages/x-svg-converter/package.json
@@ -27,7 +27,7 @@
   "scripts": {
     "build": "tsc",
     "postbuild": "npm pack",
-    "test": ""
+    "test": "jest -i"
   },
   "devDependencies": {
     "@types/jest": "~27.0.3",
--- a/packages/x-svg-converter/package.json
+++ b/packages/x-svg-converter/package.json
@@ -14,8 +14,7 @@
     "dist"
   ],
   "bin": {
-    "json-csv": "dist/json-to-csv.js",
-    "csv-json": "dist/csv-to-json.js"
+    "svg-vue": "dist/svg-to-vue.js"
   },
   "repository": {
     "type": "git",
--- a/packages/x-svg-converter/package.json
+++ b/packages/x-svg-converter/package.json
@@ -26,7 +26,7 @@
   },
   "scripts": {
     "prebuild": "rimraf dist ./*.tgz",
-    "build": "tsc",
+    "build": "tsc --project tsconfig.build.json",
     "postbuild": "npm pack",
     "test": "jest -i"
   },
--- a/packages/x-svg-converter/package.json
+++ b/packages/x-svg-converter/package.json
@@ -1,0 +1,40 @@
+{
+  "name": "@empathyco/x-svg-converter",
+  "version": "0.1.0",
+  "private": true,
+  "description": "A project which allows the user to convert SVGs into Vue components formatted to be used with the x-tailwindcss design system.",
+  "author": "Empathy Systems Corporation S.L.",
+  "license": "Apache-2.0",
+  "homepage": "https://github.com/empathyco/x/tree/main/packages/x-svg-converter#readme",
+  "keywords": [
+    "svg",
+    "converter"
+  ],
+  "files": [
+    "dist"
+  ],
+  "bin": {
+    "json-csv": "dist/json-to-csv.js",
+    "csv-json": "dist/csv-to-json.js"
+  },
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/empathyco/x.git",
+    "directory": "packages/x-svg-converter"
+  },
+  "engines": {
+    "node": "16"
+  },
+  "scripts": {
+    "build": "tsc",
+    "postbuild": "npm pack",
+    "test": "jest -i"
+  },
+  "devDependencies": {
+    "@types/jest": "~27.0.3",
+    "@types/node": "~16.4.12",
+    "jest": "~27.3.1",
+    "ts-jest": "~27.0.7",
+    "typescript": "~4.6.2"
+  }
+}
--- a/packages/x-svg-converter/src/__tests__/svg-to-vue.spec.ts
+++ b/packages/x-svg-converter/src/__tests__/svg-to-vue.spec.ts
@@ -7,20 +7,10 @@ describe('test SVG to Vue script', () => {
   const sourcePath = './src/__tests__/svgs';
 
   afterEach(() => {
-    process.argv = [];
-    jest.restoreAllMocks();
+    jest.resetAllMocks();
 
-    const absoluteSourcePath = path.resolve(process.cwd(), sourcePath);
-
-    if (fs.existsSync(absoluteSourcePath)) {
-      [...Array(2)].forEach((_, i) => {
-        if (!fs.existsSync(`${absoluteSourcePath}/test_svg_${i}.svg`)) {
-          fs.writeFileSync(`${absoluteSourcePath}/test_svg_${i}.svg`, svgStub);
-        }
-      });
-      fs.rmSync(`${absoluteSourcePath}/test_svg_0.vue`, { recursive: true });
-      fs.rmSync(`${absoluteSourcePath}/test_svg_1.vue`, { recursive: true });
-    }
+    generateTestSVGs();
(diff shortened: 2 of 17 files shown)


packages/x-svg-converter/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "url": "https://github.com/empathyco/x.git",
+    "directory": "packages/x-svg-converter"
+  },
+  "engines": {
+    "node": "16"
+  },
+  "scripts": {
+    "build": "tsc",
+    "postbuild": "npm pack",
+    "test": "jest -i"
```
CachedaCodes, line 30:
```suggestion
    "test": ""
```

packages/x-svg-converter/src/regex-utils.ts
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Adds `fill="none"` to the root of an SVG if it doesn't have it.
+ *
+ * @param svg - The SVG that will be processed.
+ *
+ * @returns The processed SVG.
+ */
+function addFillNoneInRoot(svg: string): string {
+  const matchRootWithoutFillNone = /svg ?:(?!.*?fill="none")(?=.*?>)/gm;
+  return svg.replace(matchRootWithoutFillNone, '$1 fill="none"');
```
CachedaCodes, line 38:
```suggestion
  const matchRootWithoutFillNone = /svg (?:(?!.*?fill="none"))(?=.*?>)/gm;
  return svg.replace(matchRootWithoutFillNone, 'svg fill="none" ');
```

packages/x-svg-converter/src/regex-utils.ts
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Replaces all the colors in the SVG with `currentColor`, expect white.
+ *
+ * @param svg - The SVG that will be processed.
+ *
+ * @returns The processed SVG.
+ */
+function replaceColorWithCurrentColor(svg: string): string {
+  const matchColoredFillOrStroke = /(?<=(fill|stroke)=")(?!#fff"|none).*?(?=")/gim;
```
mnavarroespinosa, line 61:
```suggestion
  const matchColoredFillOrStroke = /(?<=(fill|stroke)=")(?!#fff"|#ffffff"|white|none).*?(?=")/gim;
```

packages/x-svg-converter/src/regex-utils.ts
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @returns The processed SVG.
+ */
+function addXClass(svg: string): string {
+  const matchRoot = /svg /gm;
+  return svg.replace(matchRoot, 'svg :class="[\'x-icon\'].concat(data.staticClass, data.class)"');
+}
+
+/**
+ * Replaces all the colors in the SVG with `currentColor`, expect white.
```
mnavarroespinosa, line 54:
```suggestion
 * Replaces all the colors in the SVG with `currentColor`, except white.
```

packages/x-svg-converter/src/svg-to-vue.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  runPrettier();
+
+  if (!getParams().keepSVGs) {
+    svgList.forEach(removeSourceSVG);
+  }
+}
+
+/**
+ * Lod the SVG's info from the provided folder in the params.
```
mnavarroespinosa, line 19:
```suggestion
 * Load the SVG's info from the provided folder in the params.
```

mnavarroespinosa (changes requested):
good job!

packages/x-svg-converter/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "@empathyco/x-svg-converter",
+  "version": "0.1.0",
+  "private": true,
```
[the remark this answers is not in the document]
CachedaCodes: Yes, until we decide to release and publish

mnavarroespinosa (changes requested):

packages/x-svg-converter/src/__tests__/svg-to-vue.spec.ts
```diff
@@ -0,0 +1,102 @@
+import path from 'path';
+import fs from 'fs';
+import { svgToVue } from '../svg-to-vue';
+import svgStub from './svg-stub';
+
+// Mock to prevent the prettier from running in test env.
+jest.mock('child_process', () => {
```
ajperezbau, line 7: What do you think about adding a test to check if `prettier` is being called after creating the files?

packages/x-svg-converter/src/svg-to-vue.ts
```diff
@@ -0,0 +1,119 @@
...
+}
+
+/**
+ * Load SVGs from provided folder, turn them into Vue Components and apply
+ * prettier to them. By default, the source SVGs are deleted unless the `keepSVGs`
+ * parameter is passed.
+ *
+ */
+export function svgToVue(): void {
+  const svgList = loadSVGsFromFolder();
```
ajperezbau, line 18: Should we catch here the errors and give a more user-friendly output?
Like showing a text with the usage of the script or something.

packages/x-svg-converter/src/__tests__/svg-to-vue.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  afterEach(() => {
+    jest.resetAllMocks();
+
+    generateTestSVGs();
+    removeVueTestFiles();
+  });
+
+  it('should create a vue component for each svg in the source folder', () => {
+    process.argv = ['param1', 'param2', sourcePath];
```
ajperezbau, line 28: What do you think about extracting this to a helper function such as `setScriptArgs` or something like that, and pass to that function only the arguments that go after the first two process args?

I think it can make the test a bit clearer.

ajperezbau (changes requested):
Good job!

packages/x-svg-converter/src/__tests__/svg-to-vue.spec.ts
CachedaCodes: In the case of these tests, we get the same result by setting the parameters in a `beforeEach` and overriding them when needed (is needed in 1 test).

packages/x-svg-converter/src/svg-to-vue.ts
CachedaCodes: I'm guessing you want to catch the errors that we produce in `loadSVGsFromFolder`, `getParams` and `runPrettier`. In that case, don't see why we should catch them here because we can modify the message directly where we are throwing them (and that way we won't catch and format unexpected errors).

Regarding the friendlier output. The script is intended for users of the library, so I don't see the messages too difficult to understand, but if you have a proposal I can change them.

packages/x-svg-converter/src/__tests__/svg-to-vue.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+  beforeEach(() => {
+    process.argv = ['param1', 'param2', sourcePath];
+
+    generateTestSVGs();
+  });
+
+  afterEach(() => {
+    removeVueTestFiles();
+  });
+
```
ajperezbau, line 30: Restore the original process arguments after the tests, just in case.

```suggestion

  afterAll(() => {
    process.argv = originalProcessArgv;
  });
  
```

packages/x-svg-converter/src/__tests__/svg-to-vue.spec.ts
```diff
@@ -0,0 +1,112 @@
...
+  return {
+    __esModule: true,
+    ...originalModule,
+    exec: jest.fn(() => true)
+  };
+});
+
+describe('test SVG to Vue script', () => {
+  const sourcePath = './src/__tests__/svgs';
+
```
ajperezbau, line 20:
```suggestion
  const originalProcessArgv = process.argv;
  
```

packages/x-svg-converter/src/svg-to-vue.ts
ajperezbau: I was referring to just show in the console the error message, instead of all the error stack.
A solution that comes to my mind is to create a custom error class, throw using this class, and in the try/catch check whether the error is a controlled error or not.
If it is a controlled error, showing only the error in the console should be enough, if it is not, it can re-throw the error.
I think that maybe this can be nice to have, but it is not a required behavior, so it is up to you.

ajperezbau (commented):
Good job 👌

packages/x-svg-converter/src/__tests__/svg-to-vue.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+    generateTestSVGs();
+  });
+
+  afterEach(() => {
+    removeVueTestFiles();
+  });
+
+  afterAll(() => {
+    process.argv = originalArgv;
+    generateTestSVGs();
```
CachedaCodes, line 33: The svgs folder can't be empty, or it's not generated

packages/x-svg-converter/src/svg-to-vue.ts
CachedaCodes: Still not convinced xD but we can meet to talk about it.

ajperezbau (approved):
Good job! 👌

mnavarroespinosa (approved):
Good job!

PR #888, opened 2022-11-28
fix(base-result-image): reset images state when results prop changes

joseacabaneros:
[MOTPRD-9540](https://searchbroker.atlassian.net/browse/MOTPRD-9540)

`BaseResultImage` didn't restore its images state when the `result` prop changed. So if the `images` array of the `result` changed, the `BaseResultImage` didn't display the current images. 
This is the case of `ResultVariantsProvider`. The selected variant is merged with the result info and if the variant contains `images`, the `result` provided to the `BaseResultImage` component contains the images of the variant, but they are not reactively referenced.

EXTRA: I would like to do a cypress component test, but I don't know how to change the prop value dynamically. Something like: 

```TS
  it('resets images state when `result` prop changes', () => {
    const result = createResultStub('Result', { images: ['/img/test-image-1.jpeg'] });
    const { component, getResultPictureImage } = mountBaseResultImage({
      result
    });

    getResultPictureImage().should('have.attr', 'src', '/img/test-image-1.jpeg');
    // Change `result` prop replacing `images` for `['/img/test-image-2.jpeg']`;
    getResultPictureImage().should('have.attr', 'src', '/img/test-image-2.jpeg');
  });
```

### DIFF
--- a/packages/x-components/src/components/result/base-result-image.vue
+++ b/packages/x-components/src/components/result/base-result-image.vue
@@ -41,7 +41,7 @@
 <script lang="ts">
   import { Result } from '@empathyco/x-types';
   import Vue from 'vue';
-  import { Component, Prop } from 'vue-property-decorator';
+  import { Component, Prop, Watch } from 'vue-property-decorator';
   import { NoElement } from '../no-element';
 
   /**
@@ -96,7 +96,7 @@
      *
      * @internal
      */
-    protected pendingImages: string[] = [...(this.result.images ?? [])];
+    protected pendingImages: string[] = [];
 
     /**
      * Contains the images that have been loaded successfully.
@@ -133,6 +133,17 @@
       pointerEvents: 'none !important',
       visibility: 'hidden !important'
     };
+
+    /**
+     * Initializes images state and resets when the result changes.
+     *
+     * @internal
+     */
+    @Watch('result', { immediate: true })
+    resetImagesState(): void {
+      this.pendingImages = [...(this.result.images ?? [])];
+      this.loadedImages = [];
+    }
 
     /**
      * Animation to be used.
--- a/packages/x-components/src/components/result/base-result-image.vue
+++ b/packages/x-components/src/components/result/base-result-image.vue
@@ -135,11 +135,11 @@
     };
 
     /**
-     * Initializes images state and resets when the result changes.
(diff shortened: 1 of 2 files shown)


packages/x-components/src/components/result/base-result-image.vue
```diff
@@ -134,6 +134,17 @@
       visibility: 'hidden !important'
     };
 
+    /**
+     * Initializes images state and resets when the result changes.
+     *
+     * @internal
+     */
+    @Watch('result', { immediate: true })
```
javieri-empathy, line 142: I would probably watch `result.images`, just in case the images arrays mutates.

packages/x-components/src/components/result/base-result-image.vue
```diff
@@ -41,7 +41,7 @@
 <script lang="ts">
   import { Result } from '@empathyco/x-types';
```
javieri-empathy, line 42: About adding a test with Cypress component testing, the problem with that framework is that it is intended to test the component from a user perspective, so if you want to change props, you have to build a system that allows the user to change them. It is more an integration test than pure unit testing. 
For example, in this case I imagine you can write a test combining the component with the ResultProvider/ResultSelector components. Which will give you 100% real guarantee that these components work properly together, which is really nice.

javieri-empathy (changes requested):

packages/x-components/src/components/result/base-result-image.vue
javieri-empathy: That should check both `result` and `images`.

joseacabaneros: Do you suggest to change watch for `'result.images'`?

packages/x-components/src/components/result/base-result-image.vue
joseacabaneros: Check the test I added and if it fits with the requirements

packages/x-components/src/components/result/base-result-image.vue
javieri-empathy: Yes: `@Watch('result.images', { immediate: true })`

javieri-empathy (approved):

LuisMartinez15 (approved):
good job!

PR #899, opened 2022-12-02
feat(desing-system): Add `ghost` variant for `suggestion` in X Design System

tajespasarela:
EX-7498

### DIFF
--- a/packages/x-tailwindcss/src/x-tailwind-plugin/components/suggestion/default.ts
+++ b/packages/x-tailwindcss/src/x-tailwind-plugin/components/suggestion/default.ts
@@ -19,7 +19,6 @@ export function suggestionDefault(helpers: TailwindHelpers) {
     justifyContent: 'center',
     alignItems: 'baseline',
     alignContent: 'center',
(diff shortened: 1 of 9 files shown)


packages/x-tailwindcss/src/x-tailwind-plugin/components/suggestion/default.ts
```diff
@@ -14,7 +14,8 @@ export function suggestionDefault(helpers: TailwindHelpers) {
     fontFamily: theme('fontFamily.main'),
     fontWeight: theme('fontWeight.regular'),
 
-    display: 'flex',
+    display: 'grid',
+    gridAutoFlow: 'column',
     justifyContent: 'center',
     alignItems: 'baseline',
     alignContent: 'center',
```
annacv, line 21: Hi @tajespasarela: I think we should remove `flexFlow: 'row nowrap'` property as we are now using the `CSSGrid` system

annacv (changes requested):
gj!

annacv (approved):
🚀

PR #903, opened 2022-12-07
feat(search-box): add status using a state machine

tiborux:
To test it, check the state machine in: packages/x-components/src/x-modules/search-box/search-box.machine.ts

There you can see which events should change the status of the search box. So you can open the vue dev tools and check the store if the status is changing based on the defined steps.

EX-6474


packages/x-components/src/x-modules/search-box/store/actions/set-status.action.ts
```diff
@@ -0,0 +1,19 @@
+import { SearchBoxXStoreModule } from '../types';
+import { StateMachine } from '../../../../services/state-machine.service';
+import { SearchBoxMachine } from '../../search-box.machine';
+
+const machine = new StateMachine(SearchBoxMachine);
```
tiborux, line 5: I am not sure where should be created the machine

packages/x-components/src/x-modules/search-box/store/module.ts
```diff
@@ -9,17 +10,22 @@ import { SearchBoxXStoreModule } from './types';
  */
 export const searchBoxXStoreModule: SearchBoxXStoreModule = {
   state: () => ({
-    query: ''
+    query: '',
+    status: 'initial'
```
tiborux, line 14: Since we have status property in other modules, maybe we should name it state or something else. Or change the status in the other modules to be a state machine.

javieri-empathy: You have the same problem with `state`. If I have to choose between the 2 I would leave status. In the other modules it refers to a request status, in this one as we don't have a request it refers to the input status.

packages/x-components/src/x-modules/search-box/store/actions/set-status.action.ts
javieri-empathy: I think here it is fine. Next to the action that uses it. I would suggest to inline the configuration in the constructor `new StateMachine({...})`. Or do you have any plan to use it elsewhere?

javieri-empathy (changes requested):

packages/x-components/src/x-modules/search-box/search-box.machine.ts
```diff
@@ -0,0 +1,34 @@
+import { Machine } from '../../services/services.types';
+import { XEvent } from '../../wiring/events.types';
+import { Status } from './store/types';
+
+/**
+ * State machine that defines the different states through which the search box
+ * can transition.
+ */
+export const SearchBoxMachine: Machine<Status, XEvent> = {
```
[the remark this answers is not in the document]
tiborux: Oh I have to update the diagram sorry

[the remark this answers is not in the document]
tiborux: I updated the diagram in the task and I should change some transitions in the machine

packages/x-components/src/x-modules/search-box/store/module.ts
diegopf: if it refers to input status why don't just called it that? `inputStatus`

packages/x-components/src/x-modules/search-box/store/actions/set-status.action.ts
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * @param context - The {@link https://vuex.vuejs.org/guide/actions.html | context} of the actions,
+ * provided by Vuex.
+ *
+ * @param event - The event used to transition the state machine.
+ * @public
+ */
+export const setStatus: SearchBoxXStoreModule['actions']['setStatus'] = ({ commit }, event) => {
+  machine.transition(event);
+  commit('setStatus', machine.currentState);
```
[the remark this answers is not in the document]
diegopf: Following https://github.com/empathyco/x/pull/903/files#r1043061784 maybe we can go with a more explicit name `setInputStatus`

packages/x-components/src/x-modules/search-box/search-box.machine.ts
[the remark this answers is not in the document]
tiborux: image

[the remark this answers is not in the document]
tiborux: shit...I didn't realize about loading a query from the URL.

packages/x-components/src/x-modules/search-box/search-box.machine.ts
```diff
@@ -0,0 +1,34 @@
...
+      UserIsTypingAQuery: 'typing'
+    },
+    typing: {
+      UserIsTypingAQuery: 'typing',
+      UserAcceptedAQuery: 'filled',
+      UserClearedQuery: 'empty'
+    },
+    filled: {
+      UserFocusedSearchBox: 'focused',
+      UserClearedQuery: 'empty'
```
diegopf, line 23: According to the diagram I think it is missing 
```suggestion
      UserClearedQuery: 'empty',
      UserIsTypingAQuery: 'typing'
```

packages/x-components/src/x-modules/search-box/search-box.machine.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+export const SearchBoxMachine: Machine<Status, XEvent> = {
+  initial: 'initial',
+  states: {
+    initial: {
+      UserAcceptedAQuery: 'filled',
+      UserIsTypingAQuery: 'typing'
+    },
+    typing: {
+      UserIsTypingAQuery: 'typing',
```
diegopf, line 17: We can remove this, can't we?

```suggestion

```

diegopf (changes requested):

packages/x-components/src/x-modules/search-box/search-box.machine.ts
tiborux: yep

support-empathy:
 [Check out EPDocs preview](https://docs-dev.empathy.co/preview/x-903/index.html)

diegopf (approved):

packages/x-components/src/x-modules/search-box/wiring.ts
```diff
@@ -46,20 +50,36 @@ const clearSearchBoxQuery = wireCommit('setQuery', '');
...
 /**
  * SearchBox wiring.
  *
  * @internal
  */
 export const searchBoxWiring = createWiring({
   ParamsLoadedFromUrl: {
-    setUrlParams
+    setUrlParams,
+    transitionState: filter(setInputStatus('UserAcceptedAQuery'), ({ eventPayload: urlParams }) => {
```
javieri-empathy, line 72: Hmm smells like ñapa... I don't think this event re-mapping is a good idea... Probably the url module should emit the `UserAcceptedAQuery` event itself somehow... But not for this PR

javieri-empathy (approved):

PR #906, opened 2022-12-08
test(e2e): make `HistoryQueries` test wait for no results before runn…

javieri-empathy:
Rendered history queries list is empty if no results are provided test does not wait until the no-results is triggered, so it can fail as the needed events to store the number of results for the history query might not have been emitted


tiborux (approved):

LuisMartinez15 (approved):

tiborux (approved):
gj

PR #917, opened 2022-12-09
build: update to vue 2.7

javieri-empathy:
Draft to update to vue 2.7

We should test this branch with:

- Archetype. So far it works for me. I haven't seen any weird behaviors, warnings, or errors.
- Motive
- Empathy-components

### DIFF
--- a/packages/react-wrapper/package.json
+++ b/packages/react-wrapper/package.json
@@ -42,7 +42,7 @@
   "peerDependencies": {
     "react": "^16.9.0",
     "react-dom": "^16.9.0",
-    "vue": "~2.7.0"
+    "vue": "^2.6.0"
   },
   "devDependencies": {
     "@types/jest": "~27.0.3",
@@ -51,8 +51,8 @@
     "cypress": "~8.4.0",
     "eslint-plugin-react": "~7.20.0",
     "jest": "~27.3.1",
-    "react": "^16.9.0",
-    "react-dom": "^16.9.0",
+    "react": "~16.9.0",
+    "react-dom": "~16.9.0",
     "start-server-and-test": "~1.14.0",
     "ts-jest": "~27.0.7",
     "typescript": "~4.6.2",
--- a/packages/react-wrapper/package.json
+++ b/packages/react-wrapper/package.json
@@ -42,7 +42,7 @@
   "peerDependencies": {
     "react": "^16.9.0",
     "react-dom": "^16.9.0",
-    "vue": "~2.6.0"
+    "vue": "~2.7.0"
   },
   "devDependencies": {
     "@types/jest": "~27.0.3",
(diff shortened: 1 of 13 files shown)


packages/x-components/src/components/layouts/fixed-header-and-asides-layout.vue
```diff
@@ -108,8 +109,8 @@
   })
   export default class FixedHeaderAndAsidesLayout extends mixins(LayoutsMixin) {
     protected scrollPosition = 0;
-    protected rightAsideAnimation = animateTranslate('right');
-    protected leftAsideAnimation = animateTranslate('left');
+    protected rightAsideAnimation: VueConstructor = animateTranslate('right');
+    protected leftAsideAnimation: VueConstructor = animateTranslate('left');
```
javieri-empathy, line 113: I don't know why, if we let these two properties type to be inferred, it adds a `MainScroll` type there, which is incorrect and very strange.

packages/x-components/src/shims-tsx.d.ts
```diff
@@ -4,8 +4,5 @@ declare global {
   namespace JSX {
     interface Element extends VNode {}
     interface ElementClass extends Vue {}
-    interface IntrinsicElements {
-      [elem: string]: any;
-    }
```
javieri-empathy, line 9: We don't use JSX but this is already  a record type in the base JSX definition

packages/react-wrapper/src/__tests__/tsconfig.json
```diff
@@ -1,8 +1,8 @@
 {
   "extends": "../../tsconfig.json",
   "compilerOptions": {
-    "types": ["jest"]
+    "types": ["jest", "./jsx.d.ts"]
   },
   "include": ["**/*.ts", "**/*.tsx"],
-  "exclude": []
+  "exclude": ["node_modules"]
```
javieri-empathy, line 7: Every time we add an `exclude` property we should add `node_modules`

packages/react-wrapper/package.json
```diff
@@ -42,7 +42,7 @@
   "peerDependencies": {
     "react": "^16.9.0",
     "react-dom": "^16.9.0",
-    "vue": "~2.6.0"
+    "vue": "^2.6.0"
```
javieri-empathy, line 45: As the code does not change, we can still keep the 2.6 version as the oldest we support with react-wrapper

packages/x-components/src/shims-tsx.d.ts
javieri-empathy: I think we can totally get rid of this file. Improting from Vue will now load all the jsx custom types, which is the issue why it conflicts with React

PR #919, opened 2022-12-12
ci: use matrix steps and jest projects

tiborux:
I decided to divide the current build into multiple steps using [Matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs)

With the matrix, you can run in parallel different steps, because they run on different runners.

I tried to separate the build but I was not able to achieve it. I tried to upload the artifact, but it takes too much time and I think that it's not a good solution, because we will spend more time uploading the artifact and downloading it than run the build in all steps.

Also, I separate the test into 3 different steps, and using jest projects, we are able to run in parallel the different test for the different packages.

With the current change, running on matrix, we win too much time on "relaunching" the build, because we are going to be able to run only the failed step.

Posts of interest:

[Reduce time on gh actions](https://manzoorwani.dev/how-we-brought-down-github-actions-workflow-execution-time-from-20-minutes-to-5-minutes/)

[Impoving jest](https://bharathvaj.me/blog/speedup-jest-github-actions)

[Jest projects](https://jestjs.io/docs/configuration#projects-arraystring--projectconfig)

PD: I tried to use --changeSince in jest, but It only detects if you change a test file, I mean, if you change a `file.vue`, it will not be detected as a change to run the test again.

The pending thing will be to separate the different actions and create workflows using those actions.


package.json
```diff
@@ -7,7 +7,9 @@
     "build": "lerna run build",
     "lint": "eslint packages --ext .ts,.tsx,.vue --quiet --cache --cache-strategy content",
     "lint:fix": "eslint packages --ext .ts,.tsx,.vue --fix",
-    "test": "lerna run test --parallel",
+    "test:unit": "jest",
```
diegopf, line 10: I see that running this locally it creates a file `rootDir/output/en.messges.json` but I'm still not sure why

packages/x-components/src/x-modules/next-queries/components/__tests__/next-queries.spec.ts
```diff
@@ -126,12 +126,12 @@ describe('testing next queries component', () => {
             <NextQuery :suggestion="suggestion"
                        :highlightCurated="highlightCurated">
               <template #default="suggestionContentScope">
-                <span data-test="next-query-highlight">
-                  <img
+              <span data-test="next-query-highlight">
+                  <img\
```
diegopf, line 130:
```suggestion
                  <img
```

packages/x-priority-queue/src/x-priority-queue.ts
```diff
@@ -91,7 +91,7 @@ export class XPriorityQueue<Key = string> {
    *
    * @internal
    */
-  protected comparatorFn: AnyFunction<boolean>;
+  protected comparatorFn: AnyFunction;
```
diegopf, line 94: I guess changes in this file are not intended

packages/x-tailwindcss/CHANGELOG.md
```diff
@@ -1,16 +1,15 @@
 # Change Log
 
-All notable changes to this project will be documented in this file.
-See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
+All notable changes to this project will be documented in this file. See
```
diegopf, line 3: I guess changes in this file are not intended

packages/x-components/CHANGELOG.md
```diff
@@ -1,16 +1,15 @@
 # Change Log
```
diegopf, line 2: I guess changes in this file are not intended

diegopf (changes requested):

packages/x-priority-queue/src/x-priority-queue.ts
tiborux: we have problem with types in this file, if you change that and run the test, you will see that an error will be launched.

packages/x-tailwindcss/CHANGELOG.md
tiborux: it was the linter, I dont know why but the file was wrong. check how the other versions has the See word in the first line

.github/workflows/build.yml
```diff
@@ -25,28 +29,32 @@ jobs:
...
-        with:
-          path: ~/.cache/Cypress
-          key: cypress-cache
       - name: Cache eslint
+        if: ${{ matrix.task == 'lint' }}
         id: cache-eslint
         uses: actions/cache@v3
         with:
           path: .eslintcache
           key: eslint-${{github.sha}}
```
ajperezbau, line 38: A restore-key can be added here to have an incremental cache with `eslint`.
If a cache for the commit `sha` doesn't exist, it will retrieve the cache of the last commit.
When the lint has finished, it will save a new cache for the current commit.

```suggestion
          key: eslint-${{github.sha}}
          restore-keys: eslint-
```

.github/workflows/build.yml
```diff
@@ -25,28 +29,32 @@ jobs:
...
+      - name: Cache Cypress
+        if: ${{ matrix.task == 'test:e2e:ci' || matrix.task == 'test:component' }}
+        id: cache-cypress
+        uses: actions/cache@v3
+        with:
+          path: ~/.cache/Cypress
+          key: cypress-cache
       - name: Install lerna and all packages
```
ajperezbau, line 46: To save the `jest` cache, something like this can be done:

```
      - name: Cache Jest
        if: ${{ matrix.task == 'test:unit' }}
        id: cache-jest
        uses: actions/cache@v3
        with:
          path: ./tmp/jest
          key: jest-${{github.sha}}
          restore-keys: jest-
```

For this to work, we should change the cache directory of `jest`.
To do it, I would create a `base-config` on the root of the monorepo, set there the `cacheDirectory` and extend from that base config in all the `jest.config` files.

Something like:

`jest-base.config.js`
```
module.exports = {
  cacheDirectory: '<rootDir>/../../tmp/jest'
};
```

`package/*/jest-config.js`
```
const base = require('../../jest-base.config');

module.exports = {
  ...base,
  // The package specific jest options ...
};
```

Also add the `tmp` directory to the `.gitignore`.

ajperezbau (commented):
Good job! 👌

packages/x-svg-converter/src/__tests__/svg-to-vue.spec.ts
```diff
@@ -89,7 +89,7 @@ describe('test SVG to Vue script', () => {
    * Generate the SVG files used in the tests.
    */
   function generateTestSVGs(): void {
-    const absoluteSourcePath = path.resolve(process.cwd(), sourcePath);
+    const absoluteSourcePath = path.resolve(__dirname, sourcePath);
```
ajperezbau, line 92: I would say that this `absoluteSourcePath` is not needed, the `sourcePath` is already absolute, it is generated above with `path.resolve` and it contains already the `svg` folder for the tests.
So I would use `sourcePath` directly instead of creating this variable.

packages/x-svg-converter/src/__tests__/svg-to-vue.spec.ts
```diff
@@ -104,7 +104,7 @@ describe('test SVG to Vue script', () => {
    * Remove the vue files product of the tests.
    */
   function removeVueTestFiles(): void {
-    const absoluteSourcePath = path.resolve(process.cwd(), sourcePath);
+    const absoluteSourcePath = path.resolve(__dirname, sourcePath);
```
ajperezbau, line 107: The same here, I would just use `sourcePath`.

packages/x-translations/src/__tests__/utils.spec.ts
```diff
@@ -32,6 +32,6 @@ describe('test utils', () => {
 
     exportToFile('en.messages.json', 'Test file');
 
-    expect(fs.existsSync('./output/en.messages.json')).toBe(true);
+    expect(fs.existsSync(path.join(__dirname, '../../output/en.messages.json'))).toBe(true);
```
diegopf, line 35:
```suggestion
    expect(fs.existsSync(path.join(__dirname, '../../output/en.messages.json'))).toBe(true);
    expect(fs.existsSync(pathFromProjectRoot('/output/en.messages.json'))).toBe(true);
```

.github/actions/cache/action.yml
```diff
@@ -0,0 +1,46 @@
+name: Enables Cache
+description: Enables the cache for different task
```
diegopf, line 2:
```suggestion
description: Enables the cache for different tasks
```

.github/actions/cache/action.yml
```diff
@@ -0,0 +1,46 @@
...
+    - name: Cache Jest
+      if: ${{ inputs.task_name == 'test:unit' }}
+      id: cache-jest
+      uses: actions/cache@v3
+      with:
+        path: ./tmp/jest
+        key: jest-${{github.sha}}
+        restore-keys: jest-
+    - name: Cache Cypress
+      if: ${{ inputs.task_name== 'test:e2e:ci' || inputs.task_name == 'test:unit-cypress' }}
```
diegopf, line 41:
```suggestion
      if: ${{ inputs.task_name == 'test:e2e:ci' || inputs.task_name == 'test:unit-cypress' }}
```

diegopf (approved):

diegopf (dismissed):

ajperezbau (approved):
Niice

diegopf (approved):

PR #921, opened 2022-12-12
Create Readme for x-adapter

annacv:
[EX-6046](https://searchbroker.atlassian.net/browse/EX-6046).

- [x] `x-adapter` Readme

### DIFF
--- a/packages/x-adapter/README.md
+++ b/packages/x-adapter/README.md
@@ -52,7 +52,7 @@ configured are:
 - `defaultRequestOptions`: Default values for the endpoint configuration. You can use it to define
   if a request is cancelable, a unique id to identify it, anything but the `endpoint` can be set. 
   Check [EndpointAdapterOptions](https://github.com/empathyco/x/blob/main/packages/x-adapter/src/endpoint-adapter/types.ts) to see the available options.
-- `requestMapper`: A function to apply to the request parameters and transform them.
+- `requestMapper`: A function to transform the unprocessed request into parameters the API can understand.
 - `responseMapper`: A function to apply to the response and transform it.
 
 <br>
--- a/packages/x-adapter/README.md
+++ b/packages/x-adapter/README.md
@@ -217,7 +217,7 @@ npm test
 ## Contributing
 
 To start contributing to the project, please take a look to
-our **[Contributing Guide](https://github.com/empathyco/x/blob/HEAD/.github/CONTRIBUTING.md).**
+our **[Contributing Guide](https://github.com/empathyco/x/blob/main/.github/CONTRIBUTING.md).**
 
 ## License
 
--- a/packages/x-adapter/README.md
+++ b/packages/x-adapter/README.md
@@ -40,7 +40,7 @@ async function searchOnClick() {
 
 ### Implement your own adapter
 
-To create an `EndpointAdapter` you can use the `endpointAdapterFactory` options. This function will
+To create an `EndpointAdapter` you can use the `endpointAdapterFactory` function. This function will
 receive an `EndpointAdapterOptions` object containing all the needed data to perform and map a
 request, and return a function that when invoked will trigger the request. The options that can be
 configured are:
--- a/packages/x-adapter/README.md
+++ b/packages/x-adapter/README.md
@@ -164,7 +164,6 @@ definition should be agnostic enough to support the change:
 export const getItemById = endpointAdapterFactory({
   endpoint: 'https://dummyjson.com/quotes/{id}',
   // ... rest of options to configure
-  }
 });
 
 // You would pass the new endpoint in the function call
--- a/packages/x-adapter/README.md
+++ b/packages/x-adapter/README.md
@@ -174,9 +174,9 @@ getItemById({ id: 1 }, { endpoint: 'https://dummyjson.com/products/{id}');
 
 #### Using the httpClient function
 
-The `x-adapter` uses the `Fetch API` interface by default to perform the requests. But you can use
-your own HTTP Client as part of the configurable `EndpointAdapterOptions`. The `httpClient` is a
-function type that accepts two parameters: the `endpoint` and an additional
+Every adapter created using `endpointAdapterFactory` uses the `Fetch API` by default to perform the requests. But you can use
+your own `HttpClient` as part of the configurable `EndpointAdapterOptions`. A `HttpClient` is a
+function that accepts two parameters: the `endpoint` string, and an additional
 [`options`](https://github.com/empathyco/x/blob/main/packages/x-adapter/src/http-clients/types.ts)
 object to make the request with.
 
--- a/packages/x-adapter/README.md
+++ b/packages/x-adapter/README.md
@@ -141,33 +141,28 @@ async function searchOnClick() {
 
 #### Using a dynamic endpoint
 
-You can use a mapper function to do so and pass the dynamic part of the url as a parameter when the
-adapter function is called.
-
-```ts
-export const getProductById = endpointAdapterFactory({
-  endpoint: ({ id }: GetProductByIdRequest) => 'https://dummyjson.com/products/' + id,
-  responseMapper(response: Readonly<ApiProduct>): AppProduct {
-    return productMap(response);
-  }
-});
-```
-
-Or you can place your parameter(s) inside curly brackets, and they will be added to the final
-string, thanks to the
-[interpolate](https://github.com/empathyco/x/blob/main/packages/x-adapter/src/utils/interpolate.ts)
-utility function. If a value is not provided, its parameter would just will be removed from the
-string.
+If you need to generate an endpoint url dynamically, you can add parameters inside curly brackets to
+the endpoint string. By default, these parameters will be replaced using the request data. If a
+parameter is not found inside the request, an empty string will be used.
 
 ```ts
 export const getItemById = endpointAdapterFactory({
   endpoint: 'https://dummyjson.com/{section}/{id}'
   // ... rest of options to configure
 });
+getItemById({ section: 'products', id: 1 }); // 'https://dummyjson.com/products/1'
+getItemById({ section: 'quotes', id: 3 }); // 'https://dummyjson.com/quotes/3'
+getItemById({ section: 'quotes' }); // 'https://dummyjson.com/quotes/'
+```
 
-// You would pass the parameter's value in the function call
-getItemById({ section: 'products', id: 1 });
-getItemById({ section: 'quotes', id: 3 });
+For more complex use cases, you can use a mapper function. This function receives the request, and
+must return the URL string.
+
+```ts
+export const getProductById = endpointAdapterFactory({
+  endpoint: ({ id }: GetProductByIdRequest) => 'https://dummyjson.com/products/' + id
+  // ... rest of options to configure
+});
 ```
 
 Additionally, you can also overwrite your adapter's endpoint definition using the
@@ -426,10 +421,10 @@ it:
 
 - `replace`: Replaces completely the original Schema.
 - `override`: Merges the original schema with the new one.
-- `extends`: Creates a new Schema based on the original Schema. The original remains unchanged.
+- `extends`: Creates a new Schema based on the original one. The original remains unchanged.
 
 ```ts
-// TODO: Use mutable schema's methods: '$replace', '$override', '$extends'
+// TODO: Mutable schema's methods: '$replace', '$override', '$extends'
 ```
 
 <br>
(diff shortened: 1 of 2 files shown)


packages/x-adapter/README.md
```diff
@@ -1,13 +1,224 @@
...
+## Contributing
+
+To start contributing to the project, please take a look to
+our **[Contributing Guide](https://github.com/empathyco/x/blob/HEAD/.github/CONTRIBUTING.md).**
+
+## License
 
-To start contributing to the project, please take a look to our
-[Contributing Guide](../../.github/CONTRIBUTING.md).
+[empathyco/x License](https://github.com/empathyco/x/blob/main/LICENSE)
```
herrardo, line 224: We should put the license file in every package. Could you add it to this package? You can copy the same file that we have in x-components package: https://github.com/empathyco/x/blob/main/packages/x-components/LICENSE

Then, in this README, link it to that new file.

packages/x-adapter/README.md
```diff
@@ -1,13 +1,224 @@
...
+| <a href="https://github.com/CachedaCodes"><img src="https://avatars.githubusercontent.com/u/7124620?v=4" width="100px;" alt=""/><br /><br /><b>Guillermo Cacheda</b></a> |      <a href="https://github.com/diegopf"><img src="https://avatars.githubusercontent.com/u/7504736?v=4" width="100px;" alt=""/><br /><br /><b>Diego Pascual</b></a>       | <a href="https://github.com/mnavarroespinosa"><img src="https://avatars.githubusercontent.com/u/77450928?v=4" width="100px;" alt=""/><br /><br /><b>Manuel Navarro</b></a> | <a href="https://github.com/ajperezbau"><img src="https://avatars.githubusercontent.com/u/75546736?v=4" width="100px;" alt=""/><br /><br /><b>Abraham Pérez</b></a> |
+
+## Contributing
+
+To start contributing to the project, please take a look to
+our **[Contributing Guide](https://github.com/empathyco/x/blob/HEAD/.github/CONTRIBUTING.md).**
```
herrardo, line 220: I think is better use the link with the branch instead:

```suggestion
our **[Contributing Guide](https://github.com/empathyco/x/blob/main/.github/CONTRIBUTING.md).**
```

packages/x-adapter/README.md
```diff
@@ -1,13 +1,224 @@
...
+
+```
+npm test
+```
+
+## Changelog
+
+[Changelog summary](https://github.com/empathyco/x/blob/main/packages/x-adapter/CHANGELOG.md)
+
+## Contributors
```
herrardo, line 211: I think I wouldn't add a specific section for contributors in the package, as we have one for the whole repository, and adding one to each package seems too granular and difficult to maintain.

herrardo (changes requested):
A little review! Good job!

packages/x-adapter/README.md
```diff
@@ -1,13 +1,218 @@
...
+}
+interface ApiResponseModel {
+  items: T[];
+  limit?: number;
+  offset?: number;
+  total: number;
+}
+
+// Maps both the request and the response to connect your model with API you are working with.
+const searchRequestSchema: Schema<SearchRequest, ApiRequestModel> = {
```
javieri-empathy, line 129: I miss some talk about the features that schema has, like using paths, functions, mapping arrays, type safety...

packages/x-adapter/README.md
```diff
@@ -1,13 +1,218 @@
...
+  searchRequestSchema
+);
+export const searchResponseMapper = schemaMapperFactory<ApiResponseModel, SearchResponse>(
+  searchResponseSchema
+);
+```
+
+#### Create more complex models with Subschemas
+
+For more complex data mapping, the `x-adapter` library also provides a `subSchema` type that
```
javieri-empathy, line 150: I wouldn't say that `x-adapter` provides with `subSchemas`.  The idea of subschemas is being able to re-use a schema between different adapters and with different paths. Something like this maybe?. I would also change the example reusing the same schema under different paths in two adapters. 
> When you are creating adapters for different APIs you might find the case that you have to map the same model in different places. To help you with that, schemas allows to use subschemas. To use them you just have to provide with the path of the data to map, and the schema to apply to it.

packages/x-adapter/README.md
```diff
@@ -1,13 +1,218 @@
...
+    }
+  }
+});
+```
+
+#### Using a mutable schema
+
+The `x-adapter` library also provides a `MutableSchema` type, a `Schema` that can replace an
+original schema, override it or create a new one based on the original one. This can be useful to
+define a shared pattern that you will need to reuse in different endpoint adapters.
```
javieri-empathy, line 178: I wouldn't say that "a `MutableSchema` type, a `Schema` that can replace an
original schema, override it". 

> With the `x-adapter` you can create mutable schemas. These schemas are helpful when writing libraries where you want to have some defaults, but allow to modify or extend them.

I would also add an example for each one of the options that the mutable schemas has.

packages/x-adapter/README.md
```diff
@@ -1,13 +1,218 @@
...
+
+### Extend an adapter that uses schemas
+
+You can check the
+[x-platform-adapter](https://github.com/empathyco/x/tree/main/packages/x-adapter-platform) library.
+You will find a sample implementation of the `x-adapter` library based on the
+[Search Platform API](https://docs.empathy.co/develop-empathy-platform/api-reference/search-api.html),
+and also some guidance on how to extend it for your needs.
+
+## Test
```
javieri-empathy, line 201: I would add a little introduction, explaining that we use `jest` for the tests and where are they located.

packages/x-adapter/README.md
```diff
@@ -1,13 +1,218 @@
 # x-adapter
 
-`x-adapter` is a library of utils to ease the communication with any API. Some features that it
-provides:
+**Empathy Search Adapter** is a library of utils to ease the communication with any API.
```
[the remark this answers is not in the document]
annacv: It's the way I found it is named in docs: 
`Empathy Platform Search Adapter to communicate with the Empathy Platform Search API (or the Empathy Search Adapter to connect with the search API you are using).` [(in before you begin section)](https://docs.empathy.co/develop-empathy-platform/build-search-ui/web-archetype-development-guide.html#configuring-the-plugin)

packages/x-adapter/README.md
```diff
@@ -1,13 +1,218 @@
...
-### Contributing
+## Installation
+
+```
+npm i @empathyco/x-adapter
+```
+
+You can additionally install the
+[@empathyco/x-types](https://github.com/empathyco/x/tree/main/packages/search-types) package and
+take the advantage of it in your project development.
```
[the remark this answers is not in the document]
annacv: I'll do so. 
Based on this I see I was mistaken: I understood that they could be used for type safety, regardless of whether or not `x-components` was used. So thanks for the appointment 👀

packages/x-adapter/README.md
```diff
@@ -1,13 +1,218 @@
...
+
+## Configuration & Usage
+
+The `Adapter` instance is meant to be a configuration object for each `EndpointAdapter`, handling
```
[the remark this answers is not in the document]
annacv: Hi @javieri-empathy, 🤔 the idea was to show that in the end, they would have an adapter object to export & use in their app. So should we add this part at the end? Or I'm mistaken/losing something??? 🙏

packages/x-adapter/README.md
```diff
@@ -1,13 +1,218 @@
...
+each data model that will need to be passed/mapped/transformed to meet your app’s environment and
+the API requirements.
+
+```tsx
```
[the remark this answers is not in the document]
annacv: True! That happened by copy-pasting code blocks from the Notion app (it seems they codify Typescript using tsx 👀 ).

packages/x-adapter/README.md
```diff
@@ -1,13 +1,456 @@
...
+```
+
+<br>
+
+### Implement your own adapter using schemas
+
+The `x-adapter` library includes a `schemaMapperFactory` function that returns a mapper function to
+execute the `schema` you provide. The function accepts a `Source` object (the data to be
+transformed) and a `Target` object. You just need to create the request and response models to
+define the structure of your data vs the APIs data, and pass them to the factory function.
```
javieri-empathy, line 233: Let's try to justify using schemas.

```suggestion
Sometimes the transformations you will need to do in the mappers are just renaming parameters. What the API calls `q` might be called `query` in your request. To ease this transformations, `@empathyco/x-adapter` allows to create mappers using schemas.

A schema is just a dictionary where the key is the desired parameter name, and the value is the path of the source object that has the desired value or a simple mapper function if you need to transform the value somehow.
```

packages/x-adapter/README.md
```diff
@@ -1,13 +1,456 @@
...
+interface AppUser {
+  id: string;
+  name: string;
+}
+```
+
+###### Schema's mapper factory function implementation
+
+```ts
+// Object Mapper to use in the Response Mapper
```
javieri-empathy, line 268: Like above, I would just inline this mapper inside the `userSearchResponseMapper` `users.map(...)` call.

javieri-empathy (changes requested):

packages/x-adapter/README.md
```diff
@@ -1,13 +1,456 @@
...
+  }
+});
+
+// You would pass the parameter's value in the function call
+getItemById({ section: "products", id: 1 });
+getItemById({ section: "quotes", id: 3 });
+```
+
+Additionally, you can also overwrite your adapter's endpoint definition using the
+`RequestOptions.endpoint` parameter in the function call. Take in account that your `responseMapper`
```
diegopf, line 161:
```suggestion
`RequestOptions.endpoint` parameter in the function call. Take into account that your `responseMapper`
```

packages/x-adapter/README.md
```diff
@@ -1,13 +1,456 @@
...
+definition should be agnostic enough to support the change:
+
+```ts
+export const getItemById = endpointAdapterFactory({
+  endpoint: 'https://dummyjson.com/quotes/{id}',
+  // ... rest of options to configure
+  }
```
diegopf, line 168:
```suggestion
```

diegopf (changes requested):

packages/x-adapter/README.md
```diff
@@ -1,13 +1,467 @@
...
+// Function call
+async function searchOnClick() {
+  const response = await searchProducts({ query: 'phone' });
+  this.items = response.products;
+}
+```
+
+<br>
+
+#### Using a dynamic endpoint
```
javieri-empathy, line 142: RE-posting because I believe this comment was lost when modifying the file 😅
I would probably invert this section. Start with the simple use case and then go to the advanced one.

Using a dynamic endpoint
If you need to generate an endpoint url dynamically, you can add parameters inside curly brackets to the endpoint string. By default these parameters will be replaced using the request data. If a parameter is not found inside the request, an empty string will be used.
```ts
export const getItemById = endpointAdapterFactory({
  endpoint: 'https://dummyjson.com/{section}/{id}',
  // ... rest of options to configure
  }
});
getItemById({ section: "products", id: 1 }); // 'https://dummyjson.com/products/1'
getItemById({ section: "quotes", id: 3 }); // 'https://dummyjson.com/quotes/3'
getItemById({ section: "quotes" }); // 'https://dummyjson.com/quotes/'
For more complex use cases, you can use a mapper function. This function receives the request, and must return the URL string.

export const getProductById = endpointAdapterFactory({
  endpoint: ({ id }: GetProductByIdRequest) => 'https://dummyjson.com/products/' + id,
  // ... rest of options to configure
});
```

javieri-empathy (changes requested):

packages/x-adapter/README.md
```diff
@@ -1,13 +1,462 @@
...
+- `override`: Merges the original schema with the new one.
+- `extends`: Creates a new Schema based on the original one. The original remains unchanged.
+
+```ts
+// TODO: Mutable schema's methods: '$replace', '$override', '$extends'
+```
+
+<br>
+
+### Extend an adapter that uses schemas
```
annacv, line 432: @javieri-empathy @diegopf: This is the part I doubt to add here this way -----> The idea I have is documenting it in `x-adapter-platform` and linking to that readme section when it's done instead of linking to the full package.

javieri-empathy: I would probably leave it here for the next PR. We can link the platform adapter as an example use case. Think that Extending an adapter is a feature provided by `x-adapter`. It is NOT an specific part of `x-platform-adapter`.

javieri-empathy (approved):

diegopf (approved):

PR #924, opened 2022-12-14
feat(design-system): modify typography component

tiborux:
EX-7638

### DIFF
--- a/packages/x-tailwindcss/src/x-tailwind-plugin/components/typography/title1/sizes.ts
+++ b/packages/x-tailwindcss/src/x-tailwind-plugin/components/typography/title1/sizes.ts
@@ -14,7 +14,7 @@ export function titleSizes({ theme }: TailwindHelpers) {
       lineHeight: theme('lineHeight.md')
     },
     md: {
-      fontSize: theme('fontSize.xl'),
+      fontSize: theme('fontSize.2xl'),
       lineHeight: theme('lineHeight.sm')
     },
     lg: {
--- a/packages/x-tailwindcss/src/x-tailwind-plugin/components/typography/title1/sizes.ts
+++ b/packages/x-tailwindcss/src/x-tailwind-plugin/components/typography/title1/sizes.ts
@@ -10,15 +10,15 @@ import { TailwindHelpers } from '../../../../types';
 export function titleSizes({ theme }: TailwindHelpers) {
   return {
     sm: {
-      fontSize: theme('fontSize.md')
-    },
-    md: {
       fontSize: theme('fontSize.xl'),
       lineHeight: theme('lineHeight.md')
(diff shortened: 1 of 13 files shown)


packages/x-tailwindcss/src/x-tailwind-plugin/components/typography/title1/sizes.ts
```diff
@@ -10,15 +10,15 @@ import { TailwindHelpers } from '../../../../types';
 export function titleSizes({ theme }: TailwindHelpers) {
   return {
     sm: {
-      fontSize: theme('fontSize.md')
+      fontSize: theme('fontSize.xl'),
+      lineHeight: theme('lineHeight.md')
     },
     md: {
       fontSize: theme('fontSize.xl'),
```
mnavarroespinosa, line 17: Check with the design. This should be `xl` or `32px` the values don't match.

packages/x-tailwindcss/src/x-tailwind-plugin/components/typography/title2/sizes.ts
```diff
@@ -10,15 +10,15 @@ import { TailwindHelpers } from '../../../../types';
 export function titleSizes({ theme }: TailwindHelpers) {
   return {
     sm: {
-      fontSize: theme('fontSize.md'),
-      lineHeight: theme('lineHeight.md')
+      fontSize: theme('fontSize.xl')
```
mnavarroespinosa, line 13: Check with the design. This should be `xl` or `20px`, the values don't match.

mnavarroespinosa (changes requested):

mnavarroespinosa:
Is it intended that `Text 1` and `Text 2` are identical, isn't it? Just to double-check.

ajperezbau:
The `font-weight` of the bold variants seem to be 700 in the designs, but it is set to 600 in the project.

ajperezbau (commented):
Nothing more apart from the things that @mnavarroespinosa commented.
As a side note, I was talking with @javieri-empathy before, and he suggested that maybe in the designs only the tokens should be specified, to avoid confusions with the values and to ease the maintenance of the designs.

tiborux:
> Nothing more apart from the things that @mnavarroespinosa commented. As a side note, I was talking with @javieri-empathy before, and he suggested that maybe in the designs only the tokens should be specified, to avoid confusions with the values and to ease the maintenance of the designs. It makes sense to me, we can comment it tomorrow with the team if you think alike.

I think it is better, I usually check the token, not the value...

tiborux:
image

mnavarroespinosa (approved):
good job!

ajperezbau (approved):

PR #940, opened 2022-12-19
refactor: add types to `getSafePropertyChain` and move it to `@empatyco/x-utils`

tajespasarela:
BREAKING CHANGE: `getSafePropertyChain` moved to `@empathyco/x-utils` package.

EX-7647


packages/x-adapter/package.json
```diff
@@ -36,11 +36,10 @@
   },
   "dependencies": {
     "@empathyco/x-deep-merge": "^1.3.0-alpha.25",
-    "@empathyco/x-get-safe-property-chain": "^1.3.0-alpha.4",
+    "@empathyco/x-utils": "^1.0.0-alpha.11",
```
javieri-empathy, line 39: Should we deprecate the get-safe-property-chain package in npm?

packages/x-utils/src/get-safe-property-chain.ts
```diff
@@ -19,11 +24,11 @@ export function getSafePropertyChain<T = any>(obj: any, propertyChain: string):
...
-export function getSafePropertyChain<T = any>(
-  obj: any,
-  propertyChain: string,
-  defaultReturn?: T
-): T;
+export function getSafePropertyChain<SomeObject, Path extends ExtractPath<SomeObject>>(
+  obj: SomeObject,
+  propertyChain: Path | '',
+  defaultReturn?: ExtractType<SomeObject, Path>
+): ExtractType<SomeObject, Path>;
```
javieri-empathy, line 31: Should this overload just handle the empty string version? Why add both?
```suggestion
export function getSafePropertyChain<SomeObject>(
  obj: SomeObject,
  propertyChain: '',
  defaultReturn?: SomeObject
): SomeObject;
```

packages/x-utils/src/get-safe-property-chain.ts
```diff
@@ -7,7 +9,10 @@
  * @returns The value of the chain of properties, or `undefined` if any of the intermediate values
  * is not defined.
  */
-export function getSafePropertyChain<T = any>(obj: any, propertyChain: string): T | undefined;
+export function getSafePropertyChain<SomeObject, Path extends ExtractPath<SomeObject>>(
+  obj: SomeObject,
+  propertyChain: Path
```
javieri-empathy, line 14: Isn't this missing the `defaultReturn` parameter? As now we have the nullish coalescing, I would prefer to remove it from all signatures.

packages/x-utils/src/__tests__/get-safe-property-chain.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-expect-error
+    const anTypeError: string | undefined = getSafePropertyChain(obj, 'aNumber');
+
+    expect(Array.isArray(anArray)).toBe(true);
+    expect(aNumber).toBe(1);
+    expect(aString).toBe('2');
+    expect(aBool).toBe(true);
+    expect(typeof anObject).toBe('object');
+    expect(anTypeError).toBe(1);
```
javieri-empathy, line 112: In order to assert that it properly returns the desired type, I would use stricter types:

- an string array instead of any array.
- a typed object instead of `object`. One of the issues with `object` is that functions can be assigned to this type, as they are "callable" objects.

```suggestion
   const obj = {
      anArray: ['value'],
      aNumber: 1,
      aString: '2',
      aBool: true,
      anObject: { child: 'string-child' }
    };
    const anArray: string[] | undefined = getSafePropertyChain(obj, 'anArray');
    const aNumber: number | undefined = getSafePropertyChain(obj, 'aNumber');
    const aString: string | undefined = getSafePropertyChain(obj, 'aString');
    const aBool: boolean | undefined = getSafePropertyChain(obj, 'aBool');
    const anObject: { child: string } | undefined = getSafePropertyChain(obj, 'anObject');
    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
    // @ts-expect-error
    const anTypeError: string | undefined = getSafePropertyChain(obj, 'aNumber');

    expect(anArray).toEqual(['value']);
    expect(aNumber).toBe(1);
    expect(aString).toBe('2');
    expect(aBool).toBe(true);
    expect(anObject).toEqual({ child: 'string-child' });
    expect(anTypeError).toBe(1);
```

javieri-empathy (changes requested):

packages/x-utils/src/__tests__/get-safe-property-chain.spec.ts
tajespasarela: 👍🏻

packages/x-utils/src/get-safe-property-chain.ts
tajespasarela: yes, I will do it when this PR is merged

https://docs.npmjs.com/deprecating-and-undeprecating-packages-or-package-versions

packages/x-adapter/package.json
tajespasarela: yes, I will do it when we merge this PR



https://docs.npmjs.com/deprecating-and-undeprecating-packages-or-package-versions

packages/x-utils/src/get-safe-property-chain.ts
tajespasarela: 👍🏻

packages/x-utils/src/get-safe-property-chain.ts
tajespasarela: 👍🏻

javieri-empathy (approved):

LuisMartinez15 (approved):
**praise:** good job!

PR #941, opened 2022-12-19
feat(desing-system): add input component to XDS

diegopf:
EX-7494

[Designs](https://www.figma.com/file/9a1DMgqIHt5ZZBvS2waaOs/X-DS?node-id=2391%3A30127&t=Tta6YAV5SbCWakqB-0)

### DIFF
--- a/packages/x-tailwindcss/demo/src/components/xds-input.vue
+++ b/packages/x-tailwindcss/demo/src/components/xds-input.vue
@@ -48,24 +48,13 @@
     @Prop({ default: () => ['x-input-line'] })
     public line!: string;
 
-    @Prop({
-      default: () => [
-        'x-input-lead x-input-sm',
-        'x-input-auxiliary x-input-circle x-input-outlined',
-        'x-input-accent x-input-tight ',
-        'x-input-warning x-input-ghost'
-      ]
-    })
-    public combinations!: string[];
-
     protected get sections(): ShowcaseSections {
       return {
         Default: [this.base],
         Colors: this.colors.map(addParentClasses(this.base)),
         Sizes: this.sizes.map(addParentClasses(this.base)),
         line: this.colors.map(addParentClasses(this.base, this.line)),
-        Disabled: [this.base, addParentClasses(this.base)(this.line)],
-        Combinations: this.combinations.map(addParentClasses(this.base))
+        Disabled: [this.base, addParentClasses(this.base)(this.line)]
       };
     }
   }
--- a/packages/x-tailwindcss/demo/src/components/xds-input.vue
+++ b/packages/x-tailwindcss/demo/src/components/xds-input.vue
@@ -45,11 +45,8 @@
     })
     public colors!: string[];
 
-    @Prop({ default: () => ['x-input-circle', 'x-input-square'] })
-    public layouts!: string[];
-
-    @Prop({ default: () => ['x-input-outlined'] })
-    public outline!: string;
+    @Prop({ default: () => ['x-input-line'] })
+    public line!: string;
 
     @Prop({
       default: () => [
@@ -66,9 +63,8 @@
         Default: [this.base],
         Colors: this.colors.map(addParentClasses(this.base)),
         Sizes: this.sizes.map(addParentClasses(this.base)),
-        Layout: this.layouts.map(addParentClasses(this.base)),
-        Outline: this.colors.map(addParentClasses(this.base, this.outline)),
-        Disabled: [this.base, addParentClasses(this.base)(this.outline)],
+        line: this.colors.map(addParentClasses(this.base, this.line)),
+        Disabled: [this.base, addParentClasses(this.base)(this.line)],
         Combinations: this.combinations.map(addParentClasses(this.base))
       };
     }
--- a/packages/x-tailwindcss/demo/src/components/xds-input.vue
+++ b/packages/x-tailwindcss/demo/src/components/xds-input.vue
@@ -1,0 +1,76 @@
+<template>
+  <XdsBaseShowcase
+    #default="{ cssClass, section, copyCssClassesToClipboard, removeClassPrefix }"
+    title="Button"
+    :sections="sections"
+  >
+    <input
+      :key="cssClass"
+      @click="copyCssClassesToClipboard"
+      :class="cssClass"
+      title="Click me to copy CSS classes"
+      :disabled="section === 'Disabled'"
+      :placeholder="removeClassPrefix(cssClass, base) || 'input'"
+    />
+  </XdsBaseShowcase>
+</template>
+
+<script lang="ts">
+  import { Vue, Component, Prop } from 'vue-property-decorator';
+  import { ShowcaseSections } from '../types/types';
+  import { addParentClasses } from '../utils';
+  import XdsBaseShowcase from './xds-base-showcase.vue';
+
+  @Component({
+    components: {
+      XdsBaseShowcase
+    }
+  })
+  export default class XdsInputShowcase extends Vue {
+    @Prop({ default: () => 'x-input' })
+    public base!: string;
+    @Prop({ default: () => ['x-input-sm', 'x-input-md', 'x-input-lg'] })
+    public sizes!: string[];
+    @Prop({
+      default: () => [
+        'x-input-neutral',
+        'x-input-lead',
+        'x-input-auxiliary',
+        'x-input-accent',
+        'x-input-highlight',
+        'x-input-success',
+        'x-input-warning',
+        'x-input-error'
+      ]
+    })
+    public colors!: string[];
+
+    @Prop({ default: () => ['x-input-circle', 'x-input-square'] })
+    public layouts!: string[];
+
+    @Prop({ default: () => ['x-input-outlined'] })
+    public outline!: string;
+
+    @Prop({
+      default: () => [
+        'x-input-lead x-input-sm',
+        'x-input-auxiliary x-input-circle x-input-outlined',
+        'x-input-accent x-input-tight ',
+        'x-input-warning x-input-ghost'
+      ]
+    })
+    public combinations!: string[];
+
+    protected get sections(): ShowcaseSections {
+      return {
+        Default: [this.base],
+        Colors: this.colors.map(addParentClasses(this.base)),
+        Sizes: this.sizes.map(addParentClasses(this.base)),
+        Layout: this.layouts.map(addParentClasses(this.base)),
+        Outline: this.colors.map(addParentClasses(this.base, this.outline)),
+        Disabled: [this.base, addParentClasses(this.base)(this.outline)],
+        Combinations: this.combinations.map(addParentClasses(this.base))
+      };
+    }
+  }
+</script>
--- a/packages/x-tailwindcss/demo/src/components/xds-input.vue
+++ b/packages/x-tailwindcss/demo/src/components/xds-input.vue
@@ -48,13 +48,24 @@
     @Prop({ default: () => ['x-input-line'] })
     public line!: string;
 
+    @Prop({
+      default: () => [
+        'x-input-lead x-input-sm',
+        'x-input-auxiliary x-input-line x-input-sm',
+        'x-input-accent x-input-lg ',
+        'x-input-warning x-input-line x-input-lg'
+      ]
+    })
+    public combinations!: string[];
+
     protected get sections(): ShowcaseSections {
       return {
         Default: [this.base],
         Colors: this.colors.map(addParentClasses(this.base)),
         Sizes: this.sizes.map(addParentClasses(this.base)),
         line: this.colors.map(addParentClasses(this.base, this.line)),
-        Disabled: [this.base, addParentClasses(this.base)(this.line)]
+        Disabled: [this.base, addParentClasses(this.base)(this.line)],
+        Combinations: this.combinations.map(addParentClasses(this.base))
       };
     }
   }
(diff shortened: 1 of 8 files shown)


packages/x-tailwindcss/src/x-tailwind-plugin/components/input/line.ts
```diff
@@ -0,0 +1,33 @@
...
+      '&:focus': {
+        borderColor: theme('colors.neutral.90'),
+        color: theme('colors.neutral.90'),
+        borderBottomWidth: theme('borderWidth.2')
+      },
+
+      '&:disabled': {
+        backgroundColor: theme('colors.neutral.10'),
+        borderColor: theme('colors.neutral.25')
+      }
```
tajespasarela, line 30: To avoid size sifting, we can use the background as bottom line:

```suggestion
      '--border-color': 'var(--input-color-50)',
      borderWidth: 0,
      backgroundImage: 'linear-gradient(var(--border-color) ,var(--border-color))',
      backgroundPosition: 'bottom',
      backgroundSize: '100% 1px',
      backgroundRepeat: 'no-repeat',

      '&:hover': {
        backgroundSize: '100% 2px'
      },

      '&:focus': {
        backgroundSize: '100% 2px',
        color: theme('colors.neutral.90')
      },

      '&:disabled': {
        backgroundColor: theme('colors.neutral.10'),
        '--border-color': theme('colors.neutral.25')
      }
```

packages/x-tailwindcss/src/x-tailwind-plugin/components/input/default.ts
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * @param helpers - The {@link TailwindHelpers} to generate CSS.
+ * @returns The {@link CssStyleOptions} for the component.
+ */
+// eslint-disable-next-line  @typescript-eslint/explicit-function-return-type
+export function inputDefault(helpers: TailwindHelpers) {
+  const { theme } = helpers;
+  return {
+    fontFamily: theme('fontFamily.main'),
+    borderRadius: 'none',
```
tajespasarela, line 16: `none` is not a valid value for border radius. Moreover by default it has `0` as border radius value, so, no need to set this value.
```suggestion
```

packages/x-tailwindcss/src/x-tailwind-plugin/components/input/default.ts
```diff
@@ -0,0 +1,36 @@
...
+    borderWidth: theme('borderWidth.1'),
+    cursor: 'pointer',
+    fontWeight: theme('fontWeight.regular'),
+    outline: 'none',
```
tajespasarela, line 20: remove this, as it will be necessary for accessibility. This will be disabled/enabled for the whole project:
```suggestion
```

packages/x-tailwindcss/src/x-tailwind-plugin/components/input/sizes.ts
```diff
@@ -0,0 +1,28 @@
...
+  return {
+    sm: {
+      minHeight: theme('spacing.32'),
+      fontSize: theme('fontSize.xs')
+    },
+    md: {
+      minHeight: theme('spacing.40'),
+      gap: theme('spacing.8'),
+      paddingInlineStart: theme('spacing.16'),
+      paddingInlineEnd: theme('spacing.16'),
```
tajespasarela, line 20: If these 3 properties are not the responsibility of the size variant, then move them to the default component.
```suggestion
```

packages/x-tailwindcss/demo/src/components/xds-input.vue
```diff
@@ -0,0 +1,61 @@
...
+        'x-input-success',
+        'x-input-warning',
+        'x-input-error'
+      ]
+    })
+    public colors!: string[];
+
+    @Prop({ default: () => ['x-input-line'] })
+    public line!: string;
+
```
tajespasarela, line 50: add some combinations examples

tajespasarela (changes requested):

tajespasarela (dismissed):

packages/x-tailwindcss/src/x-tailwind-plugin/components/input/default.ts
```diff
@@ -0,0 +1,37 @@
...
+  const { theme } = helpers;
+  return {
+    fontFamily: theme('fontFamily.main'),
+    borderWidth: theme('borderWidth.1'),
+    cursor: 'pointer',
+    fontWeight: theme('fontWeight.regular'),
+    overflow: 'hidden',
+    textOverflow: 'ellipsis',
+    whiteSpace: 'nowrap',
+    gap: theme('spacing.8'),
```
javieri-empathy, line 22: question: why do we need these styles? are we planning to apply the input classes to other elements?

javieri-empathy (changes requested):

packages/x-tailwindcss/src/x-tailwind-plugin/components/input/default.ts
diegopf: It is specified on the designs that we should do ellipsis if necessary

packages/x-tailwindcss/src/x-tailwind-plugin/components/input/index.ts
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Returns the component `input` CSS.
+ *
+ * @param helpers - The {@link TailwindHelpers} to generate CSS.
+ * @returns The {@link CssStyleOptions} for the component.
+ */
+// eslint-disable-next-line  @typescript-eslint/explicit-function-return-type
+export function input(helpers: TailwindHelpers) {
+  return {
+    '.input': deepMerge(
```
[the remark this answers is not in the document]
diegopf: To be honest I'm not sure either. I just took the `button` as an example and followed the same approach.

packages/x-tailwindcss/src/x-tailwind-plugin/components/input/sizes.ts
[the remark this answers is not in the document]
diegopf: We could but I'd rather have them separate, at least IMO.

packages/x-tailwindcss/src/x-tailwind-plugin/components/input/default.ts
javieri-empathy: Oh, I think I broke Chrome 😅. Yesterday this wasn't working for me, now it does. AFAIK this only works when you don't have the focus in the input. If I type a long query and I lose the focus by moving it into another screen it doesn't add the ellipsis.
image

 If I lose the focus by clicking on the page it does add it:
image

However, I still don't get the `gap` style.  The input does not have a flex or grid display.

javieri-empathy: image
Seems that `white-space` `overflow` and `gap` are not needed for that effect.

packages/x-tailwindcss/src/x-tailwind-plugin/components/input/default.ts
```diff
@@ -0,0 +1,37 @@
...
+      color: theme('colors.neutral.50'),
+      fontFamily: theme('fontFamily.main'),
+      fontSize: theme('fontSize.md'),
+      fontWeight: theme('fontWeight.regular')
+    },
+    '&:disabled': {
+      cursor: 'not-allowed'
+    },
+    ...inputColors(helpers).neutral,
+    ...inputSizes(helpers).md
```
javieri-empathy, line 35: question: I've seen this in more places and now I'm not sure about the approach. If we modify the `md` styles in a customer, they won't be inherited here. Not a big issue so far, but  maybe something confusing for the developers.

Also, by destructuring below the pseudoselectors, tailwind generates 3 `.x-input` rules.

packages/x-tailwindcss/src/x-tailwind-plugin/components/input/default.ts
javieri-empathy: Are they needed for the animated placeholder? Also this styles are conflicting with the base input and input group one, so we have to be careful when enabling them.

javieri-empathy (changes requested):

packages/x-tailwindcss/src/x-tailwind-plugin/components/input/index.ts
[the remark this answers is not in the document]
tajespasarela: depending if we want to merge nested objects or not.

packages/x-tailwindcss/src/x-tailwind-plugin/components/input/default.ts
```diff
@@ -0,0 +1,37 @@
...
+    overflow: 'hidden',
+    textOverflow: 'ellipsis',
+    whiteSpace: 'nowrap',
+    gap: theme('spacing.8'),
+    paddingInlineStart: theme('spacing.16'),
+    paddingInlineEnd: theme('spacing.16'),
+    '&::placeholder': {
+      color: theme('colors.neutral.50'),
+      fontFamily: theme('fontFamily.main'),
+      fontSize: theme('fontSize.md'),
```
tajespasarela, line 28: I think the font size of the placeholder should change with sizes too, so let's move to there

packages/x-tailwindcss/src/x-tailwind-plugin/components/input/default.ts
tajespasarela: I think the `gap` is not needed as the input would not have multiple children.  The `white-space` & `overflow` are not needed because it is an `input` with that implicit behaviour. So we can remove them

Surely we must do some more work to make this work with the animated placeholder. That's one of the reasons I preferred to use the proper placeholder to be animated instead of another element.

I think we have to use the `input group` to use the animated placeholder, so I would leave that case for that task.


@javieri-empathy I don't understand what you mean by "conflicting with the base input and input group"?
do you mean the old DS? those will be deprecated once we use the new ones.

packages/x-tailwindcss/src/x-tailwind-plugin/components/input/default.ts
tajespasarela: That's on purpose, we don't want to modify the default `input` when we modify de `md` size. We are just reusing code. But in a setup, we want to modify the default and the sizes separately. It happens the same with the color variant. The default color of the component is the same color as the neutral variant. But I want to modify them separately in the setup.

We will see how to make it easier to override these default styles from setups., and surely we have to modify something, but at the moment we don't have all the information.

packages/x-tailwindcss/src/x-tailwind-plugin/components/input/default.ts
tajespasarela, line 28: I think the font size of the placeholder should change with sizes too, so let's move to there

tajespasarela (changes requested):

packages/x-tailwindcss/src/x-tailwind-plugin/components/input/default.ts
javieri-empathy: That was it, just wanted to confirm that this was the intention of this PR. Thanks.

packages/x-tailwindcss/src/x-tailwind-plugin/components/input/default.ts
javieri-empathy: Understood, thanks.

packages/x-tailwindcss/src/x-tailwind-plugin/components/input/default.ts
tiborux: Yes you are right, we can remove those properties. The thing is that you can not have an ellipsis on the input if you have the focus, I think that this is intended to read everything when you have the focus (at least in In Webkit-based browsers). I don't know if there is a hack to add an ellipsis always, but I think that it makes no sense.

https://jsfiddle.net/simurai/TjGyz/

packages/x-tailwindcss/src/x-tailwind-plugin/components/input/default.ts
```diff
@@ -16,22 +17,22 @@ export function inputDefault(helpers: TailwindHelpers) {
...
+    ...deepMerge(
+      { ...inputSizes(helpers).md },
+      {
+        '&::placeholder': {
+          color: theme('colors.neutral.50'),
+          fontFamily: theme('fontFamily.main'),
+          fontWeight: theme('fontWeight.regular')
+        }
+      }
+    )
```
tajespasarela, line 36: maybe it is more readable to move the `deepMerge` to the root:

```javascript
  return deepMerge(
    {
      fontFamily: theme('fontFamily.main'),
      borderWidth: theme('borderWidth.1'),
      cursor: 'text',
      fontWeight: theme('fontWeight.regular'),
      textOverflow: 'ellipsis',
      paddingInlineStart: theme('spacing.16'),
      paddingInlineEnd: theme('spacing.16'),
      '&:disabled': {
        cursor: 'not-allowed'
      },
      '&::placeholder': {
        color: theme('colors.neutral.50'),
        fontFamily: theme('fontFamily.main'),
        fontWeight: theme('fontWeight.regular')
      }
    },
    inputColors(helpers).neutral,
    inputSizes(helpers).md
  );
```

tajespasarela (changes requested):

tajespasarela (approved):

javieri-empathy (approved):

PR #943, opened 2022-12-21
feat(design-system): Add suggestion group component

tiborux:
EX-7496


packages/x-tailwindcss/src/x-tailwind-plugin/components/suggestion-group/utils/overrides-children.ts
```diff
@@ -0,0 +1,30 @@
+import { TailwindHelpers } from '../../../../types';
+
+/**
+ * Overrides the `suggestion` styles when there is inside a `suggestion group`.
```
tiborux, line 4:
```suggestion
 * Overrides the `suggestion` styles when it is inside a `suggestion group`.
```

packages/x-tailwindcss/src/x-tailwind-plugin/components/suggestion-group/button/layout.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Returns the `layout` variant for component `suggestion group button`.
+ *
+ * @param helpers - The {@link TailwindHelpers} to generate CSS.
+ * @returns The {@link CssStyleOptions} for the variant.
+ */
+// eslint-disable-next-line  @typescript-eslint/explicit-function-return-type
+export function suggestionGroupButtonLayout(helpers: TailwindHelpers) {
+  const { theme } = helpers;
+  return {
+    layout: {
```
tiborux, line 13: Can anyone find a better name? `suggestion-group-button-no-ratio`?

packages/x-tailwindcss/src/x-tailwind-plugin/components/suggestion-group/button/default.ts
```diff
@@ -0,0 +1,24 @@
...
+    alignItems: 'center',
+    alignContent: 'center',
+    boxSizing: 'border-box',
+    backgroundColor: 'transparent',
+    cursor: 'pointer',
+    color: 'currentColor',
+    minHeight: 'inherit',
+    aspectRatio: '1',
+    '&:hover': {
+      transform: 'scale(1.2)'
```
annacv, line 21: Just in case: in XDS Figma the scale default value is set to `1.4`.

packages/x-tailwindcss/src/x-tailwind-plugin/components/suggestion-group/button/layout.ts
annacv: I've found something here: https://math.stackexchange.com/questions/2156892/what-is-the-name-for-a-rectangle-that-is-not-a-square ---> Based on this it could be named "Oblong" but not sure if everyone would find it meaningful

packages/x-tailwindcss/src/x-tailwind-plugin/components/suggestion-group/utils/overrides-children.ts
```diff
@@ -0,0 +1,33 @@
...
+      color: 'currentColor',
+      minHeight: 'inherit',
+      fontSize: 'inherit',
+      fontFamily: 'inherit',
+      fontWeight: 'inherit'
+    },
+
+    '&:hover': {
+      '.suggestion': {
+        textDecoration: 'underline'
```
javieri-empathy, line 29: As per the designs, in `ghost` and `outline`, the underline should not be applied
image
image

javieri-empathy (changes requested):

packages/x-tailwindcss/src/x-tailwind-plugin/components/suggestion-group/button/default.ts
diegopf: We decided to reduce the default scale and it seems that the figma doc was not updated. It is correct the 1.2.

packages/x-tailwindcss/src/x-tailwind-plugin/components/suggestion-group/button/default.ts
```diff
@@ -0,0 +1,31 @@
+/**
+ * Returns the default styles for component `suggestion group button`.
+ *
+ * @returns The {@link CssStyleOptions} for the component.
+ */
+// eslint-disable-next-line  @typescript-eslint/explicit-function-return-type
+export function suggestionGroupButtonDefault() {
+  return {
+    display: 'grid',
+    gridAutoFlow: 'column',
```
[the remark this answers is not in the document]
diegopf: I think the reason to go with grid is to ease the alignment with the suggestion. The suggestion is already using grid for the layout with the curated icon and I think it is just easier to follow the same approach with the suggestion group button even though is only one element.

packages/x-tailwindcss/src/x-tailwind-plugin/components/suggestion-group/button/default.ts
```diff
@@ -0,0 +1,31 @@
...
+    cursor: 'pointer',
+    color: 'currentColor',
+    minHeight: 'inherit',
+    aspectRatio: '1',
+    '&:hover': {
+      transform: 'scale(1.2)'
+    },
+
+    /*To fix icon baseline alignment needs text to know the baseline*/
+    '&::after': {
```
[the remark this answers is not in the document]
diegopf: Aligning the suggestion-group-button when it was only an icon was problematic until Iván came up with this hack. I don't know if there's a better solution to be honest 😅

packages/x-components/CHANGELOG.md
```diff
@@ -1,25 +1,22 @@
 # Change Log
 
-All notable changes to this project will be documented in this file.
-See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
+All notable changes to this project will be documented in this file. See
```
javieri-empathy, line 3: question: did prettier break?

packages/x-tailwindcss/src/x-tailwind-plugin/components/suggestion-group/utils/overrides-children.ts
```diff
@@ -0,0 +1,33 @@
...
+      fontFamily: 'inherit',
+      fontWeight: 'inherit'
+    },
+
+    '&.suggestion-group-ghost:hover, &.suggestion-group-outlined:hover': {
+      '.suggestion': {
+        textDecoration: 'none'
+      }
+    }
+  };
```
javieri-empathy, line 32: suggestion: Wouldn't make more sense to have this defined in the `ghost` and `outlined` variant files? As I told you before I believe it will be harder to maintain if we start to place styles for the same variant in different files rather than in the same file. Moreover these styles would also have less specifity.

packages/x-tailwindcss/src/x-tailwind-plugin/components/suggestion-group/outlined.ts
```diff
@@ -0,0 +1,23 @@
...
+ * @returns The {@link CssStyleOptions} for the variant.
+ */
+// eslint-disable-next-line  @typescript-eslint/explicit-function-return-type
+export function suggestionGroupOutlined(helpers: TailwindHelpers) {
+  const { theme } = helpers;
+  return {
+    outlined: {
+      backgroundColor: theme('colors.neutral.0'),
+      borderWidth: theme('borderWidth.1'),
+      borderColor: 'var(--suggestion-group-color-50)',
```
javieri-empathy, line 16: issue: This variable does not have to be defined unless you use one of the color classes. Should we add a fallback value from the theme like the figma file says with `neutral-90`?

image

packages/x-tailwindcss/src/x-tailwind-plugin/components/suggestion-group/outlined.ts
```diff
@@ -0,0 +1,23 @@
...
+      '&:hover': {
+        backgroundColor: theme('colors.neutral.10')
+      },
```
javieri-empathy, line 19: suggestion: When you add sub-selectors like this one in the middle of more CSS rules, tailwind generates 2 different rules. Moreover, I believe it is easier to read if we group all the styles together before the nested selectors.
image

packages/x-tailwindcss/src/x-tailwind-plugin/components/suggestion-group/outlined.ts
```diff
@@ -0,0 +1,23 @@
...
+      gap: theme('spacing.4')
+    }
```
javieri-empathy, line 21: issue: arren't outlined and ghost variants missing the padding left/right styles depending on their size? 
image

javieri-empathy (changes requested):

packages/x-tailwindcss/src/x-tailwind-plugin/components/suggestion-group/outlined.ts
diegopf: yes, you are right! Besides, I noticed that the border color should point to `--suggestion-group-color-75` and not `--suggestion-group-color-50`

packages/x-tailwindcss/src/x-tailwind-plugin/components/suggestion-group/outlined.ts
diegopf: you are right again, good catch!

packages/x-tailwindcss/src/x-tailwind-plugin/components/suggestion-group/utils/overrides-children.ts
diegopf: I just got rid of this util when I adjusted the margin by size variant

packages/x-components/CHANGELOG.md
diegopf: I'm not sure what happened here, seems the formatting was wrong somehow

packages/x-tailwindcss/src/x-tailwind-plugin/components/suggestion-group/button/layout.ts
diegopf: I've renamed it to rectangle

javieri-empathy (approved):

annacv (approved):
🚀

PR #960, opened 2022-12-27
fix: mutable schema types

javieri-empathy:
BREAKING CHANGE: `MutableSchema` now receives a `Source` and a `Target` object instead of a `Schema`
BREAKING CHANGE: `$replace`, `$extends` and `$override` function types now reflect the allowed operations more correctly.

EX-7702

Modifies `MutableSchema` types so overriding, extending, and replacing function types reflect correctly the operations made.

- `$replace` should replace **completely** the old schema. That's why in the implementation there are some casts, as this operation is not modelable with TypeScript.
- `$override` should **mutate** the current schema, allowing to add new properties or modifying mappings. Therefore  you are allowed to pass a new `Source` and `Target` generic types that will be combined with the old ones.
- `$extends` works pretty much the same than `$override` with the exception that it does **not** mutate the current schema, it creates a new one.

Additionally `MutableSchema` type has been modified to allow easier mapping this operations types, by receiving the `Source` and `Target` types directly rather than a `Schema`.

```ts
const base = createMutableSchema<{ _name: string; _id: string; _label: string }, { title: string }>(
  {
    title: '_name'
  }
);

schemaMapperFactory<{ _name: string; _id: string; _label: string }, { title: string }>(base);

const new1 = base.$override<{ _new: string }, { id: string }>({ title: '_label', id: '_id' });
const new2 = new1.$extends<{ _a: unknown }, { a?: unknown }>({ title: '_new' });
const new3 = new2.$extends<{ _image: string }, { images: string[] }>({ images: a => [a._image] });
```

### DIFF
--- a/packages/x-adapter/src/schemas/types.ts
+++ b/packages/x-adapter/src/schemas/types.ts
@@ -1,7 +1,6 @@
 import {
   AnyFunction,
   DeepPartial,
-  EmptyObject,
   ExtractPath,
   ExtractPathByType,
   ExtractType,
@@ -62,7 +61,7 @@ export type MutableSchema<Source, Target> = Schema<Source, Target> & {
    * @param newSchema - The {@link Schema | schema} to use instead of the original one.
    * @returns The new {@link Schema | schema} that will be used.
    */
-  $replace<NewSource, NewTarget = EmptyObject>(
+  $replace<NewSource, NewTarget>(
     newSchema: Schema<NewSource, NewTarget>
   ): MutableSchema<NewSource, NewTarget>;
   /**
--- a/packages/x-adapter/src/schemas/types.ts
+++ b/packages/x-adapter/src/schemas/types.ts
@@ -1,5 +1,7 @@
 import {
   AnyFunction,
+  DeepPartial,
+  EmptyObject,
   ExtractPath,
   ExtractPathByType,
   ExtractType,
@@ -60,7 +62,7 @@ export type MutableSchema<Source, Target> = Schema<Source, Target> & {
    * @param newSchema - The {@link Schema | schema} to use instead of the original one.
    * @returns The new {@link Schema | schema} that will be used.
    */
-  $replace<NewSource, NewTarget>(
+  $replace<NewSource, NewTarget = EmptyObject>(
     newSchema: Schema<NewSource, NewTarget>
   ): MutableSchema<NewSource, NewTarget>;
   /**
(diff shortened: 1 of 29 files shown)


packages/x-adapter/src/schemas/utils.ts
```diff
@@ -16,32 +16,30 @@ const mutableSchemasInternalMethods = ['$replace', '$override', '$extends', 'toS
...
   return {
     ...schema,
-    $replace: function <Source = any, Target = any>(
-      newSchema: Schema<Source, Target>
-    ): MutableSchema<Schema<Source, Target>> {
-      Object.keys(this).forEach(key => {
-        if (isInternalMethod(key)) {
+    $replace(newSchema) {
+      forEach(this, key => {
+        if (isInternalMethod(key as string)) {
```
javieri-empathy, line 26: I'm willing to remove `isInternalMethod` using either a class, defining these methods as non iterable, or using a custom iterator function.

packages/x-adapter/src/schemas/utils.ts
```diff
@@ -16,32 +16,30 @@ const mutableSchemasInternalMethods = ['$replace', '$override', '$extends', 'toS
...
+  schema: Schema<Source, Target>
+): MutableSchema<Source, Target> {
   return {
     ...schema,
-    $replace: function <Source = any, Target = any>(
-      newSchema: Schema<Source, Target>
-    ): MutableSchema<Schema<Source, Target>> {
-      Object.keys(this).forEach(key => {
-        if (isInternalMethod(key)) {
+    $replace(newSchema) {
```
javieri-empathy, line 24: When working with objects and interfaces we don't have to define the types again unless we need them. They are correctly inferred.

packages/x-adapter/src/schemas/types.ts
```diff
@@ -53,45 +54,48 @@ export type Schema<Source = any, Target = any> = {
...
   /**
    * Replaces all usages of the original {@link Schema | schema} with the given one.
    *
    * @param newSchema - The {@link Schema | schema} to use instead of the original one.
    * @returns The new {@link Schema | schema} that will be used.
    */
-  $replace: <Source, Target>(
-    newSchema: Schema<Source, Target>
-  ) => MutableSchema<Schema<Source, Target>>;
+  $replace<NewSource, NewTarget>(
```
javieri-empathy, line 64: By changing to this signature, the `this` type is correctly inferred, and we don't allow to destructure the functions.

annacv (dismissed):
👀  seems ok to me

LuisMartinez15 (approved):
**praise**: well done!

annacv (approved):
🚀

PR #961, opened 2022-12-28
feature/EX-7706-fix-whitespace-rollup-vue

javieri-empathy:
EX-7706

Sets rollup-plugin-vue `whitespace` option to `'condense'` to match the [Vue CLI defaults](https://cli.vuejs.org/migrations/migrate-from-v3.html#vue-cli-service). This should prevent empty text nodes from being generated in the output, and prevent us from having different behaviors between dev and production builds.

image

### DIFF
--- a/packages/x-components/build/rollup.config.ts
+++ b/packages/x-components/build/rollup.config.ts
@@ -87,6 +87,11 @@ export const rollupConfig = createRollupOptions({
       // lang is set to ts:
       // https://github.com/vuejs/rollup-plugin-vue/issues/272#issuecomment-491721842
       needMap: false,
+      template: {
+        compilerOptions: {
+          whitespace: 'condense'
+        }
+      } as any,
       style: {
         postcssCleanOptions: { disabled: true }
       }
--- a/packages/x-components/build/rollup.config.ts
+++ b/packages/x-components/build/rollup.config.ts
@@ -7,7 +7,7 @@ import del from 'rollup-plugin-delete';
 import rename from 'rollup-plugin-rename';
 import styles from 'rollup-plugin-styles';
 import typescript from 'rollup-plugin-typescript2';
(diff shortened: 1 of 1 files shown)


packages/x-components/build/rollup.config.ts
```diff
@@ -87,6 +87,11 @@ export const rollupConfig = createRollupOptions({
       // lang is set to ts:
       // https://github.com/vuejs/rollup-plugin-vue/issues/272#issuecomment-491721842
       needMap: false,
+      template: {
+        compilerOptions: {
+          whitespace: 'condense'
+        }
+      } as VuePluginOptions['template'],
```
joseacabaneros, line 94:
```typescript
vue({
    css: false,
    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
    // @ts-ignore Undocumented option to disable vue sourcemap generation because it breaks if
    // lang is set to ts:
    // https://github.com/vuejs/rollup-plugin-vue/issues/272#issuecomment-491721842
    needMap: false,
    template: {
      compilerOptions: {
        whitespace: 'condense'
      }
    },
    style: {
      postcssCleanOptions: { disabled: true }
    }
  } VuePluginOptions)
```

WDYT? 🤔

joseacabaneros (approved):

packages/x-components/build/rollup.config.ts
LuisMartinez15: That change turns the `template.compiler` property to be required. I don't know why with @javieri-empathy approach it doesn't require `template.compiler` to be defined, as he is using the very same type. Also, the type is already properly inferred, there's no need to type assert it. I believe @javieri-empathy did it to avoid the issue I was talking before.

LuisMartinez15 (approved):
**praise**: good research!

PR #970, opened 2023-01-02
Feature/ex 7721 fix cache ci

javieri-empathy:
EX-7721

Previously Cypress binary was cached with a static key. This caused Cypress updates to break the CI cache, as the project cached Cypress version did not have a proper Cypress binary cached as well.

image

Now by combining both caches using the lock file hash, we should be safe. Also as we shouldn't update the package-lock in every PR, we should not see any performance drop in most of the PRs, only the ones that update the package-lock.

Also to clear the previous cache, I've updated the minor cypress version.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,7 +37,6 @@ jobs:
           CI: 1
         # print Cypress and OS info
         run: |
-          npm ci
           npx cypress verify
           npx cypress info
           npx cypress version
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,27 +23,10 @@ jobs:
         id: cache-node-modules
         uses: actions/cache@v3
         with:
-          path: '**/node_modules'
+          path: |
+           **/node_modules
+            ~/.cache/Cypress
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yml
```diff
@@ -23,14 +23,10 @@ jobs:
...
+           **/node_modules
+            ~/.cache/Cypress
           key: node-modules-${{ hashFiles('**/package-lock.json') }}
-      - name: Cache Cypress
-        id: cache-cypress
-        uses: actions/cache@v3
-        with:
-          path: ~/.cache/Cypress
-          key: cypress-cache
       - name: Cache eslint
```
ajperezbau, line 30: Could you add a `eslint-` restore key?
It is needed for the action to take the previous commit cache, use it and create a new one with the current comment `sha`.

```
      - name: Cache eslint
        id: cache-eslint
        uses: actions/cache@v3
        with:
          path: .eslintcache
          key: eslint-${{github.sha}}
          restore-keys: eslint-
```

ajperezbau (commented):
Nice 👌

mnavarroespinosa (approved):
looks good to me!

ajperezbau (approved):

PR #971, opened 2023-01-03
feat: update `vue` to version `2.7`

javieri-empathy:
EX-7730

Updates to Vue 2.7

We should test this branch with:

Archetype. So far it works for me. I haven't seen any weird behaviors, warnings, or errors.
Motive
Empathy-components


packages/react-wrapper/package.json
```diff
@@ -42,7 +42,7 @@
   "peerDependencies": {
     "react": "^16.9.0",
     "react-dom": "^16.9.0",
-    "vue": "~2.6.0"
+    "vue": "^2.6.0"
```
javieri-empathy, line 45: As the code does not change, we can still keep the 2.6 version as the oldest we support with react-wrapper

packages/react-wrapper/src/__tests__/tsconfig.json
```diff
@@ -4,5 +4,5 @@
     "types": ["jest"]
   },
   "include": ["**/*.ts", "**/*.tsx"],
-  "exclude": []
+  "exclude": ["node_modules"]
```
javieri-empathy, line 7: Every time we add an exclude property we should add node_modules

packages/search-types/CHANGELOG.md
```diff
@@ -1,24 +1,16 @@
 # Change Log
 
-All notable changes to this project will be documented in this file.
-See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
+All notable changes to this project will be documented in this file. See
```
tajespasarela, line 3: al the changelog changes are done by prettier or by hand?

packages/x-tailwindcss/package.json
```diff
@@ -68,10 +68,10 @@
     "typescript": "~4.6.2",
     "vite": "~3.2.0",
     "vite-plugin-vue2": "~2.0.2",
-    "vue": "~2.6.14",
+    "vue": "~2.7.14",
```
tajespasarela, line 71: in the moment we start to use the composition API, we have to change the peer dependency to `2.7.x`?

why don't have it already changed?

tajespasarela (changes requested):

packages/x-components/src/shims-tsx.d.ts
```diff
@@ -4,8 +4,5 @@ declare global {
   namespace JSX {
     interface Element extends VNode {}
     interface ElementClass extends Vue {}
```
javieri-empathy, line 6: We don't use JSX but this is already a record type in the base JSX definition
I think we can totally get rid of this file. Improting from Vue will now load all the jsx custom types, which is the issue why it conflicts with React.

packages/x-components/src/components/layouts/fixed-header-and-asides-layout.vue
```diff
@@ -108,8 +109,8 @@
   })
   export default class FixedHeaderAndAsidesLayout extends mixins(LayoutsMixin) {
```
javieri-empathy, line 110: I don't know why, if we let these two properties type to be inferred, it adds a MainScroll type there, which is incorrect and very strange.
image

image

packages/search-types/CHANGELOG.md
javieri-empathy: These ones (`-`) are done by Prettier. The ones with `*` are generated by lerna. I don't know why some times the changelog is formatted and some times not. I've created a task to dig deeper into it.

tajespasarela (approved):

LuisMartinez15 (approved):
**praise**: good job!

PR #990, opened 2023-01-09
feat: add `x-priority-bus`

LuisMartinez15:
EX-7458


packages/x-bus/src/x-bus.types.ts
```diff
@@ -0,0 +1,144 @@
...
+/**
+ * Alias representing a positive number to represent the priority of an {@link XEvent} in an
+ * {@link XPriorityBus}.
+ *
+ * @internal
+ */
+export type Priority = number;
+
+/**
+ * Alias representing a dictionary where the key is the name of the {@link XEvent} and the type
```
tiborux, line 22: This description is not right if you are using a default dictionary, because the key can be anything not a XEvent.

packages/x-bus/src/x-bus.types.ts
```diff
@@ -0,0 +1,144 @@
...
+  ? undefined
+  : XEvents[Event];
+
+/**
+ * Represents the payload of a subject with a given type.
+ *
+ * @internal
+ */
+export type SubjectPayload<SomePayload> = {
+  /** The payload of the event. */
```
tiborux, line 49:
```suggestion
  /** 
  * The payload of the event. 
  */
```

packages/x-bus/src/x-bus.types.ts
```diff
@@ -0,0 +1,144 @@
...
+  eventPayload: SomePayload;
+  /** Extra data of the event. */
```
tiborux, line 51:
```suggestion
  /** 
   * Extra data of the event. 
   */
```

packages/x-bus/src/x-bus.ts
```diff
@@ -0,0 +1,245 @@
...
+   * @internal
+   */
+  protected priorities: Dictionary<number>;
+
+  /**
+   *
+   *
+   * @internal
+   */
+  protected emitCallbacks: AnyFunction[];
```
tiborux, line 51: Missing documentation

packages/x-bus/src/x-bus.ts
```diff
@@ -0,0 +1,245 @@
...
+  /**
+   * Creates a new instance of a {@link XPriorityBus}.
+   *
+   * @param config
+   * @param queue - A {@link XPriorityQueue} to store the events.
+   * @param priorities - A {@link @empathyco/x-utils#Dictionary} defining the priorities associated
+   * to a given string.
+   */
+  public constructor(
+    config: {
```
tiborux, line 84: Should we create a type for the config?

packages/x-bus/src/x-bus.ts
```diff
@@ -0,0 +1,245 @@
...
+   * A list of pending pop operations {@link TimeoutId | timeout identifiers}.
+   *
+   * @internal
+   */
+  protected pendingPops: TimeoutId[] = [];
+
+  /**
+   * Creates a new instance of a {@link XPriorityBus}.
+   *
+   * @param config
```
tiborux, line 78: Add details about what config is

packages/x-bus/src/x-bus.ts
```diff
@@ -0,0 +1,245 @@
...
+   * - the max positive value is assigned
+   *
+   * @param event - The {@link XEvent | event} to get the priority from.
+   * @param metadata - The {@link XEvent | event} metadata.
+   * @returns The priority for the given {@link XEvent | event}.
+   *
+   * @internal
+   */
+  getEventPriority(event: XEvent, metadata: Dictionary): Priority {
+    const [, priorityKey] = Object.keys(this.priorities).reduce(
```
tiborux, line 143: Because of how this code is made, we are deciding which priority is more refined using the length of the matched part. For example, if you emit an event 'UserRequestChanged' if you don't compare which priority is more refined, the priority will be 
 `Changed` instead of `RequestChanged`
 
After a discussion, we decided to not do this. It doesn't make sense to check if a 'matched' key is more refined than another one, to avoid that, we should change the name of the `RequestChanged`events to   `RequestUpdated`  or create a more restricted regex. In this case, we should rename the events to be `RequestUpdated`

Apart from this:

packages/x-bus/src/x-bus.ts
```diff
@@ -0,0 +1,245 @@
...
+            metadata: eventMetadata
+          };
+
+          emitter.next(payloadObj);
+
+          this.emitCallbacks.forEach(callback => callback(key, payloadObj));
+          resolve({ event: key, ...payloadObj });
+
+          this.pendingPops = this.pendingPops.filter(timeoutId => timeoutId !== popTimeoutId);
+        }) as unknown as number;
```
tiborux, line 187: Remove this cast, it is not necessary.

packages/x-bus/src/x-bus.ts
```diff
@@ -0,0 +1,245 @@
...
+
+        this.pendingPops.push(popTimeoutId);
+      }
+    }) as unknown as number;
```
tiborux, line 191: Remove this cast, it is not necessary.

packages/x-bus/src/x-bus.types.ts
ajperezbau: +1

packages/x-bus/src/x-bus.types.ts
ajperezbau: +1

packages/x-bus/src/x-bus.ts
```diff
@@ -0,0 +1,245 @@
...
+   *
+   * @param config
+   * @param queue - A {@link XPriorityQueue} to store the events.
+   * @param priorities - A {@link @empathyco/x-utils#Dictionary} defining the priorities associated
+   * to a given string.
+   */
+  public constructor(
+    config: {
+      queue?: XPriorityQueue;
+      priorities?: Dictionary<number>;
```
ajperezbau, line 86: Should we use here the `Priority` type?
```suggestion
      priorities?: Dictionary<Priority>;
```

ajperezbau (changes requested):
¡Good job! 👌

packages/x-bus/src/x-bus.types.ts
tiborux: Yep, should be 'SomeEvent' or something like that

support-empathy:
 [Check out EPDocs preview](https://docs-dev.empathy.co/preview/x-990/index.html)

support-empathy:
 [Check out EPDocs preview](https://docs-dev.empathy.co/preview/x-990/index.html)

packages/x-bus/src/x-bus.types.ts
LuisMartinez15: Renamed

packages/x-bus/src/x-bus.ts
LuisMartinez15: I don't think so. It isn't something we are going to reuse anywhere else.

packages/x-bus/src/x-bus.ts
LuisMartinez15: Refactored based on discussion

packages/x-bus/src/x-bus.ts
```diff
@@ -0,0 +1,245 @@
...
+  }
+}
+
+/**
+ * A {@link XBus} implementation using an {@link XPriorityQueue} as its data structure to
+ * prioritise the emission of events. The priorities are preconfigured based on event naming.
+ *
+ * @public
+ */
+export class XPriorityBus implements XBus {
```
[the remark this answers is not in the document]
LuisMartinez15: Done

packages/x-bus/src/x-bus.ts
LuisMartinez15: Good eye, done!

packages/x-bus/src/x-bus.ts
```diff
@@ -0,0 +1,245 @@
...
+   */
+  public constructor(
+    config: {
+      queue?: XPriorityQueue;
+      priorities?: Dictionary<number>;
+      emitCallbacks?: AnyFunction[];
+    } = {}
+  ) {
+    this.queue = config.queue ?? new XPriorityQueue();
+    this.priorities = config.priorities ?? {
```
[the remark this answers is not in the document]
LuisMartinez15: Agree, I've removed them.

packages/x-bus/src/x-bus.ts
```diff
@@ -0,0 +1,245 @@
...
+    const [, priorityKey] = Object.keys(this.priorities).reduce(
+      ([matchedCount, matchedName], name) => {
+        const count = event.match(new RegExp(name))?.[0].length ?? 0;
+
+        return count > matchedCount ? [count, name] : [matchedCount, matchedName];
+      },
+      [0, ''] as [count: number, priorityKey: string]
+    );
+
+    return metadata.priority ?? this.priorities[priorityKey] ?? Number.MAX_VALUE;
```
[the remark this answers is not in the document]
LuisMartinez15: Refactored

packages/x-bus/src/__tests__/x-bus.spec.ts
```diff
@@ -1,7 +1,191 @@
...
+    });
+  });
+
+  describe('getEventPriority', () => {
+    it('can override the priority behavior', async () => {
+      // Given a priority bus
+      const bus = new XPriorityBus();
+
+      // When the priority assignment function is overriden
+      bus.getEventPriority = (event, metadata) => metadata.customPriority;
```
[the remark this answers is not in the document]
LuisMartinez15: I have mixed feelings about this. Lets wait for the other reviewer thoughts on this.

packages/x-bus/src/__tests__/x-bus.spec.ts
```diff
@@ -1,7 +1,191 @@
...
+      // Given a priority bus
+      const queue = new XPriorityQueue();
+      const bus = new XPriorityBus({ queue });
+      const emittedEvents: XEvent[] = [];
+      const pushEmittedEvent: AnyFunction = ({ event }) => emittedEvents.push(event);
+
+      // When multiple events are emitted at the same time
+      // Then they're added to the queue
+      // And cancel an ongoing flushing of the queue
+      let pendingFlush = bus['pendingFlush'];
```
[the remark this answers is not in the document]
LuisMartinez15: As we discussed privately, I think it doesn't "harm" to also check the value of this property, even though it is private, as we are in a test. Let's see what the other reviewer thinks.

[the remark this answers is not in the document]
ajperezbau: Actually I think that using that variable can make the test more fragile.
If for whatever reason we modify that variable in the source code, for example, to rename it or change its type, we will also need to update the test, even if the result is the same as before.
Yep, let's see what @tiborux thinks.

ajperezbau (commented):
Awesome 👌

packages/x-bus/src/x-bus.ts
```diff
@@ -0,0 +1,274 @@
...
+} from './x-bus.types';
+
+/**
+ * A default {@link XBus} implementation using a
+ * {@link @empathyco/x-priority-queue#XPriorityQueue | priority queue} as its data structure to
+ * prioritise the emission of events. The priorities are preconfigured based on event naming.
+ *
+ * @public
+ */
+export class XPriorityBus<SomeEvents extends Dictionary, SomeEventMetadata extends Dictionary>
```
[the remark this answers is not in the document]
annacv: And maybe set the `replaceable` as `readonly`. Does it make sense?

packages/x-bus/src/__tests__/x-bus.spec.ts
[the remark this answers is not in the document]
LuisMartinez15: Well, if we change anything related to that method, we will 99% need to update the test. I know testing private implementation details is not the norm and should be avoided, but in this case it provides a better insight and offers a better reflection of what is expected when running the code.

tiborux (dismissed):

packages/x-bus/src/__tests__/x-bus.spec.ts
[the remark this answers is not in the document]
LuisMartinez15: At the end we decided to remove the private property checks.

packages/x-bus/src/__tests__/x-bus.spec.ts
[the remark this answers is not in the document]
LuisMartinez15: I've changed the visibility of the function to protected and removed the test. We decided that it was a little messy and in case you want a custom getEventPriority function you can always create your own bus implementation or we can later in the future expose this functionality. Additionally, a new property has been created, `defaultEventPriority`, to set a default priority for any event that doesn't have a priority defined in its metadata nor its name matches any key of the priorities dictionary.

packages/x-bus/src/x-bus.ts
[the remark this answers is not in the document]
LuisMartinez15: As we discussed we are not going to set a default type for the metadata for the moment. We might do it later if we see it necessary in the future.

tiborux (approved):

ajperezbau (approved):

PR #999, opened 2023-01-11
fix: remove extra '>', fix search button

CachedaCodes:
EX-7809

* Remove an extra `>`
* Replace the search icon for the one used in the archetype
* Fix the button and it's background color

### DIFF
--- a/packages/x-archetype-utils/src/home/home-template.html
+++ b/packages/x-archetype-utils/src/home/home-template.html
@@ -87,20 +87,16 @@
         />
         <button class="x-archetype-button x-archetype-button--square" onclick="InterfaceX.search()">
           <svg
-            class="x-archetype-logo"
-            width="20"
-            height="20"
-            viewBox="0 0 20 20"
-            fill="none"
+            class="x-archetype-button-icon"
+            viewBox="0 0 16 16"
+            version="1.1"
             xmlns="http://www.w3.org/2000/svg"
+            xmlns:xlink="http://www.w3.org/1999/xlink"
           >
-            <path
-              d="M19.1336 18.571L14.0318 13.9273M15.6897 8.83908C15.6897 12.9315 12.3714 16.2491 8.27798 16.2491C4.18457 16.2491 0.866211 12.9315 0.866211 8.83908C0.866211 4.74665 4.18457 1.42908 8.27798 1.42908C12.3714 1.42908 15.6897 4.74665 15.6897 8.83908Z"
-              stroke="white"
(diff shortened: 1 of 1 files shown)


packages/x-archetype-utils/src/home/home-template.html
```diff
@@ -83,24 +83,20 @@
           class="x-archetype-header__input"
```
tiborux, line 83: is the lens so big?
image

packages/x-archetype-utils/src/home/home-template.html
```diff
@@ -1706,6 +1702,7 @@ <h1 class="x-archetype-section1__title x-margin--top-05 x-margin--bottom-07">
 
     --x-archetype-color-base-neutral-10: #283034;
     --x-archetype-color-lead: #243d48;
+    --x-archetype-button-color: #5e7782;
```
LuisMartinez15, line 1705: isn't this color different than the one we are using in the archetype? Also, shouldn't the icon have `border-radius` instead of being square? Ii is only visible when you zoom in, though.

`home-template`
image

`archetype`
image

LuisMartinez15 (changes requested):

packages/x-archetype-utils/src/home/home-template.html
annacv: @LuisMartinez15 I'm checking [Figma](https://www.figma.com/file/mYfJZHn668mGEGf61Sji6m/X-Archetype?node-id=5541%3A83202&t=vomH4eTjmAU5fgOo-0) and the color and icon size are the ones we have in the designs. 
Agree with border-radius, set to `border-radius: 0px 4px 4px 0px;` in designs.

LuisMartinez15 (approved):

packages/x-archetype-utils/src/home/home-template.html
tiborux: In figma, the home page has these styles:

image

in the home-template:

image

image

The color and size are differents

CachedaCodes: Color is changed because the color in the designs doesn't align with the color of the buttons in the archetype. If you use the template with the color in the designs in the achetype the button color "changes" and it's weird.

I'll check the border radius.

tiborux (dismissed):

LuisMartinez15 (dismissed):

LuisMartinez15 (approved):

tiborux (approved):

PR #1011, opened 2023-01-17
feat(design-system): modify default styles in the `suggestion-group`

tiborux:
This will allow the last element of a suggestion group to go to the end of the container.

EX-7845


annacv (approved):

tiborux:
> Should we add an example without setting the `width` to `200px`? To show the alignment behaviour in the showcase. Good job! 👌

I think that its not necessary, by default its positioning the button at the end.

ajperezbau (approved):

PR #1014, opened 2023-01-17
feat(history-queries): add dynamic classes to the history query suggestion

tiborux:
EX-7853


LuisMartinez15 (approved):
good job! 👏🏿 👏🏿

ajperezbau (commented):
Should we set a more generic name?
For example, `suggestionClass` and `removeButtonClass`?
That way, if we end up adding classes to the other suggestion components, all will be similar.
Good job! 👌

ajperezbau (approved):
Nice 👌

LuisMartinez15 (approved):

PR #1018, opened 2023-01-19
feat: Rename XDS product image class to make easier using it in x-components

annacv:
EX-7836

As discussed in meeting, class changed


diegopf (dismissed):

ajperezbau (dismissed):

ajperezbau (commented):
I think we should rename the function `productImage`, the folder and the `xds-component`, right?

diegopf (approved):

ajperezbau (approved):

PR #1025, opened 2023-01-20
feat(design-system): add `product-image` `XDS` component

annacv:
[EX-7836](https://searchbroker.atlassian.net/browse/EX-7836)

As you can check on [designs](https://www.figma.com/file/9a1DMgqIHt5ZZBvS2waaOs/X-DS?node-id=3252%3A32566&t=UULy6bDnwoUByduZ-0) this component will have 2 hover effects as variants, `x-picture-zoom` and `x-picture-overlay`. These classes will be applied to the `x-archetype` implementation, we need to disconnect them from the component, and also they shouldn't be applied at the same time (UI shouldn`t be overloaded with 2 effects), so we can't inherit a default behavior from the component itself.

**BREAKING CHANGE**

[EX-7836]: https://searchbroker.atlassian.net/browse/EX-7836?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/x-components/src/components/result/base-result-image.vue
```diff
@@ -208,14 +208,13 @@
 
 <style lang="scss" scoped>
   .x-result-picture {
+    // the loaderStyles prop is positioning absolutely to this container
```
annacv, line 211: Not sure if it's worth to leave this comment here

packages/x-components/src/components/result/base-result-image.vue
```diff
@@ -208,14 +208,13 @@
...
+    position: relative;
     min-width: 1px;
     min-height: 1px;
-    position: relative;
 
     &__image {
-      max-width: 100%;
```
annacv, line 216: If I'm not wrong.... 'max-width: 100%' is the default value for image tags, so removed

packages/x-components/src/components/result/base-result-image.vue
```diff
@@ -208,14 +208,13 @@
...
       max-height: 100%;
-      object-fit: contain;
```
annacv, line 218: `object-fit` gets applied with the `x-picture  img` styles coming from the new XDS

packages/x-components/src/design-system-deprecated/components/picture/default.scss
```diff
@@ -0,0 +1,43 @@
...
+  &__image {
+    // layout
+    display: block;
+
+    // others
+    object-fit: var(--x-object-fit-picture-default);
+    mix-blend-mode: var(--x-mix-blend-mode-picture-default);
+
+    &--fallback {
+      --x-object-fit-picture-default: var(--x-object-fit-picture-fallback-default) !important;
```
annacv, line 20: Lots of doubts here: I kept this on the not deprecated stylesheet, but with the values defined in the new XDS (contain, normal, respectively)

packages/x-components/src/design-system-deprecated/components/picture/default.scss
```diff
@@ -0,0 +1,43 @@
...
+      --x-mix-blend-mode-picture-default: var(--x-mix-blend-mode-picture-fallback-default) !important;
```
annacv, line 21: Lots of doubts here: I kept this on the not deprecated stylesheet, but with the values defined in the new XDS (contain, normal, respectively)

packages/x-components/src/design-system-deprecated/components/picture/default.tokens.scss
```diff
@@ -8,16 +8,18 @@
...
-  --x-color-fill-picture-placeholder-rect-default: var(--x-color-base-neutral-95);
-  --x-color-fill-picture-fallback-path-default: var(--x-color-base-neutral-70);
-  --x-color-fill-picture-fallback-rect-default: var(--x-color-base-neutral-95);
 
   // others
   --x-object-fit-picture-default: contain;
   --x-object-fit-picture-fallback-default: var(--x-object-fit-picture-default);
   --x-object-fit-picture-placeholder-default: var(--x-object-fit-picture-default);
-  --x-mix-blend-mode-picture-default: none;
+  --x-mix-blend-mode-picture-default: normal;
```
annacv, line 16: If I'm not wrong `none` is not a valid value, so although we are deprecating this, better with `normal`

packages/x-components/src/design-system-deprecated/components/picture/default.tokens.scss
```diff
@@ -8,16 +8,18 @@
...
   --x-mix-blend-mode-picture-fallback-default: var(--x-mix-blend-mode-picture-default);
   --x-mix-blend-mode-picture-placeholder-default: var(--x-mix-blend-mode-picture-default);
+
+  // color
+  --x-color-fill-picture-placeholder-path-default: var(--x-color-base-neutral-70);
+  --x-color-fill-picture-placeholder-rect-default: var(--x-color-base-neutral-95);
+  --x-color-fill-picture-fallback-path-default: var(--x-color-base-neutral-70);
+  --x-color-fill-picture-fallback-rect-default: var(--x-color-base-neutral-95);
```
annacv, line 24: F! When the doubts I left in comments are solved, I will move this part "upstairs" at line 11 with the other color token

packages/x-components/src/design-system/components/picture/default.scss
```diff
@@ -1,48 +1,20 @@
 .x-picture {
-  // layout
-  display: block;
-  overflow: hidden;
```
annacv, line 4: `overflow` is removed as will be inherited from x-tailwind, and `block` is removed as it's being applied in an already HTML block element (`div`)

packages/x-components/src/design-system/components/picture/default.scss
```diff
@@ -1,48 +1,20 @@
...
-    var(--x-size-border-radius-bottom-picture-default)
-    var(--x-size-border-radius-left-picture-default);
-
-  // color
-  background: var(--x-color-background-picture-default);
-
   &__image {
-    // layout
-    display: block;
-
```
annacv, line 18: Idem, as this class ends up being applied on `div`, `img` and `svg` and all of them have `display: block` applied by default in browsers.

packages/x-components/src/design-system/components/picture/default.scss
```diff
@@ -1,48 +1,20 @@
...
+      mix-blend-mode: normal;
+
       & > rect {
-        fill: var(--x-color-fill-picture-placeholder-rect-default);
+        fill: var(--x-color-base-neutral-95);
       }
       & > path {
-        fill: var(--x-color-fill-picture-placeholder-path-default);
+        fill: var(--x-color-base-neutral-70);
       }
```
annacv, line 18: I procrastinated a lot thinking about this....., maybe we could add these styles in x-tailwind?

packages/x-components/tailwind.config.js
```diff
@@ -17,6 +17,7 @@ module.exports = {
   // See https://github.com/empathyco/x/pull/655#discussion_r948923711
   corePlugins: [
     'alignItems',
+    'backgroundColor',
```
annacv, line 20: We need to enable it to be able to implement the `x-picture-overlay` variant. 
It will need to add an `x-bg-neutral-50/60` gradient in the parent `div`.

packages/x-tailwindcss/src/x-tailwind-plugin/components/picture/default.ts
```diff
@@ -9,7 +9,6 @@ import { TailwindHelpers } from '../../../types';
 export function pictureDefault(helpers: TailwindHelpers) {
   const { theme } = helpers;
   return {
-    display: 'block',
```
annacv, line 12: Removed as this class ends up being applied on a `div`, & it has `display: block` applied by default in browsers.

packages/x-components/src/components/result/base-result-image.vue
```diff
@@ -31,7 +31,7 @@
         :key="imageSrc"
         :alt="result.name"
         :src="imageSrc"
-        class="x-picture__image x-result-picture__image"
+        class="x-result-picture__image"
```
ajperezbau, line 34: I think this class can be renamed to `x-result-picture-image`.
Update the selector below too. (Check if it can affect something, just in case)

packages/x-components/src/components/result/base-result-image.vue
```diff
@@ -1,5 +1,5 @@
 <template>
```
ajperezbau, line 1: Remove the old DS used by the components in the demo.
I see classes on:

- `x-components/src/views/home/result.vue` -> Use here `x-picture-zoom`
- `x-components/src/x-modules/next-queries/components/next-query-preview.vue`
- `x-components/src/x-modules/queries-preview/components/query-preview.vue`
- `x-components/tests/unit/base-result-image.spec.ts`

Some of them are only in the component docs, but I think we should remove them from there as well to avoid confusion.

packages/x-components/src/design-system-deprecated/components/picture/default.scss
```diff
@@ -0,0 +1,43 @@
...
+
+      & > rect {
+        fill: var(--x-color-fill-picture-fallback-rect-default) !important;
+      }
+      & > path {
+        fill: var(--x-color-fill-picture-fallback-path-default) !important;
+      }
```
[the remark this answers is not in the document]
ajperezbau: It looks ok to me 👌

packages/x-components/src/design-system/components/picture/default.scss
ajperezbau: Be sure to set the `overflow: hidden` in the deprecated styles.

ajperezbau (changes requested):
Good job! 👌

mnavarroespinosa (changes requested):
Good job! 🚀

packages/x-components/src/components/result/base-result-image.vue
annacv: Yes, you're right @mnavarroespinosa . Thks!

packages/x-components/src/components/result/base-result-image.vue
annacv: @ajperezbau, @mnavarroespinosa I left the class `x-result__picture` as it comes from the `result.vue` component. Let me know if I'm missing something please!

packages/x-components/src/design-system/components/picture/default.scss
mnavarroespinosa: I would move the `display: block` to the deprecated styles too, for the same reason as [here](https://github.com/empathyco/x/pull/1025#discussion_r1083994594).

packages/x-components/src/components/result/base-result-image.vue
ajperezbau: Yep, you are right, `x-result__picture` will not disappear til the `result` xds component is deprecated.

packages/x-components/src/components/result/base-result-image.vue
```diff
@@ -208,14 +208,13 @@
...
-    // the loaderStyles prop is positioning absolutely to this container
     position: relative;
     min-width: 1px;
     min-height: 1px;
-
-    &__image {
-      max-height: 100%;
-    }
+  }
+  .x-result-picture-image {
```
ajperezbau, line 215: You can set this like:

```
<style lang="scss" scoped>
  .x-result-picture {
    position: relative;
    min-width: 1px;
    min-height: 1px;

    &-image {
      max-width: 100%;
      max-height: 100%;
    }
  }
</style>
```

packages/x-components/src/design-system-deprecated/components/picture/default.scss
```diff
@@ -1,4 +1,6 @@
 .x-picture {
+  overflow: hidden;
```
ajperezbau, line 2: Set here the `display: block`, it was already set before and it's been set in the new XDS.

```suggestion
  overflow: hidden;
  display: block;
```

ajperezbau (commented):
Good job 👌

packages/x-components/src/components/result/base-result-image.vue
annacv: Sure!

mnavarroespinosa (approved):

ajperezbau (approved):

PR #1026, opened 2023-01-20
style: update `prettier` and format code

LuisMartinez15:
EX-7288

With this update, `css` classes are not wrapped any more and are all displayed in the same line. In order to avoid `eslint` errors, @ajperezbau found an option to avoid triggering errors in lines with more than 100 characters, which is our limit, using the [ignorePattern](https://eslint.org/docs/latest/rules/max-len#ignorepattern) config.


ajperezbau (approved):
Beautiful 👌

mnavarroespinosa (approved):

PR #1006, opened 2023-01-16
feat(design-sytem): add sliding panel to XDS

diegopf:
EX-7687

PR #1027, opened 2023-01-23
refactor: make sliding panel take care buttons position and fade

diegopf:
EX-7687

After #1006 and trying to use the new XDS sliding panel in x we saw some potential issues and the need to duplicate css and classes in the sliding panel component from x. In order to avoid those issues, this PR reduces the scope for the XDS sliding panel and makes the design system only care about buttons position and fade. The scroll behaviour for the list will be a responsibility for the consumer of the design system. You can see how this will be used in x in this [branch](https://github.com/empathyco/x/tree/refactor/sliding-panel-xds)

### DIFF
--- a/packages/x-tailwindcss/demo/src/components/xds-sliding-panel.vue
+++ b/packages/x-tailwindcss/demo/src/components/xds-sliding-panel.vue
@@ -1,6 +1,6 @@
 <template>
   <XdsBaseShowcase
-    #default="{ cssClass, copyCssClassesToClipboard }"
+    #default="{ cssClass, section, copyCssClassesToClipboard }"
     title="Sliding panel"
     :sections="sections"
   >
@@ -10,8 +10,17 @@
       :class="cssClass"
       title="Click me to copy CSS classes"
     >
-      <button class="x-button x-sliding-panel-button x-sliding-panel-button-left">ᐸ</button>
-      <div class="x-sliding-panel-scroll x-gap-12">
+      <button class="x-button x-sliding-panel-button-left">ᐸ</button>
+      <div
+        class="x-gap-12 x-flex x-flex-row"
+        :class="{
+          'x-sliding-panel-fade': section === 'Fade',
+          'x-sliding-panel-fade-sm':
+            section === 'Combinations' && cssClass.includes('x-sliding-panel-buttons-outside'),
+          'x-sliding-panel-fade-lg':
+            section === 'Combinations' && cssClass.includes('x-sliding-panel-buttons-center')
+        }"
+      >
         <div
           v-for="(item, index) of items"
           :key="index"
@@ -22,7 +31,7 @@
           <span>{{ item }}</span>
         </div>
       </div>
-      <button class="x-button x-sliding-panel-button x-sliding-panel-button-right">ᐳ</button>
+      <button class="x-button x-sliding-panel-button-right">ᐳ</button>
     </div>
   </XdsBaseShowcase>
 </template>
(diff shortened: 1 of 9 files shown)


packages/x-tailwindcss/src/x-tailwind-plugin/components/sliding-panel/index.ts
```diff
@@ -1,16 +1,18 @@
 import { TailwindHelpers } from '../../../types';
 import { slidingPanelDefault } from './default';
+import { slidingPanelFade } from './fade';
 
 /**
  * Returns the component `sliding panel` CSS.
  *
- * @param helpers - The {@link TailwindHelpers} to generate CSS.
+ * @param helpers
```
annacv, line 8: Es lint is complaining without this 👀 
```suggestion
 * @param helpers - The {@link TailwindHelpers} to generate CSS.
```

packages/x-tailwindcss/src/x-tailwind-plugin/components/sliding-panel/buttons/hover.ts
```diff
@@ -21,18 +34,6 @@ export function slidingPanelButtonsHover() {
...
-      '.sliding-panel-button-left': {
-        opacity: 1,
-        pointerEvents: 'all'
-      }
-    },
-    '&:not(.sliding-panel-show-buttons-on-hover):not(.sliding-panel-at-end)': {
-      '.sliding-panel-button-right': {
-        opacity: 1,
-        pointerEvents: 'all'
-      }
```
mnavarroespinosa, line 35: After removing these rules,`pointerEvents` value is always `none` for Sliding Panels that have the "show buttons on hover" feature disabled, so the buttons are not clickable.

packages/x-tailwindcss/demo/src/components/xds-sliding-panel.vue
```diff
@@ -69,22 +78,17 @@
...
 
     public combinations = [
-      'x-sliding-panel-buttons-center x-sliding-panel-fade-sm',
-      'x-sliding-panel-buttons-center x-sliding-panel-fade-md',
-      'x-sliding-panel-buttons-outside x-sliding-panel-fade-lg',
+      'x-sliding-panel-buttons-center x-sliding-panel-show-buttons-on-hover',
+      'x-sliding-panel-buttons-outside x-sliding-panel-show-buttons-on-hover',
       // eslint-disable-next-line max-len
-      'x-sliding-panel-buttons-outside x-sliding-panel-fade-sm x-sliding-panel-show-buttons-on-hover x-sliding-panel-at-start',
+      'x-sliding-panel-buttons-outside x-sliding-panel-show-buttons-on-hover x-sliding-panel-at-start',
```
mnavarroespinosa, line 89: The fade seems to be displayed on the left side of the panel when it has the class `x-sliding-panel-at-start`, and the same happens with the right side of the panel with `x-sliding-panel-at-end`.

mnavarroespinosa (changes requested):

packages/x-tailwindcss/demo/src/components/xds-sliding-panel.vue
diegopf: Good catch! When creating the PR I've modified a little bit the code from the aux branch leading to this issue.

mnavarroespinosa (approved):
Looks good to me, good job!

annacv (approved):
Great job!!

PR #1033, opened 2023-01-24
feat: add `prettier-plugin-tailwindcss`

LuisMartinez15:
EX-7884


tiborux (approved):

tiborux:
> Interesting 👀 . I thought files were formatted in "oriental mode" == ordering alphabetically from right to left.... until I arrived at the `xds-base-showcase`. I cannot see any logic there 🤯 🤣

They are not ordered alphabetically, here, you can check why the classes are sorted like that: https://tailwindcss.com/blog/automatic-class-sorting-with-prettier#how-classes-are-sorted

PR #1035, opened 2023-01-24
fix(design-system): Use class instead of img tag

annacv:
[EX-7836](https://searchbroker.atlassian.net/browse/EX-7836)

Update XDS picture component to avoid using an `img` tag in XDS, use a `class` instead.

[EX-7836]: https://searchbroker.atlassian.net/browse/EX-7836?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mnavarroespinosa (approved):
LGTM!

ajperezbau (approved):
Good job 👌

PR #1036, opened 2023-01-24
feat: Deprecate x-picture__image and placeholder/fallback styles

annacv:
EX-7889

Merge this [PR](https://github.com/empathyco/x/pull/1035) first.

**BREAKING CHANGE:** 
* class `x-picture__image` changed to `x-picture-image`
* `x-picture__image--placeholder` and `x-picture__image--fallback` are deprecated. SVG's colors won't be customizable using tokens.


mnavarroespinosa (approved):
good job!

mnavarroespinosa (dismissed):

packages/x-components/src/components/result/base-result-image.vue
```diff
@@ -31,7 +32,7 @@
         :key="imageSrc"
         :alt="result.name"
         :src="imageSrc"
-        class="x-result-picture-image"
+        class="x-result-picture-image x-picture-image"
```
ajperezbau, line 35: I think that the `x-picture-image` class is not needed here, it will be appended from the dynamic component tag.

packages/x-components/src/views/home/result.vue
```diff
@@ -3,7 +3,7 @@
     <BaseResultLink :result="result">
       <BaseResultImage
         :result="result"
-        class="x-result__picture x-picture-zoom"
+        class="x-result-picture x-picture-zoom"
```
ajperezbau, line 6: This class should not be changed.

packages/x-components/src/components/result/base-result-fallback-image.vue
```diff
@@ -1,12 +1,12 @@
 <template functional>
   <svg
-    :class="['x-picture__image--fallback'].concat(data.staticClass, data.class)"
```
ajperezbau, line 3: I think we should keep here the class attribute:
 
```
    :class="[].concat(data.staticClass, data.class)"
```

This is a functional component, if we do not do that, we will not be able to add classes when using it.

packages/x-components/src/components/result/base-result-placeholder-image.vue
```diff
@@ -1,12 +1,12 @@
 <template functional>
   <svg
-    :class="['x-picture__image--placeholder'].concat(data.staticClass, data.class)"
```
ajperezbau, line 3: I think we should keep here the class attribute:
 
```
    :class="[].concat(data.staticClass, data.class)"
```

This is a functional component, if we do not do that, we will not be able to add classes when using it.

ajperezbau (changes requested):
Good job 👌

packages/x-components/src/views/home/result.vue
annacv: F! True

packages/x-components/src/components/result/base-result-image.vue
annacv: I'm not catching you, before we had the `aspect-ratio` and other styles coming from the `img` tag styles attribution made on XDS. In the last related [PR](https://github.com/empathyco/x/pull/1035) we changed from `img` to use `class`, so I applied here he class accordingly.

ajperezbau: I think that I didn't explain myself very well xd
The `x-picture-image` class is already set, the `<component :is="animation"/>` tag will render a Vue Transition or a NoElement.
Because these two are renderless components, `Vue` will inherit the attr class and set it to the element that is rendering.
You can test this by removing the class from the image and inspecting the element in the devtools, it should contain the class because it's been set in the `component` parent.

annacv: I've seen it now! Thks!!!

mnavarroespinosa (approved):

ajperezbau (approved):

PR #1039, opened 2023-01-26
feat: Adapt Result mapping according to new API standard

annacv:
[EX-6376](https://searchbroker.atlassian.net/browse/EX-6376)

**Main changes:** 
* All the fields will be preceded by a double dash `__` to follow the new [API contract](https://github.com/empathyco/search-empathy-plugin/blob/main/docs/api/openapi.yml#L276)
* Image changes from `string` to`string[]` type
* Added new values from Platform Response:
      * ` __externalId`: will be the one used to map as the identifier (used for SKU searches)
      * `price.future.value` is included [in the ResultPrice model](https://github.com/empathyco/search-empathy-plugin/blob/main/docs/api/openapi.yml#L1497) as an optional field related to the prewarming price feature.
* Update mocked objects and fix tests.

[EX-6376]: https://searchbroker.atlassian.net/browse/EX-6376?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/x-adapter-platform/src/types/models/result.model.ts
```diff
@@ -6,10 +6,21 @@ import { PlatformTagging } from './tagging.model';
  * @public
  */
 export interface PlatformResult {
-  id: string;
-  name: string;
-  image: string;
-  price: number;
-  url: string;
+  __id: string;
```
tiborux, line 9: Can use add the documentation about these properties? Maybe you can use the one in the `openapi.yml`

packages/x-adapter-platform/src/__tests__/__fixtures__/recommendations.response.ts
```diff
@@ -2,26 +2,31 @@ export const platformRecommendationsResponse = {
   topclicked: {
     content: [
       {
-        image: 'https://assets.empathy.co/images-demo/31335.jpg',
+        __images: ['https://assets.empathy.co/images-demo/31335.jpg'],
```
tiborux, line 5: Be careful with this, it seems that the top-clicked response is not standardized. We are going to wait until the response will be standarized.

packages/x-adapter-platform/src/__tests__/__fixtures__/identifier-results.response.ts
```diff
@@ -3,8 +3,8 @@ export const platformIdentifierResultsResponse = {
   catalog: {
```
tiborux, line 3: Be careful with this, it seems that the top-clicked response is not standardized. We are going to wait until the response will be standarized.

tiborux:
We are going to wait until "skuSearch" and "recommendations" will be standardized because right now they are using the result schema and with that schema, the should be standardized, if not, we should create a new schema for them.

PR #1045, opened 2023-01-27
fix(search): reset `query`, `page` and `sort` on clear query

mnavarroespinosa:
EX-7663

- Fixes issue that keeps the `page` param after clicking `Clear` query button in the search box. The same issue was happening with the `sort` parameter so I reset it too. 
- I opted for adding a new action `clearSearchParams` over using the existing `resetState` as the latter is only dispatched when requests are made and it's meant to evaluate the params that have changed between the old and new request. 
- Added an `e2e` regression test marked with `@regression` tag to facilitate identifying the regression tests.


tiborux:
It looks nice, but I was trying to find a way to not "hardcoding" the parameters that we want to clean. Right now, we are using the resetStateWire but it needs the request to check if should we clean the parameters, so I don't know if someone can find a better way to clean these parameters, in the past, we did the same but cleaning one by one the parameters.

packages/x-components/src/x-modules/search/store/actions/clear-search-params.action.ts
```diff
@@ -0,0 +1,18 @@
+import { SearchXStoreModule } from '../types';
```
tajespasarela, line 1: remove unnecessary action

tajespasarela (changes requested):

diegopf (commented):
This new approach seems that is not grouping changes properly when selecting filters now:

1. Scroll to page 2 (or beyond)
2. Select a filter
3. Two requests are made: one with the wrong page and another one with the page reset. The one with the wrong page is cancelled, but still we should avoid this.

packages/x-components/src/x-modules/search/events.types.ts
```diff
@@ -39,9 +39,14 @@ export interface SearchXEvents {
   SearchRequestChanged: InternalSearchRequest | null;
   /**
    * Any property of the search request has been updated.
-   * Payload: The new internal request object.
+   * Payload: The new search request or `null` if there is not enough data in the state to
+   * conform a valid request.
+   *
+   * @remarks The difference from `SearchRequestChanged` and this event is this one will be executed
```
annacv, line 45:
```suggestion
   * @remarks The difference between `SearchRequestChanged` and this event is this one will be executed
```

annacv:
For me it's working fine

mnavarroespinosa (commented):
Looks good to me!

tiborux (approved):

tajespasarela (approved):

PR #1050, opened 2023-01-27
feat(design-system): Add layout container XDS

annacv:
[EX-7824](https://searchbroker.atlassian.net/browse/EX-7824)

**To discuss/agree on**:
These max-width values are taken from the [spike](https://github.com/empathyco/x/tree/spike/EX-5737-re-think-layouts).
Are they what we need?

`layoutMaxWidth: {
    sm: '1000px',
    md: '1440px',
    lg: '2000px',
    full: '100vw'
  },`

Here are some other references to make a decision:
- [Bootstrap max-widths for different breakpoints](https://getbootstrap.com/docs/5.0/layout/containers/)
- [Tailwind container definitions (max-width related to breakpoints)](https://tailwindcss.com/docs/container)

[EX-7824]: https://searchbroker.atlassian.net/browse/EX-7824?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/x-tailwindcss/demo/src/components/xds-layout.vue
```diff
@@ -0,0 +1,78 @@
...
+      ]
+    })
+    public sizes!: string[];
+
+    @Prop({
+      default: () => [
+        'x-layout-container-min-margin-12',
+        'x-layout-container-min-margin-20',
+        'x-layout-container-min-margin-32',
+        'x-layout-container-min-margin-48'
```
mnavarroespinosa, line 53: Maybe it can lead to confusion to only have 4 classes listed here instead of all the available (as many as spacing values in the theme) but I don't see a better approach for now.

packages/x-tailwindcss/demo/src/components/xds-layout.vue
```diff
@@ -0,0 +1,78 @@
...
+        Margins: this.minMargin.map(addParentClasses(this.base))
+      };
+    }
+  }
+</script>
+
+<style lang="scss" scoped>
+  ::v-deep h2 {
+    align-self: flex-start;
+  }
```
mnavarroespinosa, line 71: Do we need this rule?

mnavarroespinosa (changes requested):
Good job!

packages/x-tailwindcss/demo/src/components/xds-layout.vue
annacv: I think not. I'll check it. Thks!

packages/x-tailwindcss/demo/src/components/xds-layout.vue
annacv: It would be too much, and this showcase's goal should be focused on the relationship between the items and the container. Maybe we can add a comment in the showcase file to point out that we can use all the margins available in the `theme.ts` file. What do you think @mnavarroespinosa?

tiborux: Reading the task, it said that maybe we should follow the approach of creating a template like the archetype one:
```
The showcase can be similar to the layout used in the archetype. It can be started in this task and completed in the layout item task.
```

So maybe we should remove this showcase and create one that simulates a real layout.

mnavarroespinosa (approved):

tiborux (approved):

mnavarroespinosa (dismissed):

tiborux (approved):

mnavarroespinosa (approved):

PR #1052, opened 2023-01-28
fix: add missing `rxjs` dependency and narrow `setTimeout` type

LuisMartinez15:
EX-7928

add missing dependency `rxjs`.

`typescript` is messing with the `setTimeout` type, as it has a different definition in `nodejs` than in the `window` object, so we have to narrow it to the `window` one. My only concern is that this code is not `ssr` friendly, as we are using the `window.setTimeout` directly. In case we need to support `ssr`, we might have to wrap the setTimeout function to distinguish the platform we are running the code on.

### DIFF
--- a/packages/x-bus/src/x-bus.types.ts
+++ b/packages/x-bus/src/x-bus.types.ts
@@ -2,15 +2,6 @@ import { Dictionary } from '@empathyco/x-utils';
 import { Observable, Subject } from 'rxjs';
 
 /**
- * Alias representing a
- * {@link https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#return_value | timeoutID}
- * value.
- *
- * @internal
- */
-export type TimeoutId = number;
-
-/**
  * Alias representing a positive number to represent the priority of an event in an
  * {@link XPriorityBus}.
  *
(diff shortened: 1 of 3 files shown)


packages/x-bus/src/x-bus.types.ts
```diff
@@ -1,15 +1,6 @@
...
 import { Observable, Subject } from 'rxjs';
 
-/**
- * Alias representing a
- * {@link https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#return_value | timeoutID}
- * value.
- *
- * @internal
- */
-export type TimeoutId = number;
```
ajperezbau, line 11: What do you think about keep using this type, setting it to:

```
/**
 * Alias representing a
 * {@link https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#return_value | timeoutID}
 * value.
 * It is a `number` in browser/jsdom environments, or a `Timeout` object in `node` environments.
 *
 * @internal
 */
export type TimeoutId = ReturnType<typeof setTimeout>;
```

This way it should be compatible with both environments and we can use `setTimeout` instead of `window.setTimeout`.
This type is being used in some utils of `x-components`, `x-components/src/utils/debounce`, for example.
Maybe we can move it to the `x-types` package, to make it reusable across packages.

ajperezbau (commented):
Good job 👌

packages/x-bus/src/x-bus.types.ts
mnavarroespinosa: @LuisMartinez15, I think @ajperezbau already took a look, and I'm testing it now too.

ajperezbau (approved):

mnavarroespinosa (approved):

PR #1056, opened 2023-02-02
feat(design-system): Add scroll component

tiborux:
I couldnt find a way to style the hover color in firefox, I mean, in chrome you can change the color when you hover the scrollbar itself but for firefox I couldn't find a way to do it, so the scrollbar will change the color when you hover in an `x-scroll` in any of it parts.

Another thing that I am not sure it is if we should use, [scrollbar-gutter](https://css-tricks.com/almanac/properties/s/scrollbar-gutter/).  [Can I Use](https://caniuse.com/?search=scrollbar-gutter)

EX-7956

### DIFF
--- a/packages/x-tailwindcss/src/x-tailwind-plugin/components/scroll/default.ts
+++ b/packages/x-tailwindcss/src/x-tailwind-plugin/components/scroll/default.ts
@@ -13,27 +13,26 @@ export function scrollDefault(helpers: TailwindHelpers) {
     '--x-color-thumb-scroll-bar-hover': theme('colors.neutral.50'),
     '--x-color-background-scroll-bar': 'transparent',
 
-    '@media (hover: hover)': {
-      '&::-webkit-scrollbar': {
-        width: '17px',
-        '&-thumb': {
-          backgroundClip: 'content-box',
-          backgroundColor: 'var(--x-color-thumb-scroll-bar)',
-          borderColor: 'transparent',
-          borderWidth: theme('spacing.4'),
-          borderRadius: theme('spacing.24'),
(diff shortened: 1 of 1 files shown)


annacv:
Hi! I have been checking [this article](https://www.bram.us/2021/06/23/css-at-supports-rules-to-target-only-firefox-safari-chromium/) about using CSS `@supports` to target browsers. Could it be worth using it for only Firefox styles? 

On the other hand, `scrollbar-gutter` seems to be a nice approach, although Safari doesn't support it.

packages/x-tailwindcss/src/x-tailwind-plugin/components/scroll/default.ts
```diff
@@ -0,0 +1,38 @@
...
+ */
+export function scrollDefault(helpers: TailwindHelpers) {
+  const { theme } = helpers;
+  return {
+    overflowY: 'scroll',
+    '--x-color-thumb-scroll-bar': theme('colors.neutral.25'),
+    '--x-color-thumb-scroll-bar-hover': theme('colors.neutral.50'),
+    '--x-color-background-scroll-bar': theme('colors.neutral.0'),
+
+    '@media (hover: hover)': {
```
mnavarroespinosa, line 16: For some reason, the rules inside the `media` are repeated five times in the generated CSS.

mnavarroespinosa (commented):
For me, I see the same way the scroll in Firefox than Chrome, and regarding the gutter, I think we may add it later as an improvement in case we are sure to add it.

mnavarroespinosa (approved):

annacv (approved):

PR #1064, opened 2023-02-07
feat(design-system): add layout item utilities

mnavarroespinosa:
EX-7826


packages/x-tailwindcss/demo/src/components/xds-layout-utils.vue
```diff
@@ -0,0 +1,112 @@
...
+    public overlap!: string[];
+
+    @Prop({
+      default: () => ['x-layout-expand']
+    })
+    public expand!: string[];
+
+    protected get sections(): ShowcaseSections {
+      return {
+        Default: [this.base],
```
annacv, line 104: I would say we don't need the `Default` section here. We already have it in the `xds-layout-showcase.vue`, although here the `base` is referred to the `x-layout-item` and not the container, we are applying the same classes.

packages/x-tailwindcss/src/x-tailwind-plugin/components/layout/utils.ts
```diff
@@ -0,0 +1,55 @@
...
+ * Returns the component `layout` utils CSS.
+ *
+ * @param helpers - The {@link TailwindHelpers} to generate CSS.
+ * @returns The {@link CssStyleOptions} for the component.
+ */
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export function utils(helpers: TailwindHelpers) {
+  return {
+    '.layout-expand': {
+      height: '100%',
```
annacv, line 13: I'm not sure if this height: '100%' is needed. Please, ping me if I'm wrong & let's check it 🙏

tiborux (approved):

annacv (approved):
🚀

PR #1074, opened 2023-02-14
fix(modal): set focusOnOpen for the main-modal to false

CachedaCodes:
EX-7710

## Motivation and context
The `focusOnOpen` prop is set to true by default in the `BaseModal` component.

In the case of the `MainModal`, it does not make sense, because if there is a focusable element before the `SearchInput`, the user will have to click the input manually to be able to write.

This change is to avoid modifying the prop in all the setups.

## Type of change

- [X] `Main`
- [ ] Other. Specify:

## How has this been tested?
Check that the `BaseModal` inside the `MainModal` opens with `focusToOpen` as false. Try changing the prop in `MainModal` to check that the prop is working correctly.

### DIFF
--- a/packages/x-components/src/components/modals/main-modal.vue
+++ b/packages/x-components/src/components/modals/main-modal.vue
@@ -6,7 +6,7 @@
     :eventsToCloseModal="closeEvents"
     :bodyClickEvent="outOfModalClickEvent"
     :animation="animation"
-    :focusOnOpen="false"
+    :focusOnOpen="focusOnOpen"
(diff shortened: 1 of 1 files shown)


annacv:
> I think we should add `focusOnOpen` as a prop, and set it to `false` by default. If we set it to false directly, we will not be able to override it when using the component.

Agree, on the task it is specified 👀 "Keep in mind that the prop should be configurable."

ajperezbau (approved):

annacv (approved):
🚀

PR #1077, opened 2023-02-14
feat: remove `x-suggestion` class from `BaseSuggestion` component

ajperezbau:
Add the class to the components used in the lists.

This is needed because in the archetype we want to use a `NextQuery` with `x-button` styles, but adding it right now conflicts with the `x-suggestion` styles.

EX-8006

### DIFF
--- a/packages/x-components/src/x-modules/history-queries/components/history-queries.vue
+++ b/packages/x-components/src/x-modules/history-queries/components/history-queries.vue
@@ -17,6 +17,7 @@
           :suggestion="baseScope.suggestion"
           data-test="history-query-item"
           class="x-history-queries__item"
+          suggestionClass="x-suggestion"
         >
           <template #default="historyQueryScope">
             <!-- eslint-disable max-len -->
--- a/packages/x-components/src/x-modules/next-queries/components/next-queries.vue
+++ b/packages/x-components/src/x-modules/next-queries/components/next-queries.vue
@@ -18,7 +18,7 @@
           #default="nextQueryScope"
           :suggestion="baseScope.suggestion"
           :highlightCurated="highlightCurated"
-          class="x-next-queries__suggestion"
+          class="x-next-queries__suggestion x-suggestion"
         >
           <!-- eslint-disable max-len -->
           <!--
--- a/packages/x-components/src/x-modules/popular-searches/components/popular-searches.vue
+++ b/packages/x-components/src/x-modules/popular-searches/components/popular-searches.vue
@@ -15,7 +15,7 @@
       <slot name="suggestion" v-bind="{ ...baseScope }">
         <PopularSearch
           :suggestion="baseScope.suggestion"
-          class="x-popular-searches__suggestion"
+          class="x-popular-searches__suggestion x-suggestion"
           #default="popularSearchScope"
         >
           <!-- eslint-disable max-len -->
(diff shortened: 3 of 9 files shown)


packages/x-components/src/x-modules/next-queries/components/next-queries.vue
```diff
@@ -18,7 +18,7 @@
           #default="nextQueryScope"
           :suggestion="baseScope.suggestion"
           :highlightCurated="highlightCurated"
-          class="x-next-queries__suggestion"
+          class="x-next-queries__suggestion x-suggestion"
```
annacv, line 21: Hi! I think this `x-next-queries__suggestion` class can be removed

packages/x-components/src/x-modules/popular-searches/components/popular-searches.vue
```diff
@@ -15,7 +15,7 @@
       <slot name="suggestion" v-bind="{ ...baseScope }">
         <PopularSearch
           :suggestion="baseScope.suggestion"
-          class="x-popular-searches__suggestion"
+          class="x-popular-searches__suggestion x-suggestion"
```
annacv, line 18: Idem I think `x-popular-searches__suggestion` can be removed

packages/x-components/src/x-modules/query-suggestions/components/query-suggestions.vue
```diff
@@ -15,7 +15,7 @@
       <slot name="suggestion" v-bind="{ ...baseScope }">
         <QuerySuggestion
           :suggestion="baseScope.suggestion"
-          class="x-query-suggestions__suggestion"
+          class="x-query-suggestions__suggestion x-suggestion"
```
annacv, line 18: Idem, maybe `x-query-suggestions__suggestion` is removable?

packages/x-components/src/x-modules/history-queries/components/history-queries.vue
```diff
@@ -17,6 +17,7 @@
           :suggestion="baseScope.suggestion"
           data-test="history-query-item"
           class="x-history-queries__item"
+          suggestionClass="x-suggestion"
```
CachedaCodes, line 20: We should take the chance and implement the dynamic prop for the `suggestionClass` like we have in the `HistoryQuery` in the `NextQuery`, `PopularSearch` and `QuerySuggestion` components too

packages/x-components/src/x-modules/next-queries/components/next-queries.vue
CachedaCodes: I would leave them, removing them is a breaking change, we should decide what to do with BEM like classes in the project. This topic should be brought up in some meeting (refinement or standup maybe)

CachedaCodes (changes requested):

packages/x-components/src/x-modules/history-queries/components/history-queries.vue
ajperezbau: Yep, that's a good point, I will include them 👌

packages/x-components/src/x-modules/next-queries/components/next-queries.vue
ajperezbau: Yep, I would avoid removing those classes for now to prevent breaking changes

packages/x-components/src/x-modules/history-queries/components/history-queries.vue
ajperezbau: Looking at this better, you mean to implement the `suggestionItemClass` in the `NextQueries`, `PopularSearches` and `QuerySuggestions` components, right?
The components that you named have the `BaseSuggestion` as the first child, so they don't need a prop for it.

CachedaCodes (approved):

annacv (approved):
🎸

PR #1078, opened 2023-02-16
fix(components): Fix open issue when focused on body. Fix `SearchInputPlaceholder` styles.

tajespasarela:
EX-8011

Workaround to fix the issue when opening the XArchetype.

To reproduce the issue, go to XArchetype and change the dynamic import of the MainModal to a normal import. Then run the proyect and you won't be able to open X. It is open and immediately closed.

Sometimes the issue is not reproducible due to the code order 🤷🏻


annacv (approved):
🎉

ajperezbau (changes requested):
It is not working for me if a do an `install local` in the archetype

CachedaCodes (approved):

PR #1079, opened 2023-02-16
fix: Update home styles to match `inputGroup`.

tajespasarela:
EX-8014


annacv (approved):
There is another thing that differs from the x-archetype input-group: at home HTML there is no hover in the button, while in the x-archetype input-group the button changes its color to `#243d48`
I will leave this approved and if finally the hover is added, just ping me to approve again

ajperezbau (approved):
Agree with @annacv.
Besides that, it looks good to me 👌

PR #1084, opened 2023-02-20
feat(query-preview): add query preview list component to orchestrate requests

ajperezbau:
EX-7541


annacv:
Hi! I have tried to figure out what to do with tests failing, but I'm not sure how to test it: here is a (hope) useful link to check: 
[https://github.com/cypress-io/cypress/issues/5889](https://github.com/cypress-io/cypress/issues/5889)

tiborux:
Maybe with the new priority-bus we should check some things here, because in the other pr we had a lot of problems with the `query-previews` request

packages/x-components/src/utils/throttle.ts
```diff
@@ -5,27 +5,46 @@ import { ThrottleFunction } from './types';
...
  * @returns A new function with the throttle.
  *
  * @public
  */
 
 export const throttle = <Params extends any[]>(
   fn: (...args: Params) => void,
-  throttleTimeMs: number
+  throttleTimeMs: number,
+  { leading = false }: ThrottleOptions = {}
```
mnavarroespinosa, line 17: Just to understand it, Is this change needed for this task?

mnavarroespinosa (changes requested):

packages/x-components/src/utils/throttle.ts
tiborux: I am in the same spot, i don't know why the change is needed

packages/x-components/src/x-modules/queries-preview/components/query-preview-list.vue
```diff
@@ -0,0 +1,215 @@
+<template>
+  <ul>
```
tiborux, line 2: I think that we should add classes here, they are not needed right now but just in case we should have classes in `<ul>` and `<li>`

packages/x-components/src/x-modules/queries-preview/components/query-preview-list.vue
```diff
@@ -0,0 +1,215 @@
...
+
+    /**
+     * Gets all the queries to render, that are those that don't have an `error` status.
+     *
+     * @returns A list of queries.
+     * @internal
+     */
+    protected get renderedQueries(): string[] {
+      return this.queries.filter(
+        query => this.queriesStatus[query] === 'success' || this.queriesStatus[query] === 'loading'
```
tiborux, line 61:
```suggestion
        query => this.queriesStatus[query] !== 'error'
```

packages/x-components/src/x-modules/queries-preview/components/query-preview-list.vue
```diff
@@ -0,0 +1,215 @@
...
+      );
+    }
+
+    /**
+     * Resets the status of all queries if they change.
+     *
+     * @internal
+     */
+    @Watch('queries', { immediate: true })
+    protected resetStatusRecord(): void {
```
tiborux, line 71: I was trying to follow the approach that we used in the `base-result-image` where instead of mapping the status of the current query, we save a copy of the queries as pendingQueries and every time that a query is being load or set as failed, we remove it from the pendindQueries or add it to the loadedQueries, but i couldnt make it works. I am not sure if should be a better approach or not

packages/x-components/src/utils/throttle.ts
ajperezbau: Yep, it is not needed for this task.
Take it as a gift from our beloved @JaviIG 🎁

javieri-empathy: It can be removed as far as I remember. It is not needed, as requests go one by one. Either that approach or a simple delay is enough 
Salu2 ❤️

packages/x-components/src/utils/__tests__/throttle.spec.ts
```diff
@@ -23,4 +23,38 @@ describe('testing throttle util', () => {
     expect(originalFn).toHaveBeenCalledTimes(2);
     expect(originalFn).toHaveBeenLastCalledWith(4);
   });
+
```
tiborux, line 26: So I think that we can remove this test too.

packages/x-components/src/x-modules/queries-preview/components/__tests__/query-preview-list.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+      // A timeout with no time should resolve after all reactivity + promises involved
+      return new Promise(resolve => setTimeout(resolve));
+    }
+  };
+}
+
+describe('testing QueryPreviewList', () => {
+  it('renders a list of queries one by one', async () => {
+    const { getQueryPreviewItemWrappers, reRender } = renderQueryPreviewList({
+      template: `
```
tiborux, line 61: we should use this template as the default one, we are using the same template in both tests.

packages/x-components/src/x-modules/queries-preview/components/__tests__/query-preview-list.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+interface RenderQueryPreviewListOptions {
+  /** The template to render the {@link QueryPreviewList} component. */
+  template?: string;
+  /** The queries for which preview its results. */
+  queries?: string[];
+  /** The results to return from the mocked search endpoint adapter. */
+  results?: Record<string, Result[]>;
+}
+
+interface RenderQueryPreviewListApi {
```
tiborux, line 121:
```suggestion
interface RenderQueryPreviewListAPI {
```

packages/x-components/src/views/home/Home.vue
```diff
@@ -254,34 +254,28 @@
             </div>
 
             <template v-if="!$x.query.searchBox">
-              <h1 class="x-margin--bottom-05 x-title1">Brand Recommendations</h1>
+              <h1 class="x-mb-16 x-title1">Brand Recommendations</h1>
```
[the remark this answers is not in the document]
tiborux: In that case, we can manage the logic in the archetype/setup

packages/x-components/src/x-modules/queries-preview/components/__tests__/query-preview-list.spec.ts
```diff
@@ -0,0 +1,128 @@
...
+    let queryPreviews = getQueryPreviewItemWrappers();
+    expect(queryPreviews.wrappers).toHaveLength(1);
+    expect(queryPreviews.at(0).text()).toEqual('');
+
+    await reRender();
+    queryPreviews = getQueryPreviewItemWrappers();
+    expect(queryPreviews.wrappers).toHaveLength(1);
+    expect(queryPreviews.at(0).text()).toEqual('shoes - Crazy shoes');
+  });
+  // TODO When error handling is added this component should exclude also queries that failed.
```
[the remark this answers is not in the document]
mnavarroespinosa: @tiborux added test for this case

mnavarroespinosa (approved):
lgtm!

tiborux (approved):

PR #1087, opened 2023-02-24
feat: migrate `AutoProgressBar` to Composition API

tiborux:
EX-8049


packages/x-components/src/components/auto-progress-bar.vue
```diff
@@ -5,43 +5,49 @@
...
+    },
+    setup(props) {
+      /**
+       * Computed property to calculate the animation's duration.
+       *
+       * @returns The CSS styles of the animation.
+       *
+       * @internal
+       */
+      const cssStyles = computed(() => ({ animationDuration: `${props.durationInSeconds}s` }));
```
annacv, line 45: We have three options to type this:

Typing the constant: 
`const cssStyles: ComputedRef<Partial<CSSStyleDeclaration>> = computed(() => ({
        animationDuration: `${props.durationInSeconds}s`
      }));`

Typing the getter: 
`const cssStyles = computed(
        (): Partial<CSSStyleDeclaration> => ({ animationDuration: `${props.durationInSeconds}s` })
      );`

Typing both (I think it's a bit redundant doing it in both, as the getter would be inferred if we type the constant, and, in fact, it should be `ComputedGetter<Partial<CSSStyleDeclaration>>` ):
`const cssStyles: ComputedRef<Partial<CSSStyleDeclaration>> = computed(
        (): Partial<CSSStyleDeclaration> => ({ animationDuration: `${props.durationInSeconds}s` })
      );`

annacv (changes requested):
Great job! 

Maybe we can agree with the team on which strategy we want to use to type computed props in Vue3

packages/x-components/src/components/auto-progress-bar.vue
tiborux: Reading the vue doc, they suggest to type the computed using the generic argument: [vue doc](https://vuejs.org/guide/typescript/composition-api.html#typing-computed)

```
  const cssStyles = computed<Partial<CSSStyleDeclaration>>(() => ({
        animationDuration: `${props.durationInSeconds}s`
      }));
```

annacv (dismissed):
🎸

LuisMartinez15 (dismissed):
good job 👏🏿 👏🏿

LuisMartinez15 (approved):

PR #1090, opened 2023-02-28
fix(design-system): fix deprecated typography and picture components

mnavarroespinosa:
EX-8053


annacv (approved):
good job!

packages/x-components/src/design-system-deprecated/components/picture/default.scss
```diff
@@ -11,7 +11,7 @@
   // color
   background: var(--x-color-background-picture-default);
 
-  &__image {
```
tiborux, line 14: I am having doubts about this change. If you change this name, it means that you need to use `x-picture-image` in the setup that is importing the deprecated styles. So I found X scenarios:

1. Update `x-components` and `x-tailwindcss` in the archetype and remove the import for the deprecated styles.
2. Update `x-components` in the archetype and keep using the deprecated styles. In this case, if you are using a hardcoded `x-picture__image` is not going to work.

I am a little bit confused because I am not sure whats the expected behavior.

mnavarroespinosa: I made this change because the `BaseResultImage` was modified, in `3.0.0-alpha.302` release, using `x-picture-image` class for the image element instead of `x-picture__image`.
In the 2nd case you mentioned, `x-components`'s `BaseResultImage` is not using `x-picture__image` anymore so importing the deprecated styles won't have any effect on the images. With this change, we will be able to use the deprecated styles for the images in case `x-components` version is updated.

tiborux (approved):

PR #1103, opened 2023-03-14
Feature/ex 8103 deprecate old ds tag component

mnavarroespinosa:
EX-8103

Added `relatedTagClass` dynamic prop to `RelatedTags` component to be able to style the `RelatedTag` component.

BREAKING CHANGE: Old `design-system` `Tag` styles are now located in the  `deprecated-full-theme.css` file.

BREAKING CHANGE: Old `x-tag--is-curated` and `x-tag--is-selected` classes are no longer used by `RelatedTag` component.


packages/x-components/src/x-modules/related-tags/components/related-tags.vue
```diff
@@ -60,7 +65,7 @@
     components: { RelatedTag },
     mixins: [xComponentMixin(relatedTagsXModule)]
   })
-  export default class RelatedTags extends Vue {
+  export default class RelatedTags extends Mixins(dynamicPropsMixin(['relatedTagClass'])) {
```
ajperezbau, line 68: What do you think about renaming the `relatedTagClass` to be more generic?
`itemClass` or something like that, that way, if we add the prop to other list components all will have a similar API.

packages/x-components/src/x-modules/related-tags/components/related-tags.vue
```diff
@@ -18,7 +18,11 @@
         @binding {boolean} highlightCurated - True if the curated RTs should be displayed.
        -->
       <slot name="related-tag" v-bind="{ relatedTag, highlightCurated }">
-        <RelatedTag :highlightCurated="highlightCurated" :relatedTag="relatedTag">
+        <RelatedTag
+          :highlightCurated="highlightCurated"
+          :relatedTag="relatedTag"
+          :class="relatedTagClass"
```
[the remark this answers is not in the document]
mnavarroespinosa: 🤦‍♂️ Sure, totally forgot! I've added some.

packages/x-components/src/x-modules/related-tags/components/related-tags.vue
mnavarroespinosa: Good point! I've renamed it to `itemClass`.

packages/x-components/src/views/home/Home.vue
```diff
@@ -138,7 +138,7 @@
               <template #sliding-panel-left-button>
                 <ChevronLeft />
               </template>
-              <RelatedTags class="x-tag--card x-list--gap-03" />
+              <RelatedTags class="x-list--gap-03" relatedTagClass="x-tag-outlined" />
```
ajperezbau, line 141:
```suggestion
              <RelatedTags class="x-list--gap-03" itemClass="x-tag-outlined" />
```

packages/x-components/src/x-modules/related-tags/components/related-tags.vue
```diff
@@ -16,9 +16,14 @@
         @slot Custom content that replaces the RelatedTag component.
         @binding {RelatedTag} relatedTag - Related tag data.
         @binding {boolean} highlightCurated - True if the curated RTs should be displayed.
+        @binding {string} itemClass - Class to add to the related tag.
        -->
-      <slot name="related-tag" v-bind="{ relatedTag, highlightCurated }">
-        <RelatedTag :highlightCurated="highlightCurated" :relatedTag="relatedTag">
+      <slot name="related-tag" v-bind="{ relatedTag, highlightCurated, itemClass }">
```
ajperezbau, line 21: Do we really need to bind `itemClass` here? 🤔 
I think that if the `related-tag` slot is used, there is no need to set the class using the prop.

ajperezbau (changes requested):
Good job 👌

packages/x-components/src/x-modules/related-tags/components/related-tags.vue
tiborux: you are right @ajperezbau because you can "rewrite" the related-tag itself to assign the class

packages/x-components/src/x-modules/related-tags/components/__tests__/related-tags.spec.ts
```diff
@@ -133,8 +143,8 @@ describe('testing related tags component', () => {
     const { getRelatedTagItems, relatedTags, wrapper } = renderRelatedTags({
       template: `
         <RelatedTags>
-          <template #related-tag="{relatedTag, highlightCurated}">
-            <button data-test="custom-related-tag" v-if="highlightCurated">
+          <template #related-tag="{relatedTag, highlightCurated, itemClass }">
```
ajperezbau, line 146: I think you forgot to remove the `itemClass` here.

ajperezbau (changes requested):

ajperezbau (approved):

PR #1104, opened 2023-03-15
feat(design-system): add button group component

ajperezbau:
The color applied in the button group is to have a default color for the separator, although it is not stated in the designs.
Also, the separator should have a vertical margin of 8px instead of 4px.

EX-8059


packages/x-tailwindcss/src/x-tailwind-plugin/components/button-group/separator.ts
```diff
@@ -0,0 +1,19 @@
...
+ * @param helpers - The {@link TailwindHelpers} to generate CSS.
+ * @returns The {@link CssStyleOptions} for the component.
+ */
+export function buttonGroupSeparator(helpers: TailwindHelpers) {
+  const { theme } = helpers;
+
+  return {
+    separator: {
+      borderLeft: '1px solid currentColor',
+      margin: `${theme('spacing.8')} 0`,
```
mnavarroespinosa, line 15: According to design, it should be `4px` of `paddingY` instead of `8px`, isn't it?
```suggestion
      margin: `${theme('spacing.4')} 0`,
```

packages/x-tailwindcss/src/x-tailwind-plugin/components/button-group/default.ts
```diff
@@ -0,0 +1,31 @@
...
+ * @param helpers - The {@link TailwindHelpers} to generate CSS.
+ * @returns The {@link CssStyleOptions} for the component.
+ */
+export function buttonGroupDefault(helpers: TailwindHelpers) {
+  const { theme } = helpers;
+
+  return {
+    display: 'inline-flex',
+    flexFlow: 'row nowrap',
+    color: theme('colors.neutral.50'),
```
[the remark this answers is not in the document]
ajperezbau: Yep, it is to have a default color

packages/x-tailwindcss/src/x-tailwind-plugin/components/button-group/separator.ts
ajperezbau: Sorry about that, I forgot to tell Alicia to update the designs to apply this change and the default color one 🤦 
It is 8px 👌

packages/x-tailwindcss/src/x-tailwind-plugin/components/button-group/default.ts
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * @param helpers - The {@link TailwindHelpers} to generate CSS.
+ * @returns The {@link CssStyleOptions} for the component.
+ */
+export function buttonGroupDefault(helpers: TailwindHelpers) {
+  const { theme } = helpers;
+
+  return {
+    display: 'inline-flex',
+    flexFlow: 'row nowrap',
```
CachedaCodes, line 14: This is the default behavior, do we have to specify it?

packages/x-tailwindcss/src/x-tailwind-plugin/components/button-group/default.ts
```diff
@@ -0,0 +1,31 @@
...
+    color: theme('colors.neutral.50'),
+
+    '&:not([class*="gap"]) > .button + .button': {
```
CachedaCodes, line 17: I don't get why we need the `:not([class*="gap"])`

packages/x-tailwindcss/src/x-tailwind-plugin/components/button-group/separator.ts
CachedaCodes, line 1: The designs say that the separator is not part of the component, was there a change about that decision?

CachedaCodes (changes requested):

packages/x-tailwindcss/src/x-tailwind-plugin/components/button-group/default.ts
ajperezbau: Yep, we can remove it.

packages/x-tailwindcss/src/x-tailwind-plugin/components/button-group/default.ts
ajperezbau: I will leave a comment in the code explaining that.
We remove the left border of the buttons to avoid double borders, which is more evident with the `outlined` variant.
If we apply a `padding` to the `button-group`, we should keep the left border because the buttons will be separated.

packages/x-tailwindcss/src/x-tailwind-plugin/components/button-group/separator.ts
ajperezbau: I think that sentence was about not using a pseudo-element, a border on the buttons, or something like that.
I asked Iván about the separator and it looked ok to him.

CachedaCodes, line 1: Rename to divider instead of separator

CachedaCodes (changes requested):

CachedaCodes (approved):

mnavarroespinosa (approved):

PR #1109, opened 2023-03-20
feat: Minor improvements.

tajespasarela:
Add `itemClass` prop to `ItemsList`.
Add `isHistoryQueriesEnabled` to `AliasApi`.
Add default border width to XDS theme.
Update Tailwind version to 3.3.0

EX-8130

### DIFF
--- a/packages/x-tailwindcss/src/x-tailwind-plugin/theme.ts
+++ b/packages/x-tailwindcss/src/x-tailwind-plugin/theme.ts
@@ -12,6 +12,7 @@ export default {
     full: '99999px'
   },
   borderWidth: {
+    DEFAULT: '1px',
     1: '1px',
     2: '2px',
     4: '4px'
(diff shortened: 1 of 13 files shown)


tiborux (approved):

packages/x-tailwindcss/src/x-tailwind-plugin/theme.ts
```diff
@@ -12,6 +12,7 @@ export default {
     full: '99999px'
   },
   borderWidth: {
+    DEFAULT: '1px',
```
CachedaCodes, line 15: Why the caps here? In aspect ratio we have te default in lowercase

CachedaCodes (approved):

packages/x-tailwindcss/src/x-tailwind-plugin/theme.ts
tajespasarela: The  `DEFAULT` key is to use the utility "without value" which, in this case, is to make the `x-border` work without having to use `x-border x-border-1`.
we can use with colors for example:

```typescript
    neutral: {
      DEFAULT: '#5F717B',
      0: '#FFFFFF',
      10: '#EEF1F2',
      25: '#DBE2E5',
      50: '#5F717B',
      75: '#3C494F',
      90: '#283034',
      100: '#000000'
    },
```

Then we could use `x-bg-neutral` directly without indicating the shade number.


If the key is in lowercase it is like any value, so in the aspect ratio the class is `x-aspect-default`. The class `x-aspect` wouldn't exist.

PR #1110, opened 2023-03-20
feat:(composables): Add `createUseDevice` composable factory.

tajespasarela:
EX-8122

### DIFF
--- a/packages/x-components/src/composables/create-use-device.composable.ts
+++ b/packages/x-components/src/composables/create-use-device.composable.ts
@@ -1,0 +1,206 @@
+import { reduce } from '@empathyco/x-utils';
+import { useWindowSize, useScreenOrientation, useEventListener } from '@vueuse/core';
+import { computed, effectScope, ref, Ref } from 'vue';
+
+/**
+ * The Return type of the composable returned by `createUseDevice`.
+ *
+ * @public
+ */
+export type UseDeviceReturn<Device extends string = string> = {
+  orientation: Ref<'landscape' | 'portrait' | undefined>;
+  isTouchable: Ref<boolean>;
+  deviceName: Ref<string>;
+} & UseDeviceFlags<Device>;
+
+/**
+ * The device flags type of the Return type of the composable returned by `createUseDevice`.
+ *
+ * @public
+ */
+export type UseDeviceFlags<Device extends string> = Record<
+  `is${Capitalize<Device>}`,
+  Ref<boolean>
+> &
+  Record<`is${Capitalize<Device>}OrGreater`, Ref<boolean>> &
+  Record<`is${Capitalize<Device>}OrLess`, Ref<boolean>>;
+
+/**
+ * Factory function that creates a composable for device detection using the devices parameter
+ * to configure breakpoints.
+ *
+ * @param devices - An object containing the breakpoints, where the key is the name of the device
+ * and the value is the screen width.
+ * @returns A composable which provides multiple reactive flags and values for detecting the
+ * current device. The flags names depends on the names passed in the `devices` parameter.
+ * @remarks The `orientation` only works for orientation- sensor devices (mobile, tablet, etc). If
+ * in a desktop, the height of the window is larger than the width, the orientation will be
+ * `landscape`.
+ *
+ * @example
+ * ´´´typescript
+ * const useDevice = createUseDevice(\{ mobile: 0, tablet: 744, desktop: 1024 \});
+ * const \{
+ *   isMobile,
+ *   isMobileOrLess,
+ *   isMobileOrGreater,
+ *   isTablet,
+ *   isTabletOrLess,
+ *   isTabletOrGreater,
+ *   isDesktop,
+ *   isDesktopOrLess,
+ *   isDesktopOrGreater,
+ *   deviceName,
+ *   orientation,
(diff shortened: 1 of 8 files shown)


packages/x-components/src/composables/create-use-device.composable.ts
```diff
@@ -0,0 +1,206 @@
...
+
+/**
+ * Util function to capitalize a string.
+ *
+ * @param str - The string to capitalize.
+ * @returns The string capitalized.
+ *
+ * @internal
+ */
+function capitalize(str: string): string {
```
tiborux, line 204: We already have a util for this inside x-components.

packages/x-components/src/composables/create-use-device.composable.ts
```diff
@@ -0,0 +1,206 @@
...
+
+/**
+ * Factory function that creates a composable for device detection using the devices parameter
+ * to configure breakpoints.
+ *
+ * @param devices - An object containing the breakpoints, where the key is the name of the device
+ * and the value is the screen width.
+ * @returns A composable which provides multiple reactive flags and values for detecting the
+ * current device. The flags names depends on the names passed in the `devices` parameter.
+ * @remarks The `orientation` only works for orientation- sensor devices (mobile, tablet, etc). If
```
CachedaCodes, line 36:
```suggestion
 * @remarks The `orientation` only works for orientation-sensor devices (mobile, tablet, etc). If
```

packages/x-components/tests/unit/useDevice.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+    getDeviceValue(key: string) {
+      return cy.getByDataTest(key);
+    }
+  };
+}
+
+const defaultDevices = { mobile: 0, tablet: 744, desktop: 1200 };
+describe('testing createUseDevice composable factory', () => {
+  it('detects the right device', () => {
+    const { getDeviceValue } = mountTestComponent(defaultDevices);
```
tiborux, line 41: Should we test the orientation? I know that we are using the composable from vue use, but we should check our own values

packages/x-components/src/composables/create-use-device.composable.ts
```diff
@@ -0,0 +1,206 @@
...
+  return reduce(
+    devices,
+    (accumulator, device, deviceWidth) => {
+      const isDevice = computed(
+        () =>
+          deviceWidth <= windowSize.value &&
+          !sortedByWidthDevices.some(
+            ([, otherDeviceWidth]) =>
+              otherDeviceWidth <= windowSize.value && otherDeviceWidth > deviceWidth
+          )
```
CachedaCodes, line 109: Can we extract this to its own function? I feel like `getDeviceFlags` does a lot already

CachedaCodes (changes requested):
My head hurts, gj!

packages/x-components/tests/unit/useDevice.spec.ts
tajespasarela: I would like, and I tried, but unfortunately, this feature only works on actual mobile/tablet/with-gyroscope devices. with cypress, you only can modify the size of the window but it always returns the "landscape" value, even in a "portrait" size.

packages/x-components/src/composables/create-use-device.composable.ts
tajespasarela: I was looking for it in the `x-utils` package. We have to move it there in the near future.

packages/x-components/src/composables/create-use-device.composable.ts
```diff
@@ -0,0 +1,206 @@
...
+ *
+ * @returns A reactive value indicating the current device
+ * orientation.
+ *
+ * @internal
+ */
+function getOrientation(): UseDeviceReturn['orientation'] {
+  const { orientation } = useScreenOrientation();
+  return computed(() => {
+    if (['landscape-primary', 'landscape-secondary'].includes(orientation.value ?? '')) {
```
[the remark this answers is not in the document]
tajespasarela: done

packages/x-components/src/composables/create-use-device.composable.ts
tajespasarela: changed

packages/x-components/src/composables/create-use-device.composable.ts
```diff
@@ -0,0 +1,206 @@
...
+ *
+ * @internal
+ */
+function getDeviceName(
+  devices: Record<string, number>,
+  devicesFlags: UseDeviceFlags<string>
+): Ref<string> {
+  return computed(
+    () =>
+      Object.keys(devices).find(device => devicesFlags[`is${capitalize(device)}` as any]?.value) ??
```
tiborux, line 180: I was trying to remove this any ⚰️

tajespasarela: done

packages/x-components/src/composables/create-use-device.composable.ts
tajespasarela: done

tiborux (approved):
gj

CachedaCodes (approved):

PR #1112, opened 2023-03-22
feat(components): add  `columns` prop in `BaseVariableColumnGrid`

tiborux:
EX-8135

### DIFF
--- a/packages/x-components/src/components/base-variable-column-grid.vue
+++ b/packages/x-components/src/components/base-variable-column-grid.vue
@@ -51,7 +51,8 @@
     protected items?: ListItem[];
 
     /**
-     * The columns to render in the grid.
+     * The columns to render by default in the grid. This property is used when the user has not
+     * selected any value in the column picker.
      *
      * @internal
      */
@@ -59,7 +60,7 @@
     protected columns!: number;
 
     /**
-     * The number of columns provided by the user.
+     * The number of columns provided by a user interaction.
      *
      * @internal
      */
--- a/packages/x-components/src/components/base-variable-column-grid.vue
+++ b/packages/x-components/src/components/base-variable-column-grid.vue
@@ -1,5 +1,5 @@
 <template>
-  <BaseGrid :animation="animation" :columns="columns" :items="items">
+  <BaseGrid :animation="animation" :columns="columnsToRender" :items="items">
     <template v-for="(_, name) in $scopedSlots" v-slot:[name]="{ item }">
       <!--
         @slot Customized item rendering. The slot name can either be default or the item's model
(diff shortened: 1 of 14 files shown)


LuisMartinez15 (approved):
👏🏿 good job 👏🏿

packages/x-components/src/components/base-variable-column-grid.vue
```diff
@@ -55,7 +55,26 @@
...
+
+    /**
+     * The number of columns provided by the user.
+     *
+     * @internal
+     */
+    protected providedColumns = 0;
+
+    /**
+     * The number of columns to render in the grid.
```
CachedaCodes, line 69: `columns` and `columnsToRender` have almost the same description.

packages/x-components/src/components/base-variable-column-grid.vue
```diff
@@ -55,7 +55,26 @@
      *
      * @internal
      */
-    protected columns = 0;
+    @Prop({ default: 0 })
+    protected columns!: number;
+
+    /**
+     * The number of columns provided by the user.
```
CachedaCodes, line 62: Can we change to `provided by a user interaction`? For me, it's clearer that way

CachedaCodes (changes requested):
gj!

packages/x-components/src/components/base-variable-column-grid.vue
```diff
@@ -51,11 +51,31 @@
...
-    protected columns = 0;
+    @Prop({ default: 0 })
+    protected columns!: number;
+
+    /**
+     * The number of columns provided by a user interaction.
+     *
+     * @internal
+     */
+    protected providedColumns = 0;
```
tajespasarela, line 67: We are going to need to provide 0 columns too as is for "automatic" columns number. So no provided columns should be null.

```suggestion
    protected providedColumns: number | null = null;
```

packages/x-components/src/components/base-variable-column-grid.vue
```diff
@@ -51,11 +51,31 @@
...
+
+    /**
+     * The number of columns to render in the grid.
+     *
+     * @returns The number of columns.
+     *
+     * @internal
+     */
+    protected get columnsToRender(): number {
+      return this.providedColumns || this.columns;
```
tajespasarela, line 77:
```suggestion
      return this.providedColumns === null?  this.columns: this.providedColumns;
```

tajespasarela (changes requested):

CachedaCodes (approved):

tajespasarela (approved):

PR #1113, opened 2023-03-23
feat: add sticky filters

diegopf:
EX-8076

### DIFF
--- a/packages/x-components/src/x-modules/facets/service/types.ts
+++ b/packages/x-components/src/x-modules/facets/service/types.ts
@@ -58,6 +58,7 @@ export interface FacetsService {
    * Deselects filter, adding it to the store if it was not present.
    *
    * @param filter - The filter to deselect.
+   * @param metadata - An optional object with the event metadata.
    */
   deselect(filter: Filter, metadata?: FiltersMetadata): void;
   /**
--- a/packages/x-components/src/x-modules/facets/entities/sticky.modifier.ts
+++ b/packages/x-components/src/x-modules/facets/entities/sticky.modifier.ts
@@ -11,6 +11,8 @@ interface Metadata {
 /**
  * Allows to persist a filter between different queries.
  *
+ * @remarks when using this modifier along with others, make sure this is the last one defined.
+ *
  * @internal
  */
 export class StickyModifier extends BaseFilterEntityModifier<Metadata> {


mnavarroespinosa (changes requested):
good job!

packages/x-components/src/main.ts
```diff
@@ -6,10 +6,21 @@ import { baseInstallXOptions, baseSnippetConfig } from './views/base-config';
...
 import { SingleSelectModifier } from './x-modules/facets/entities/single-select.modifier';
+import { StickyModifier } from './x-modules/facets/entities/sticky.modifier';
 import './tailwind/index.css';
 
 Vue.config.productionTip = false;
 FilterEntityFactory.instance.registerModifierByFacetId('age_facet', SingleSelectModifier);
+FilterEntityFactory.instance.registerModifierByFacetId(
+  'brand_facet',
+  SingleSelectModifier,
+  StickyModifier
```
CachedaCodes, line 17: I find it awkward to send a list of modifiers like this instead of with a list, also considering one could override the next. Should we maintain this method signature or change to a list?

packages/x-components/src/x-modules/facets/service/types.ts
```diff
@@ -30,13 +31,18 @@ export interface FacetsService {
    *
    * @param facetIds - An optional list of facets ids from whom deselect the filters.
    */
-  clearFilters(facetIds?: Array<Facet['id']>): void;
+  clearFilters(facetIds?: Array<Facet['id']>, metadata?: Dictionary): void;
```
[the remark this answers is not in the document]
tiborux: Should we change the type of metadata? right now if you wanna send things in the metadata, you have to know the implementation to know that you have to send `keepSticky` to make it sticky because the types are not helping.

Does make it sense to create an interface to define the metadata type?

```
export interface FiltersMetadata {
  keepSticky: boolean;
}
```

CachedaCodes (approved):

packages/x-components/src/x-modules/facets/service/types.ts
```diff
@@ -29,14 +39,27 @@ export interface FacetsService {
...
+    facetIds?: Array<Facet['id']>;
+    metadata?: FiltersMetadata;
+  }): void;
   /**
    * Deselects filter, adding it to the store if it was not present.
    *
    * @param filter - The filter to deselect.
    */
-  deselect(filter: Filter): void;
+  deselect(filter: Filter, metadata?: FiltersMetadata): void;
```
mnavarroespinosa, line 62: `metadata` doc is missing for this method

mnavarroespinosa (approved):

CachedaCodes (approved):

tiborux (approved):

PR #1118, opened 2023-03-30
feat(design-system): add XDS filters

ajperezbau:
The `expanded` option will be developed when it is needed.
It doesn't have an easy solution, because the component should be compatible with the fix to align the icon and the text.
Also, keep in mind that the hierarchical filters are part of the implementation in the X Components, the XDS will not contain anything related to them.

EX-8062


packages/x-tailwindcss/src/x-tailwind-plugin/components/filter/sizes.ts
```diff
@@ -0,0 +1,29 @@
...
+    },
+    md: {
+      minHeight: theme('spacing.32'),
+      gap: theme('spacing.4'),
+      fontSize: theme('fontSize.sm')
+    },
+    lg: {
+      minHeight: theme('spacing.48'),
+      gap: theme('spacing.12'),
+      fontSize: theme('fontSize.md')
```
tiborux, line 26: where did you get this value? I cannot find it

packages/x-tailwindcss/src/x-tailwind-plugin/components/filter/default.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    ...filterSizes(helpers).md,
+    ...alignIconWithBaseline(),
+
+    '&:hover,&.selected': {
+      color: `var(--filter-color-50, ${theme('colors.neutral.50')})`
+    },
+
+    '&.selected': {
+      fontWeight: theme('fontWeight.bold')
```
tiborux, line 41: this font-weight is not being applied, when a default filter is selected the fontWeight is 400 and not 700

CachedaCodes (approved):

packages/x-tailwindcss/src/x-tailwind-plugin/components/filter/sizes.ts
ajperezbau: That wasn't specified in the properties, but was applied in the filter example.
Cris will include it in Figma 👌

diegopf (approved):

PR #1120, opened 2023-03-31
ci: fix cypress cache path

mnavarroespinosa:
EX-8149

This change solves one of the issues related to Cypress in the pipeline:

image

In this case, the problem was caused by a wrong Cypress cache path passed to the `setup pnpm cache` step. 

- We were using the default Cypress cache path (`~/.cache/Cypress`) but it is being overridden by the environment variable `CYPRESS_CACHE_FOLDER` with value `/root/.cache/Cypress` in the docker image. 

> Check [https://github.com/search?q=repo:cypress-io/cypress-docker-images CYPRESS_CACHE_FOLDER&type=code](https://github.com/search?q=repo:cypress-io/cypress-docker-images%20CYPRESS_CACHE_FOLDER&type=code)

- When `pnpm` cache is restored, the `postinstall` of Cypress, which downloads the binary, is not executed and this causes the tests to fail if the binary is not either restored from the cache or downloaded.

- Before this change, the `setup pnpm cache` size was `~107 MB`, now it is `~267 MB`.

Regarding the other issue, I didn't find the cause yet, I suggest fixing it in an independent PR. 

image

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,7 +36,7 @@ jobs:
         with:
           path: |
             ${{ steps.pnpm-store-directory.outputs.PATH }}
-            ~/.cache/Cypress
+            /root/.cache/Cypress
           key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
           restore-keys: |
             ${{ runner.os }}-pnpm-store-
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,19 +24,20 @@ jobs:
         with:
           version: 7
 
-      - name: save pnpm store directory
-        id: pnpm-store-directory
+      - name: setup ENV variables
+        id: setup-env-vars
         shell: bash
         run: |
-          echo "PATH=$(pnpm store path)" >> $GITHUB_OUTPUT
+          echo "CYPRESS_CACHE_FOLDER=$CYPRESS_CACHE_FOLDER" >> $GITHUB_OUTPUT
+          echo "PNPM_STORE_PATH=$(pnpm store path)" >> $GITHUB_OUTPUT
 
       - name: setup pnpm cache
         id: pnpm-cache
         uses: actions/cache@v3
         with:
           path: |
-            ${{ steps.pnpm-store-directory.outputs.PATH }}
-            /root/.cache/Cypress
+            ${{ steps.setup-env-vars.outputs.PNPM_STORE_PATH }}
+            ${{ steps.setup-env-vars.outputs.CYPRESS_CACHE_FOLDER }}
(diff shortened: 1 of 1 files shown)


diegopf (dismissed):

ajperezbau (commented):
What do you think about exposing the Cypress env variable in a step output?
That way, we wouldn't need to update the path if it changes again.

We can do something like:

```
name: Check branch is releasable and release alpha on main branch update
on: [push, pull_request]
concurrency:
  group: ${{ github.workflow }}-${{ github.ref_name }}
  cancel-in-progress: true
jobs:
  build:
    runs-on: ubuntu-latest
    container:
      image: cypress/browsers:node-16.18.1-chrome-110.0.5481.96-1-ff-109.0-edge-110.0.1587.41-1
      # https://github.com/cypress-io/github-action#firefox Cypress FF image needs this user permissions to be able to install dependencies
      options: --user 1001

    steps:
      - name: checkout repository
        uses: actions/checkout@v3
        with:
          fetch-depth: 0
          persist-credentials: false

      # Install pnpm because it is not included in our container image
      - name: install pnpm
        uses: pnpm/action-setup@v2
        with:
          version: 7

      - name: setup ENV variables
        id: setup-env-vars
        run: |
          echo "CYPRESS_CACHE_FOLDER=$CYPRESS_CACHE_FOLDER" >> $GITHUB_OUTPUT |
          echo "PATH=$(pnpm store path)" >> $GITHUB_OUTPUT

      - name: setup pnpm cache
        id: pnpm-cache
        uses: actions/cache@v3
        with:
          path: |
            ${{ steps.setup-env-vars.outputs.PATH }}
            ${{ steps.setup-env-vars.outputs.CYPRESS_CACHE_FOLDER }}
          key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
          restore-keys: |
            ${{ runner.os }}-pnpm-store-

      - name: setup eslint cache
        uses: actions/cache@v3
        with:
          path: '**/.eslintcache'
          key: eslint-${{github.sha}}
          restore-keys: eslint-

      - name: setup lerna cache
        id: lerna-cache
        uses: actions/cache@v3
        with:
          path: |
            node_modules/.cache
          key: lerna-cache-${{github.sha}}
          restore-keys: lerna-cache-

      - name: install dependencies
        run: pnpm install

      - name: lint packages
        run: pnpm run lint

      - name: test packages
        run: pnpm run test

      - name: release alpha on non-release pushes to main branch
        if:
          ${{ github.ref_name == 'main' && github.actor != 'support-empathy' &&
          !startsWith(github.event.head_commit.message, 'chore(release):') }}
        uses: ./.github/actions/release-alpha
        with:
          npm_token: ${{ secrets.NPM_TOKEN }}
          github_token: ${{ secrets.SUPPORT_TOKEN }}
```

diegopf (approved):

ajperezbau (approved):

PR #1121, opened 2023-04-03
feat(design-system): deprecate old DS filters

ajperezbau:
**BREAKING CHANGES:** 
- Renamed `x-filter` to `x-facet-filter`.
- Removed `x-filter--is-disabled` class.
- Renamed `x-filter--is-selected` to `x-selected`.
- Renamed `x-filter--is-partially-selected` to `x-facet-filter--is-partially-selected`.
- Old `design-system` `Filter` styles are now located in the `deprecated-full-theme.css` file.

EX-8152


diegopf (dismissed):

packages/x-components/src/design-system-deprecated/components/filter/default.scss
```diff
@@ -1,4 +1,4 @@
-@import '../../mixins';
+@import '../../../design-system/mixins';
 
 .x-filter {
```
mnavarroespinosa, line 3: Since the `x-components` are not using `x-filter` class anymore, what do you think of renaming the deprecated styles as well to `x-facet-filter`? This way the deprecated styles will be properly applied when used along with the `x-components` filters that, from now on, use the new nomenclature. 

The same would happen with the `x-filter--is-selected` to `x-selected` renaming.

mnavarroespinosa (changes requested):

mnavarroespinosa (approved):
good job!

diegopf (approved):

PR #1122, opened 2023-04-05
feat(search): add fallback disclaimer component

tiborux:
Some e2e tests are failing, I disabled some of them because I think that its the problem that causes the CI stops abruptly and doesn't show the error in the log.

EX-8080

### DIFF
--- a/packages/x-components/src/main.ts
+++ b/packages/x-components/src/main.ts
@@ -13,13 +13,13 @@ Vue.config.productionTip = false;
 FilterEntityFactory.instance.registerModifierByFacetId('age_facet', SingleSelectModifier);
 FilterEntityFactory.instance.registerModifierByFacetId(
   'brand_facet',
-  SingleSelectModifier,
-  StickyModifier
+  StickyModifier,
+  SingleSelectModifier
 );
 FilterEntityFactory.instance.registerModifierByFacetId(
   'price',
-  SingleSelectModifier,
-  StickyModifier
+  StickyModifier,
+  SingleSelectModifier
 );
 FilterEntityFactory.instance.registerModifierByFilterModelName(
   'HierarchicalFilter',
--- a/packages/x-components/src/views/home/Home.vue
+++ b/packages/x-components/src/views/home/Home.vue
@@ -214,7 +214,7 @@
         </template>
 
         <template #main-aside>
-          <Aside v-if="$x.totalResults > 0" />
+          <Aside v-if="!$x.isNoResults || !$x.isNoResultsWithFilters" />
         </template>
 
         <template #main-body>
@@ -244,7 +244,7 @@
 
             <!--  No Results Message  -->
             <div
-              v-if="$x.noResults"
+              v-if="$x.isNoResults"
               class="x-message x-margin--top-03 x-margin--bottom-03"
               data-test="no-results-message"
             >
@@ -351,7 +351,7 @@
             </LocationProvider>
 
             <!-- Partials -->
-            <PartialResultsList :animation="resultsAnimation">
+            <PartialResultsList v-if="$x.isNoResults" :animation="resultsAnimation">
               <template #default="{ partialResult }">
                 <span data-test="partial-query">{{ partialResult.query }}</span>
                 <BaseGrid
--- a/packages/x-components/src/views/home/Home.vue
+++ b/packages/x-components/src/views/home/Home.vue
@@ -428,7 +428,6 @@
   import NextQueriesList from '../../x-modules/next-queries/components/next-queries-list.vue';
   import NextQueries from '../../x-modules/next-queries/components/next-queries.vue';
   import NextQueryPreview from '../../x-modules/next-queries/components/next-query-preview.vue';
-  import { NextQuery } from '../../x-modules/next-queries/index';
   import QueryPreviewList from '../../x-modules/queries-preview/components/query-preview-list.vue';
   import Recommendations from '../../x-modules/recommendations/components/recommendations.vue';
   import RelatedTags from '../../x-modules/related-tags/components/related-tags.vue';
@@ -457,12 +456,12 @@
   import CloseMainModal from '../../components/modals/close-main-modal.vue';
   import { XProvide } from '../../components/decorators/injection.decorators';
   import { adapterConfig } from '../adapter';
+  import NextQuery from '../../x-modules/next-queries/components/next-query.vue';
   import FallbackDisclaimer from '../../x-modules/search/components/fallback-disclaimer.vue';
   import Aside from './aside.vue';
   import PredictiveLayer from './predictive-layer.vue';
   import Result from './result.vue';
   import { HomeControls } from './types';
-  /* eslint-enable max-len */
 
   @Component({
     directives: {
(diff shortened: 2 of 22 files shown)


packages/x-components/src/x-modules/search/store/emitters.ts
```diff
@@ -34,8 +35,11 @@ export const searchEmitters = createStoreEmitters(searchXStoreModule, {
         totalResults: state.totalResults
       };
     },
-    filter: (newValue, oldValue) =>
-      newValue.status !== oldValue.status && oldValue.status === 'loading' && !!newValue.request
+    filter: (newValue, oldValue) => {
+      return (
+        newValue.status === oldValue.status && oldValue.status === 'loading' && !!newValue.request
```
mnavarroespinosa, line 40: Now with this change, we are seeing a no results page momentarily before the filters are removed

https://user-images.githubusercontent.com/77450928/230880476-9572835e-94bf-490a-aa56-14f1e5fa308d.mov

Are we sure we want it to behave this way?

mnavarroespinosa (changes requested):

packages/x-components/src/main.ts
```diff
@@ -13,13 +13,13 @@ Vue.config.productionTip = false;
...
-  StickyModifier
+  StickyModifier,
+  SingleSelectModifier
 );
 FilterEntityFactory.instance.registerModifierByFacetId(
   'price',
-  SingleSelectModifier,
-  StickyModifier
+  StickyModifier,
+  SingleSelectModifier
```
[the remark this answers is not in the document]
ajperezbau: Yep, that was a mistake in the previous PR.
The `SingleSelectModifier` should be the last one, I updated the docs in [this PR](https://github.com/empathyco/x/pull/1121).

ajperezbau (changes requested):

packages/x-components/tests/e2e/fallback-disclaimer.feature
```diff
@@ -0,0 +1,28 @@
+Feature: Fallback disclaimer component
```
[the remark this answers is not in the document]
tiborux: I updated the same scenario to check it at the end.

packages/x-components/src/x-modules/search/components/fallback-disclaimer.vue
```diff
@@ -0,0 +1,70 @@
+<template>
+  <div
+    v-if="$x.isNoResultsWithFilters"
+    class="x-fallback-disclaimer"
+    data-test="fallback-disclaimer"
+  >
+    <slot v-bind="{ query }">
```
ajperezbau, line 7: Should we bind the query from the alias api too?
I think that doing that the component will be more tolerant to changes, for example, if we change the script to the `defineComponent` api, this will continue working.
Something like:

```
    <slot v-bind="{ query: $x.query.search }">
      No results found for {{ $x.query.search }} with the selected filters. The filters have been
      unselected.
    </slot>
```

packages/x-components/src/x-modules/search/components/__tests__/fallback-disclaimer.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+  });
+
+  it('renders a custom fallback disclaimer message', async () => {
+    const { wrapper } = renderFallbackDisclaimer({
+      template: `
+      <FallbackDisclaimer>
+        <template #default="{ query }">
+          No results found for '{{ query }}'. Filters deselected
+        </template>
+      </Fall>
```
ajperezbau, line 63:
```suggestion
      </FallbackDisclaimer>
```

packages/x-components/src/x-modules/search/store/emitters.ts
```diff
@@ -14,6 +14,7 @@ export const searchEmitters = createStoreEmitters(searchXStoreModule, {
       return newValue.length !== 0 || oldValue.length !== 0;
     }
   },
+  NoResultsWithFiltersChanged: state => state.isNoResultsWithFilters,
```
ajperezbau, line 17: I think that we no longer need this emitter, right?

packages/x-components/src/x-modules/search/store/actions/save-search-response.action.ts
ajperezbau, line 1: Should we add some tests in `search/store/__tests__/actions.spec.ts` to check the behaviour of the `isNoResults` and `isNoResultsWithFilters` flags?

packages/x-components/tests/e2e/fallback-disclaimer.feature
```diff
@@ -0,0 +1,46 @@
...
+    And   fallback disclaimer is displayed
+    When  "<query>" is searched
+    Then  related results are displayed
+    And   fallback disclaimer is not displayed
+
+    Examples:
+      | query | query2            | filterNumber | facetName   |
+      | lego  | lego super mario  | 3            | price_facet |
+
+  Scenario Outline: 2. Fallback disclaimer is not displayed when there is no results
```
ajperezbau, line 34: I think the second scenario can be something like:

```
  Scenario Outline: 2. Fallback disclaimer is not displayed when there is no results
    Given a results API with no results
    When  "<query>" is searched
    Then  no results message is displayed
    And   fallback disclaimer is not displayed

    Examples:
      | query |
      | lego  |
```

ajperezbau:
The `no results` message is being rendered with the fallback disclaimer:

https://user-images.githubusercontent.com/75546736/231708398-ed496ab5-f359-491a-a0c4-8aa6d444f075.mov

ajperezbau (changes requested):
Good job! 👌

packages/x-components/src/x-modules/search/store/actions/save-search-response.action.ts
```diff
@@ -23,6 +23,19 @@ export const saveSearchResponse: SearchXStoreModule['actions']['saveSearchRespon
     queryTagging
   }
 ) => {
+  if (state.isNoResultsWithFilters && state.results.length) {
+    commit('setIsNoResultsWithFilters', false);
+  }
```
mnavarroespinosa, line 28: What do you think of moving this block to the request so the `isNoResultsWithFilters` flag is reset right when the request is made?

packages/x-components/src/views/home/Home.vue
```diff
@@ -349,7 +351,10 @@
             </LocationProvider>
 
             <!-- Partials -->
-            <PartialResultsList :animation="resultsAnimation">
+            <PartialResultsList
+              v-if="!$x.isNoResultsWithFilters && ($x.totalResults <= 4 || $x.isNoResults)"
```
mnavarroespinosa, line 355:
```suggestion
              v-if="!$x.fromNoResultsWithFilters && ($x.totalResults <= 4 || $x.isNoResults)"
```

mnavarroespinosa (changes requested):

packages/x-components/src/views/home/Home.vue
ajperezbau: Note that the alias API exposes `noResults` instead of `isNoResults`.

```suggestion
              v-if="!$x.isNoResultsWithFilters && ($x. fromNoResultsWithFilters <= 4 || $x.noResults)"
```

ajperezbau (commented):
Good job! 👌

ajperezbau (approved):
Good job! 👌

PR #1124, opened 2023-04-06
fix(facets): implement the possibility to set a strategy for what filters are sent to the request. leaves-only strategy also implemented

CachedaCodes:
EX-7550

## Motivation and context
Sometimes the backend doesn't handle the hierarchical filters sent as expected and only filters by the parent. This PR implements the posibility of applying a strategy to filter the filters that are sent to the requests or that are stored in the URL 

## Type of change

- [X] `Main`
- [ ] Other. Specify:

## How has this been tested?
The strategy is set to `leaves-only`, which means that only filters without selected children are sent to the request or stored in the URL. Check the request or the url while selecting filters and change the strategy to `all` in the `base-config.ts`.

### DIFF
--- a/packages/x-components/src/x-modules/facets/store/getters/selected-filters-for-request.getter.ts
+++ b/packages/x-components/src/x-modules/facets/store/getters/selected-filters-for-request.getter.ts
@@ -11,18 +11,21 @@ import { FacetsXStoreModule } from '../types';
  *
  * @public
  */
-export const selectedFiltersForRequest: FacetsXStoreModule['getters']['selectedFilters'] = state =>
-  Object.values(state.filters)
-    .filter(filter => filter.selected)
-    .filter((filter, _, filters) => {
-      if (
-        state.config.filtersForRequestStrategy === 'leaves-only' &&
-        isHierarchicalFilter(filter)
-      ) {
-        const childrenIds = filter.children?.map(child => child.id);
+export const selectedFiltersForRequest: FacetsXStoreModule['getters']['selectedFiltersForRequest'] =
+  state => {
+    const selectedFilters = Object.values(state.filters).filter(filter => filter.selected);
 
-        return !filters.some(newFilter => childrenIds?.includes(newFilter.id));
-      }
+    if (state.config.filtersForRequestStrategy === 'leaves-only') {
+      return selectedFilters.filter((filter, _, filters) => {
+        if (isHierarchicalFilter(filter)) {
+          const childrenIds = filter.children?.map(child => child.id);
 
-      return true;
-    });
+          return !filters.some(newFilter => childrenIds?.includes(newFilter.id));
+        }
+
+        return true;
+      });
+    }
+
+    return selectedFilters;
+  };
--- a/packages/x-components/src/x-modules/facets/store/getters/selected-filters-for-request.getter.ts
+++ b/packages/x-components/src/x-modules/facets/store/getters/selected-filters-for-request.getter.ts
@@ -1,0 +1,28 @@
+import { isHierarchicalFilter } from '@empathyco/x-types';
+import { FacetsXStoreModule } from '../types';
+
+/**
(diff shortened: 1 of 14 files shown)


packages/x-components/src/x-modules/facets/store/module.ts
```diff
@@ -17,10 +18,14 @@ export const facetsXStoreModule: FacetsXStoreModule = {
     groups: {},
     facets: {},
     preselectedFilters: [],
-    stickyFilters: {}
+    stickyFilters: {},
+    config: {
+      filtersForRequestStrategy: 'all'
```
diegopf, line 23: Shouldn't we default to the `leaves-only`? 🤔

packages/x-components/src/x-modules/facets/store/getters/selected-filters-for-request.getter.ts
```diff
@@ -0,0 +1,28 @@
...
+ * Default implementation for the {@link FacetsGetters.selectedFiltersForRequest} getter.
+ *
+ * @param state - Current {@link https://vuex.vuejs.org/guide/state.html | state} of the facets'
+ * module.
+ *
+ * @returns A list containing the selected filters that conform to the filters for request strategy.
+ *
+ * @public
+ */
+export const selectedFiltersForRequest: FacetsXStoreModule['getters']['selectedFilters'] = state =>
```
diegopf, line 14:
```suggestion
export const selectedFiltersForRequest: FacetsXStoreModule['getters']['selectedFiltersForRequest'] = state =>
```

packages/x-components/src/x-modules/facets/store/getters/selected-filters-for-request.getter.ts
```diff
@@ -0,0 +1,28 @@
...
+        state.config.filtersForRequestStrategy === 'leaves-only' &&
+        isHierarchicalFilter(filter)
+      ) {
+        const childrenIds = filter.children?.map(child => child.id);
+
+        return !filters.some(newFilter => childrenIds?.includes(newFilter.id));
+      }
+
+      return true;
+    });
```
diegopf, line 28: If we have configured the leaves-only strategy, wouldn't we be performing an unnecessary second filter loop? Perhaps we could consider doing something like

```ts
const selectedFilters = Object.values(state.filters).filter(filter => filter.selected);
    if (state.config.filtersForRequestStrategy === 'leaves-only') {
      return selectedFilters.filter((filter, _, filters) => {
        if (isHierarchicalFilter(filter)) {
          const childrenIds = filter.children?.map(child => child.id);

          return !filters.some(newFilter => childrenIds?.includes(newFilter.id));
        }

        return true;
      });
    }
    return selectedFilters;

**Review by diegopf — CHANGES_REQUESTED** [review]

File: packages/x-components/src/x-modules/facets/store/module.ts (hunk 1 above)
Comment (CachedaCodes) [reply packages/x-components/src/x-modules/facets/store/module.ts:23]: Default behavior for the library is `all` and for our implementation is `leaves-only`.

**Review by diegopf — APPROVED** [review]

File: packages/x-components/src/x-modules/facets/store/types.ts [hunk 4]
```diff
@@ -195,3 +207,20 @@ export interface MutateFilterPayload {
...
 }
+
+/**
+ * Configuration options for the {@link FacetsXModule}.
+ *
+ * @public
+ */
+export interface FacetsConfig {
+  /** The filter strategy to use when providing the selected filters for requests. */
+  filtersForRequestStrategy: FiltersForRequestStrategy;
```
tiborux, line 218: should be `filtersStrategyForRequest`? or `filtersStrategy`? or `filtersStrategyRequest`?

CachedaCodes: I like `filtersStrategyForRequest` better, changing it 👍

tiborux (approved):

tiborux (approved):

PR #1134, opened 2023-04-13
feat (design-system): deprecate old DS list component

ajperezbau:
EMP-133


packages/x-components/src/components/panels/base-tabs-panel.vue
```diff
@@ -287,7 +287,7 @@ behavior is disabled.
     activeTabClass="x-button-auxiliary"
     contentClass="x-padding--04 x-background--auxiliary"
     tabClass="x-button-ghost"
-    tabsListClass="x-list--horizontal"
+    tabListClass="x-flex-col"
```
mnavarroespinosa, line 290: I guess this should be `x-flox-row` but it collides with `x-flex x-flex-col` added in line 5.
```suggestion
    tabListClass="x-flex-row"
```

packages/x-components/src/components/suggestions/base-suggestions.vue
```diff
@@ -166,6 +166,12 @@
 <style lang="scss" scoped>
   .x-suggestions {
     list-style-type: none;
+
+    &,
```
diegopf, line 170:
```suggestion
```

diegopf (changes requested):

mnavarroespinosa (approved):

PR #1140, opened 2023-04-17
feat(design-system): deprecate old DS components

ajperezbau:
- The `list`, `option-list`, and `progress-bar` components will be deprecated in their respective classes.
- To adjust the min-width of the grid items, a CSS var is used, name suggestions are welcome 👀 
- The dropdown component is not used in the setups, so it will not be added to the design system at the moment. 

EMP-133


mnavarroespinosa (approved):
gj!

diegopf (changes requested):

diegopf (approved):

PR #1149, opened 2023-04-19
feat(design-system): add XDS progress bar component

ajperezbau:
EMP-396


packages/x-tailwindcss/demo/src/components/xds-progress-bar.vue
```diff
@@ -0,0 +1,65 @@
+<template>
+  <XdsBaseShowcase
+    title="Progress Bar"
+    :sections="sections"
+    #default="{ cssClass, copyCssClassesToClipboard }"
+  >
+    <div
```
CachedaCodes, line 7: Add text for each variant like we do in the icons and such (like, for labeling it as `lg` or `md` size, the colors being used, etc.)

CachedaCodes (changes requested):

packages/x-tailwindcss/src/x-tailwind-plugin/components/progress-bar/default.ts
```diff
@@ -0,0 +1,28 @@
...
+    borderRadius: theme('borderRadius.sm'),
+    display: 'inline-block',
+    backgroundColor: theme('colors.neutral.25'),
+    overflow: 'hidden',
+    maxWidth: '320px',
+
+    '& > &-line': {
+      height: '100%',
+      backgroundColor: `var(--progress-bar-color-50, ${theme('colors.neutral.90')})`,
+      transformOrigin: 'left'
```
diegopf, line 23:
```suggestion
```

I think we don't have to include the `transformOrigin` in the XDS  cc @mnavarroespinosa

packages/x-components/src/design-system-deprecated/components/progress-bar/progress-bar.scss
diegopf, line 1: We should rename in the deprecated styles the `x-progress-bar-line` that is now used by the component.

Besides, in the deprecated styles the tokens do not have default values, probably we should add the ones that are removed from the component.

diegopf (changes requested):

CachedaCodes (approved):

diegopf (approved):

PR #1174, opened 2023-05-03
ci: fix ci/cd errors

ajperezbau:
EMP-503

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,14 +49,14 @@ jobs:
           key: eslint-${{github.sha}}
           restore-keys: eslint-
 
-      - name: setup lerna cache
-        id: lerna-cache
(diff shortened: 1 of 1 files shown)


packages/x-components/cypress.config.ts
```diff
@@ -31,11 +31,17 @@ export default defineConfig({
     experimentalRunAllSpecs: true
   },
   component: {
+    defaultCommandTimeout: 7000,
+    experimentalSingleTabRunMode: true,
```
ajperezbau, line 35: This flag enables `Cypress` to run all the component tests in a single tab, which comes with better performance and better behavior overall.
It is experimental, but it is working just fine and it seems that, eventually, it will become the default option.

tiborux (approved):

PR #1175, opened 2023-05-04
chore(deps): update node to v18

ajperezbau:
EMP-274


diegopf (approved):
We have to make sure this appears as breaking change

tiborux (approved):

PR #1211, opened 2023-05-24
feat: set configuration for base-result-link to ignore tagging of click

CachedaCodes:
[EMP-864](https://searchbroker.atlassian.net/browse/EMP-864)

## Motivation and context
With the future inclusion of the `displayClick` event we needed a way of ignoring the click event of the base-result-link in the tagging if we decided so. This PR adds a property to the wiring metadata to do that and check it in the module's wiring.

## Type of change

- [X] `Main`
- [ ] Other. Specify:

## How has this been tested?

New tests have been done for the new features of `BaseResultLink`.

[EMP-864]: https://searchbroker.atlassian.net/browse/EMP-864?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/x-components/src/components/result/__tests__/base-result-link.spec.ts
+++ b/packages/x-components/src/components/result/__tests__/base-result-link.spec.ts
@@ -1,10 +1,13 @@
 import { mount, Wrapper } from '@vue/test-utils';
 import Vue from 'vue';
+import { Result } from '@empathyco/x-types';
 import { createResultStub } from '../../../__stubs__/results-stubs.factory';
 import { getDataTestSelector, installNewXPlugin } from '../../../__tests__/utils';
 import { FeatureLocation } from '../../../types/origin';
-import { XEvent } from '../../../wiring/events.types';
+import { XEvent, XEventsTypes } from '../../../wiring/events.types';
 import BaseResultLink from '../base-result-link.vue';
+import { WireMetadata } from '../../../wiring/index';
+import { PropsWithType } from '../../../utils/index';
 
 describe('testing BaseResultLink component', () => {
   const result = createResultStub('Product 001', {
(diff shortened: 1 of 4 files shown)


packages/x-components/src/components/result/base-result-link.vue
```diff
@@ -65,16 +66,32 @@
...
+       * have extra events so this record pairs each event to its metadata.
+       */
+      const resultLinkMetadataPerEvent = inject<
+        Partial<
+          Record<
+            PropsWithType<XEventsTypes, Result>,
+            Omit<WireMetadata, 'moduleName' | 'origin' | 'location'>
+          >
+        >
+      >('resultLinkMetadataPerEvent', {});
```
ajperezbau, line 80: What do you think about adding this injection key to the `injection.consts.ts`?
That way we wouldn't repeat this type when we provide the events.

packages/x-components/src/components/result/__tests__/base-result-link.spec.ts
```diff
@@ -63,9 +66,64 @@ describe('testing BaseResultLink component', () => {
...
+    );
+    resultLinkWrapper.vm.$x.on('UserClickedAResult', true).subscribe(resultClickListener);
+    resultLinkWrapper.vm.$x
+      .on('UserClickedResultAddToCart', true)
+      .subscribe(addToCartClickListener);
+    resultLinkWrapper.trigger('click');
+
+    expect(resultClickListener).toHaveBeenCalledWith({
+      eventPayload: result,
+      metadata: expect.objectContaining(injectedResultLinkMetadataPerEvent.UserClickedAResult)
```
ajperezbau, line 119: I would also test here that the metadata doesn't contain the `UserClickedResultAddToCart` one.

packages/x-components/src/components/result/__tests__/base-result-link.spec.ts
```diff
@@ -63,9 +66,64 @@ describe('testing BaseResultLink component', () => {
...
+        localVue,
+        propsData: { result }
+      }
+    );
+    resultLinkWrapper.vm.$x.on('UserClickedAResult', true).subscribe(resultClickListener);
+    resultLinkWrapper.vm.$x
+      .on('UserClickedResultAddToCart', true)
+      .subscribe(addToCartClickListener);
+    resultLinkWrapper.trigger('click');
+
```
ajperezbau, line 116:
```suggestion
    expect(resultClickListener).toHaveBeenCalledTimes(1);
```

packages/x-components/src/components/result/__tests__/base-result-link.spec.ts
```diff
@@ -63,9 +66,64 @@ describe('testing BaseResultLink component', () => {
...
+    expect(resultClickListener).toHaveBeenCalledWith({
+      eventPayload: result,
+      metadata: expect.objectContaining(injectedResultLinkMetadataPerEvent.UserClickedAResult)
+    });
+    expect(addToCartClickListener).toHaveBeenCalledWith({
```
ajperezbau, line 121:
```suggestion
    expect(addToCartClickListener).toHaveBeenCalledTimes(1);
    expect(addToCartClickListener).toHaveBeenCalledWith({
```

packages/x-components/src/components/result/__tests__/base-result-link.spec.ts
```diff
@@ -63,9 +66,64 @@ describe('testing BaseResultLink component', () => {
...
+      eventPayload: result,
+      metadata: expect.objectContaining(
+        injectedResultLinkMetadataPerEvent.UserClickedResultAddToCart
```
ajperezbau, line 124: I would also test here that the metadata doesn't contain the `UserClickedAResult` one.

packages/x-components/src/x-modules/tagging/wiring.ts
```diff
@@ -161,7 +161,8 @@ export function createTrackWire(property: keyof Tagging): Wire<Taggable> {
       taggingInfo.params.location = location as string;
       return taggingInfo;
     }),
-    ({ eventPayload: { tagging } }) => !!tagging?.[property]
+    ({ eventPayload: { tagging }, metadata: { ignoreInModules } }) =>
+      !!tagging?.[property] && !ignoreInModules?.includes(moduleName)
```
ajperezbau, line 165: Should we do this in the `wires.factory.ts` to make it compatible with all the modules?
Maybe we can explore it in a future refactor.

ajperezbau (commented):
Good job! 👌

packages/x-components/src/x-modules/tagging/wiring.ts
CachedaCodes: I thought about it but right now is better to do it in a case by case basis for now. We would add an if statement to all the wires and only one will use it. Better keeping it in mind for the future if we have more cases.

packages/x-components/src/components/result/base-result-link.vue
CachedaCodes: I think those only work for the `XInject` and we're not using them anymore in `2.7` so I don't know if we can use those keys in the native `inject`.

PR #1242, opened 2023-06-23
chore: add test-unit script

ajperezbau:
Instead of renaming the `test` script to `test:unit`, I have added a new script.
I made it that way to be able to run all the tests of the packages running `npm run test` from the root of the monorepo.
In almost all the packages, the `test:unit` and the `test` script are the same because the don't contain e2e tests.


lauramargar (approved):
Great job!

PR #1243, opened 2023-07-04
fix(filters): Editable number range filter should be unique

annacv:
[EMP-1228](https://searchbroker.atlassian.net/browse/EMP-1228)

## Motivation and context
Make this filter unique, so only one range is applied if we set some ranges consecutively.

- [x] Open issue. If applicable, link:

## Type of change
- [x] Change requires a documentation update

- [x] `Main`
- [ ] Other. Specify:

## How has this been tested?
- Setting a range and editing it: it should only show one filter as selected.
- Setting a range and also selecting other filters: clicking the clear filters button should remove them all from the selected filters list.
- Setting a range and deselecting it by clicking on it: the filter should be removed from the selected filters list.

# Test with Tous endpoint
1. Remove the built-in `salePrice` facet from the `aside.vue` script
2. Replace the snippet config in `base-config.ts` with:
  ```
    instance: 'tous',
    scope: 'desktop',
    env: 'staging',
    lang: 'es',
    uiLang: 'es',
    currency: 'EUR',
    consent: true,
    documentDirection: 'ltr',
    store: 'es',
  ```
4. Ad this override to the `adapter.ts`:
```
const getFacetConfigWithEditable = (facetId: string, type: PlatformFacetType): FacetConfig =>
{
  if(facetId === 'salePrice') {
    let salePriceConfig = getFacetConfig('editable-range');
    const getSalePriceFilterId = (context?: MapperContext) => (context?.requestParameters?.filter as string[])?.find(filter => filter.includes('salePrice'));

    salePriceConfig.schema = {
      ...salePriceConfig.schema,
      id: ({filter}, $context) => {
        const currentSalePriceFilterId = getSalePriceFilterId($context);

        return currentSalePriceFilterId ?? filter;
      },
      range: ({value}, $context) => {
        const currentSalePriceFilterId = getSalePriceFilterId($context);

        let stringRage = currentSalePriceFilterId?.split(':')[1] ?? value;

        const min = Number(stringRage.split('-')[0]);
        const max = Number(stringRage.split('-')[1]);

        return {
          min: Number.isNaN(min) ? null : min,
          max: Number.isNaN(max) ? null : max
        }
      }
    }

    return salePriceConfig;
  }

  return getFacetConfig(type);
}

facetSchema.$override<PlatformFacet, Partial<NumberRangeFacet>>({
  modelName: ({ type, facet }) => getFacetConfigWithEditable(facet, type).modelName as any,
  filters: {
    $path: 'values',
    $subSchema: ({ type, facet }) => getFacetConfigWithEditable(facet, type).schema,
    $context: {
      test: 'test'
    }
  }
});
```
5. Test the facet

[EMP-1228]: https://searchbroker.atlassian.net/browse/EMP-1228?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/x-components/src/views/home/aside.vue
+++ b/packages/x-components/src/views/home/aside.vue
@@ -4,7 +4,11 @@
     <ClearFilters />
     <SelectedFiltersList>
       <template #default="{ filter }">
-        <SimpleFilter :filter="filter" class="x-facet-filter-success" />
+        <SimpleFilter :filter="filter" class="x-facet-filter-success">
+          <template #label>
+            {{ filter.label ?? `${filter.range.min} - ${filter.range.max} ${filter.id}` }}
+          </template>
+        </SimpleFilter>
       </template>
     </SelectedFiltersList>
 
@@ -105,6 +109,8 @@
               </FiltersSearch>
             </SortedFilters>
           </ExcludeFiltersWithNoResults>
+          <!--  Editable Number Price Range Facet    -->
+          <EditableNumberPriceRangeFilter :filter="editableFilter" />
         </BaseHeaderTogglePanel>
       </template>
     </Facets>
@@ -112,11 +118,17 @@
 </template>
 
 <script lang="ts">
-  import { Facet, SimpleFilter as SimpleFilterModel } from '@empathyco/x-types';
+  import {
+    EditableNumberRangeFilter,
+    Facet,
+    SimpleFilter as SimpleFilterModel
+  } from '@empathyco/x-types';
   import Vue from 'vue';
   import { Component } from 'vue-property-decorator';
   import { XInject } from '../../components';
   import BasePriceFilterLabel from '../../components/filters/labels/base-price-filter-label.vue';
+  // eslint-disable-next-line max-len
+  import EditableNumberPriceRangeFilter from '../../x-modules/facets/components/filters/editable-number-range-filter.vue';
   import ChevronDown from '../../components/icons/chevron-down.vue';
   import BaseHeaderTogglePanel from '../../components/panels/base-header-toggle-panel.vue';
   import ClearFilters from '../../x-modules/facets/components/clear-filters.vue';
@@ -142,6 +154,7 @@
       BasePriceFilterLabel,
       ChevronDown,
       ClearFilters,
+      EditableNumberPriceRangeFilter,
       ExcludeFiltersWithNoResults,
       Facets,
       FacetsProvider,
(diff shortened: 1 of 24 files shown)


packages/x-components/src/views/home/aside.vue
```diff
@@ -6,7 +6,7 @@
       <template #default="{ filter }">
         <SimpleFilter :filter="filter" class="x-facet-filter-success">
           <template #label>
-            {{ filter.label ?? `${filter.range.min} - ${filter.range.max}` }}
+            {{ filter.label ?? filter.id }}
```
annacv, line 9: The id is set as an alternative label. If the min/max range were used, when the page is reloaded or we perform another search, the filter's getter won't keep that values as the filter model changes to a RawFilter.

mavmaf (changes requested):
Please take a look to the aside menu: the range is added to filters which I believe don't belongs to (brand one, for instance). 
Also, with current code, simple-filters E2E fails 🤔 
image

CachedaCodes (changes requested):
gj!

lauramargar (dismissed):
🎉

lauramargar (approved):

mavmaf (approved):

annacv (commented):
Hi! Good job!!!

I have been testing and found a use case where the `RangeFilter` behaves differently than other filters when we reload the page. This could be held in another task: 

- Test 1 --> [Hierarchical Filter Test](http://localhost:8080/?filter=promotion%3Ablack%20friday&q=bolso&scroll=195890681): Select a Hierarchical Filter and reload the page, filter is kept in the URL and the results view is rendered with the filter selected
- Test 2 --> [Number Range Filter Test](http://localhost:8080/?filter=price%3A0-10&q=bolso): Select a Number Range Filter and reload the page, filter is kept in the URL, but the results view is not loaded
- Test 3 --> [Editable Number Range Filter Test](http://localhost:8080/?filter=salePrice%3A1-200&q=collar): Select an Editable Number Range Filter and reload the page, filter is kept in the URL and the results view is kept with the filter selected
- Test 4 --> [Hierarchical + Editable Number Range test](http://localhost:8080/?filter=promotion%3Arebajas&filter=salePrice%3A2-200&q=collar): Select Select a Hierarchical Filter and an Editable Number Range Filter and reload the page, filters are kept in the URL and the results view is rendered with the filters selected
- Test 5 --> [Several filter types test](http://localhost:8080/?filter=promotion%3Arebajas&filter=price%3A0-10&filter=salePrice%3A2-200&q=collar): Select Select a Hierarchical Filter, a Number Range Filter, and an Editable Number Range Filter and reload the page, filters are kept in the URL, but the results view is not loaded

PR #1258, opened 2023-08-02
feat: injector utility for the archetype and build config to override the one in X

CachedaCodes:
[EMP-1613](https://searchbroker.atlassian.net/browse/EMP-1613)

## Motivation and context
We needed to encapsulate the styles in the x-archetype overlay from the clients' page, so we needed a way to tell to the styles where they should be injected (by default they're injected at the head of the page).
The injector in this PR does that, along with the rollup config (the implementation would only work on live envs)

## Type of change

- [X] `Main`
- [ ] Other. Specify:

## How has this been tested?
Run `build` on the `x-archetype-utils` package and the import the build in the `x-archetype` dependency.
Follow this draft in local to see it running:
[soon]

### DIFF
--- a/packages/x-components/build/rollup.config.ts
+++ b/packages/x-components/build/rollup.config.ts
@@ -104,6 +104,7 @@ export const rollupConfig = createRollupOptions({
            })();
            const useBrowserInjector =
              (typeof STRIP_SSR_INJECTOR !== 'undefined' && STRIP_SSR_INJECTOR) || isBrowser;
+           console.log('${normalizePath(id)}',{source:${varname}})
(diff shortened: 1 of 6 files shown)


packages/x-components/build/rollup.config.ts
```diff
@@ -104,6 +104,7 @@ export const rollupConfig = createRollupOptions({
            })();
            const useBrowserInjector =
              (typeof STRIP_SSR_INJECTOR !== 'undefined' && STRIP_SSR_INJECTOR) || isBrowser;
+           console.log('${normalizePath(id)}',{source:${varname}})
```
annacv, line 107: 👀 `console.log` should be there?

PR #1264, opened 2023-08-23
fix: scroll behaviour when at the end of scroll and dynamic height

lauramargar:
# Pull request template
 
The behavior of the scroll module when the scroll reaches the end and the height of the container with the scroll changes is not correct. Fix the behavior so the property hasReachedEnd doesn’t change when the scroll is at the end. We have solved it using throttles to wait until `clientHeight` param, stops changing.

## Motivation and context

- [X] Open issue. If applicable, link: [EMP-1050](https://searchbroker.atlassian.net/browse/EMP-1050)

## Type of change

- [X] `Main`
- [ ] Other. Specify:

## How has this been tested?

Go to the x-archetype project, remove from the `has-scroll-past-threshold.mixin.ts` the hasAlmostReachedEnd if statement in `updateHasScrolledPastThreshold` method. the code must seem like this:

```
    if (mainScrollData?.hasReachedStart) {
      this.hasScrolledPastThresholdFlag = false;
      return;
    } else {
      this.hasScrolledPastThresholdFlag = true;
    }
```

Steps to follow:

- Run the build in X project. This is going to generate a local file in src folder. Example: _empathyco-x-components-3.0.0-alpha.400.tgz_

- In the project that you want to test, go to the package.json and in the "install:local" script paste the local build file and run it. 

- Then search for a query with few results. Scroll to the bottom.

## Checklist:

[EMP-1050]: https://searchbroker.atlassian.net/browse/EMP-1050?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/x-components/src/components/scroll/scroll.mixin.ts
+++ b/packages/x-components/src/components/scroll/scroll.mixin.ts
@@ -99,12 +99,22 @@ export default class ScrollMixin extends Vue {
    */
   protected scrollHeight = 0;
 
-  /**
-   * Flag to stop watchers until clientHeight property stops changing.
-   *
-   * @internal
-   */
-  protected watchersFlag = false;
+  protected clientHeightChangingFlag = false;
+
+  protected restoreClientHeightFlag() {
+    this.clientHeightChangingFlag = false;
+    this.storeScrollData();
+    this.previousPosition = this.currentPosition;
+  }
+
+  protected throtteledCall = throttle(this.restoreClientHeightFlag, 100);
+
+  @Watch('clientHeight', { immediate: true })
+  protected onClientHeightChanged() {
+    this.clientHeightChangingFlag = true;
+
+    this.throtteledCall();
+  }
 
   /**
    * Throttled version of the function that stores the DOM scroll related properties.
@@ -115,8 +125,7 @@ export default class ScrollMixin extends Vue {
    * @internal
    */
   protected get throttledStoreScrollData(): () => void {
-    this.watchersFlag = false;
-    return throttle(this.storeScrollData, 300);
+    return throttle(this.storeScrollData, this.throttleMs);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


packages/x-components/src/components/scroll/scroll.mixin.ts
```diff
@@ -98,6 +98,31 @@ export default class ScrollMixin extends Vue {
...
+   * Also sets a new previous position, before update the current one.
+   *
+   * @internal
+   */
+  protected restoreClientHeightFlag(): void {
+    this.isClientHeightChanging = false;
+    this.previousPosition = this.currentPosition;
+  }
+
+  protected throtteledCall = throttle(this.restoreClientHeightFlag, 100);
```
annacv, line 125: Can we use here `this.throttleMs` prop instead of '100' or is it not worth it?

annacv (commented):
gj!!

Maybe we can take the opportunity to make a couple of fixes to the file documentation:
* Line 31: Positive vertical distance to still consider that the **an** element is the first one visible. (rm 'an')
* Line 140 / 140: Returns distance missing to **end position position.**
@returns A number for knowing the distance missing to **end position position.** (maybe endPosition position would be clearer)

packages/x-components/src/components/scroll/scroll.mixin.ts
lauramargar: They are different `throttles`, so I don't know if it's correct to use the same time. Maybe in the future it could change?

annacv: That's ok!! Thks, I wasn't sure about that!

annacv (approved):

PR #1266, opened 2023-08-28
feat: prevent special characters from url

lauramargar:
# Pull request template
 
To prevent a Cross-site Scripting  attack, we must be careful with the characters we allow to type a query. We already have a function to avoid that when a user types in the search box. But we also need to check if there are any special characters when a query is typed through the URL.

## Motivation and context

- [X] Open issue. If applicable, link: [EMP-1745](https://searchbroker.atlassian.net/browse/EMP-1745)

## Type of change

- [X] `Main`
- [ ] Other. Specify:

## How has this been tested?

Try to type a special character (`<>`) in the URL query.

## Checklist:

[EMP-1745]: https://searchbroker.atlassian.net/browse/EMP-1745?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/x-components/src/x-modules/url/components/url-handler.vue
+++ b/packages/x-components/src/x-modules/url/components/url-handler.vue
@@ -68,6 +68,13 @@
     protected isPagePersisted = false;
 
     /**
+     * Flag to know if the query from the url has special characters.
+     *
+     * @internal
+     */
+    protected hasSpecialKeys = false;
+
+    /**
      * Computed to know which params we must get from URL. It gets the params names from the initial
      * state, to get all default params names, and also from the `$attrs` to get the extra params
      * names to take into account.
@@ -156,13 +163,28 @@
     protected emitEvents(): void {
       const { all, extra } = this.parseUrlParams();
       const metadata = this.createWireMetadata();
-      this.$x.emit('ParamsLoadedFromUrl', all, metadata);
+      this.preventSpecialKey(all.query);
+      if (!this.hasSpecialKeys) {
+        this.$x.emit('ParamsLoadedFromUrl', all, metadata);
+      }
       this.$x.emit('ExtraParamsLoadedFromUrl', extra, metadata);
       // TODO: Move this logic from here.
       if (all.query) {
         this.$x.emit('UserOpenXProgrammatically', undefined, metadata);
       }
       this.urlLoaded = true;
+    }
+
+    /**
+     * Prevents the user from either typing or pasting special characters in the url query.
+     *
+     * @internal
+     * @param query - Query from the url that will be checked for special characters.
+     */
+    protected preventSpecialKey(query: string): void {
+      if (/[<>]/.test(query ?? '')) {
+        this.hasSpecialKeys = true;
+      }
     }
 
     /**
--- a/packages/x-components/src/x-modules/url/components/url-handler.vue
+++ b/packages/x-components/src/x-modules/url/components/url-handler.vue
@@ -174,7 +174,13 @@
      */
     protected hasSpecialKeys(urlValue: string): string {
       if (/[<>]/.test(urlValue ?? '')) {
-        return urlValue.replace(/<.*>/g, '');
+        if (!/</.test(urlValue ?? '')) {
+          return urlValue.replace(/.*>/g, '');
+        } else if (!/>/.test(urlValue ?? '')) {
+          return urlValue.replace(/<.*/g, '');
+        } else {
+          return urlValue.replace(/<.*>/g, '');
+        }
       } else {
         return urlValue;
       }
--- a/packages/x-components/src/x-modules/url/components/url-handler.vue
+++ b/packages/x-components/src/x-modules/url/components/url-handler.vue
@@ -166,13 +166,13 @@
     }
 
     /**
-     * Detects if the user typing or pasting special/forbidden characters in some URL param.
+     * Detects if the user typing or pasting special/forbidden characters in some URL param and removes them.
      *
      * @internal
      * @param urlValue - Param from the url that will be checked for special characters.
      * @returns Param without special characters.
      */
-    protected hasSpecialKeys(urlValue: string): string {
+    protected removeSpecialKeys(urlValue: string): string {
       if (/[<>]/.test(urlValue ?? '')) {
         let value = urlValue.replace(/<.*>/g, '');
         if (!/</.test(value ?? '')) {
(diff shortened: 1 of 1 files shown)


packages/x-components/src/x-modules/url/components/url-handler.vue
```diff
@@ -156,7 +163,10 @@
     protected emitEvents(): void {
       const { all, extra } = this.parseUrlParams();
       const metadata = this.createWireMetadata();
-      this.$x.emit('ParamsLoadedFromUrl', all, metadata);
+      this.preventSpecialKey(all.query);
+      if (!this.hasSpecialKeys) {
```
herrardo, line 167: Just curious, I think I like the approach, but why are we preventing to emit the event if we have special keys instead of removing them, and send send it without them?

packages/x-components/src/x-modules/url/components/url-handler.vue
```diff
@@ -165,6 +175,18 @@
       this.urlLoaded = true;
     }
 
+    /**
+     * Prevents the user from either typing or pasting special characters in the url query.
```
herrardo, line 179: I would say:

```suggestion
     * Detects if the user is typing or pasting special/forbidden characters in the URL query
```

packages/x-components/src/x-modules/url/components/url-handler.vue
```diff
@@ -165,6 +175,18 @@
...
+     *
+     * @internal
+     * @param query - Query from the url that will be checked for special characters.
+     */
+    protected preventSpecialKey(query: string): void {
+      if (/[<>]/.test(query ?? '')) {
+        this.hasSpecialKeys = true;
```
herrardo, line 186: Why storing this value in `hasSpecialKeys` protected property instead of returning true/false in this function and then test it on line 167. I mean, I would change this function name to `hasSpecialKeys` and then test:

`if (!this.hasSpecialKeys(this query)) {`

herrardo (changes requested):

packages/x-components/src/x-modules/url/components/url-handler.vue
herrardo, line 167:
```suggestion
      if (this.hasSpecialKeys(all.query)) {
```

packages/x-components/src/x-modules/url/components/url-handler.vue
```diff
@@ -165,6 +175,18 @@
...
+      }
```
herrardo, line 187:
```suggestion
    protected hasSpecialKeys(query: string): void {
      return (/[<>]/.test(query ?? '')) 
```

herrardo (changes requested):

packages/x-components/src/x-modules/url/components/url-handler.vue
lauramargar: Do you prefer to remove the special keys instead of the whole query? I mean, if you have this query `round <hola> shirt`, it would be transformed into `round shirt` and sent into the `ParamsLoadedFromUrl`event.

I use this approach because, in the task description, I understood that we have to clean/remove the whole query. But for me makes more sense to remove only the special characters and the string inside them.

packages/x-components/src/x-modules/url/components/url-handler.vue
```diff
@@ -165,6 +165,21 @@
...
+    /**
+     * Detects if the user typing or pasting special/forbidden characters in some URL param.
+     *
+     * @internal
+     * @param urlValue - Param from the url that will be checked for special characters.
+     * @returns Param without special characters.
+     */
+    protected hasSpecialKeys(urlValue: string): string {
+      if (/[<>]/.test(urlValue ?? '')) {
+        return urlValue.replace(/<.*>/g, '');
```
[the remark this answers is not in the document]
herrardo: What this CodeQL advice is warning is that given some combination of the urlValue, as `<script <script> >whatever<>` there is some way to a `<script` string be the result after the replace... but I don't find an example of that.

packages/x-components/src/x-modules/url/components/url-handler.vue
```diff
@@ -165,6 +165,27 @@
...
     }
 
+    /**
+     * Detects if the user typing or pasting special/forbidden characters in some URL param.
+     *
+     * @internal
+     * @param urlValue - Param from the url that will be checked for special characters.
+     * @returns Param without special characters.
+     */
+    protected hasSpecialKeys(urlValue: string): string {
```
herrardo, line 175: 2 days ago this function was testing if the URL had special keys to not emit the event.

Now the function has changed to replace the characters...so I would change to this:

```suggestion
    /**
     * Detects if the user typing or pasting special/forbidden characters in some URL param and removes them
     *
     * @internal
     * @param urlValue - Param from the url that will be checked for special characters.
     * @returns Param without special characters.
     */
    protected removeSpecialKeys(urlValue: string): string {
```

herrardo (changes requested):

annacv:
Hi! I will add this reference here, maybe it would be worth to give it a chance: https://github.com/braintree/sanitize-url/blob/main/src/index.ts

packages/x-components/src/x-modules/url/components/url-handler.vue
```diff
@@ -255,10 +256,13 @@
           const urlKey = this.getUrlKey(name);
           if (urlSearchParams.has(urlKey)) {
             if (name in initialUrlState) {
-              const urlValue = urlSearchParams.getAll(urlKey);
+              let urlValue = urlSearchParams.getAll(urlKey);
+              urlValue[0] = sanitize(urlValue[0]);
```
diegopf, line 260: To be safe we should sanitize all values and not just the first one. The query only has one occurrence but potentially any parameter could be malicious.

diegopf (changes requested):

packages/x-components/src/x-modules/url/components/url-handler.vue
lauramargar: `urlValue` is not only `query` param, all URL params are passing one by one. `[0]` is necessary because urlValue is an array of only one element, the param that is being checked.

For example, try to search for a query and select a filter, then write the special characters (`<>`) in the filter and in the query.

diegopf: Sorry,  I think I did not explain properly 😅 What I meant is that a url param can have more than one occurrence. Think about filters, you can select more than one filter at the same time. In that scenario you get an array with multiple entries like:

`['categoryIds:12fad53d7', 'price:1600.0-2000.0']` 

so if you only pick the first element, you will fail to sanitize the other values.

herrardo:
This PR is being closed as we decided that we will solve this problem in the Archetype, not letting certain information to be applied to the website through `v-html` directives

PR #1267, opened 2023-08-28
Fix Brand Recommendations animation on second time appearance

annacv:
[EMP-1707](https://searchbroker.atlassian.net/browse/EMP-1707)

# Pull request template
 
Remove all the `QueriesPreview` from the state when the component is destroyed. This way, we prevent re-rendering effects when we go back to a pre-search scenario after selecting & removing a query from the search box.

There were other two options available to handle this: using `v-show` instead of `v-if`, or preventing sending the requests if there were `QueryPreviews` on the state. These options had the con of losing the DOM animation, so I went for this removing queries approach from the state option in order to keep the current component rendering behavior.

## Motivation and context

When writing a query and then deleting it, the BrandRecommendations/QueryPreview component is displayed again, making like two times the appearance animation.

- [x] Open issue. If applicable, link: [EMP-1707](https://searchbroker.atlassian.net/browse/EMP-1707)

## Type of change

- [x] `Main`
- [ ] Other. Specify:

## How has this been tested?

- Choose a query and delete it, query preview carrousel should be shown at one. The test can be performed by setting a  slower Network value in Chrome devTools to be able to check it more accurately.
- Additionally, the project can be built and tested in archetype using the install-local command with this build: `"install:local": "npm install file:../x/packages/x-components/empathyco-x-components-3.0.0-alpha.406.tgz --no-save",` 

[EMP-1707]: https://searchbroker.atlassian.net/browse/EMP-1707?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EMP-1707]: https://searchbroker.atlassian.net/browse/EMP-1707?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
+++ b/packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
@@ -37,6 +37,8 @@
   import { Dictionary } from '@empathyco/x-utils';
   import { Result, SearchRequest } from '@empathyco/x-types';
   import { State } from '../../../components/decorators/store.decorators';
+  import { LIST_ITEMS_KEY } from '../../../components/decorators/injection.consts';
+  import { XProvide } from '../../../components/decorators/injection.decorators';
   import { xComponentMixin } from '../../../components/x-component.mixin';
   import { NoElement } from '../../../components/no-element';
   import { RequestStatus } from '../../../store';
@@ -47,8 +49,6 @@
   import { createOrigin } from '../../../utils/origin';
   import { debounce } from '../../../utils/debounce';
   import { DebouncedFunction } from '../../../utils';
-  import { LIST_ITEMS_KEY, XProvide } from '../../../components';
-
   /**
    * Retrieves a preview of the results of a query and exposes them in the default slot,
    * along with the query preview and the totalResults of the search request.
--- a/packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
+++ b/packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
@@ -98,6 +98,15 @@
     public debounceTimeMs!: number;
 
     /**
+     * Controls whether the QueryPreview should be removed from the state
+     * when the component is destroyed.
+     *
+     * @public
+     */
+    @Prop({ default: true })
+    public clearOnDestroy!: boolean;
+
+    /**
      * The results preview of the queries preview mounted.
      * It is a dictionary, indexed by the query preview query.
      */
@@ -204,11 +213,17 @@
     /**
      * Cancels the (remaining) requests when the component is destroyed
      * via the `debounce.cancel()` method.
+     * If the prop 'clearOnDestroy' is set to true, it also removes the QueryPreview
+     * from the state when the component is destroyed.
      *
      * @internal
      */
     protected beforeDestroy(): void {
       this.emitQueryPreviewRequestUpdated.cancel();
+
+      if (this.clearOnDestroy) {
+        this.$x.emit('QueryPreviewUnmountedHook', this.query, { priority: 0, replaceable: false });
+      }
     }
 
     /**
--- a/packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
+++ b/packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
@@ -212,6 +212,16 @@
     }
 
     /**
+     * Removes all the QueriesPreview from the state when the component is destroyed.
+     * This will prevent re-rendering effects when a query is selected and then deleted from the searchbox.
+     *
+     * @internal
+     */
+    protected destroyed(): void {
+      this.$x.emit('QueriesPreviewDestroyedHook');
+    }
+
+    /**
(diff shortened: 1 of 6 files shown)


packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
```diff
@@ -236,8 +219,8 @@
      */
     @Watch('queryPreviewResults.status')
     emitLoad(status: RequestStatus | undefined): void {
-      if (status === 'success') {
-        this.$emit(this.results?.length ? 'load' : 'error', this.query);
```
annacv, line 240: No need to check the length of the results here, as this scenario is handled in the `queryPreviewResults` getter

herrardo: There is something that feels not right in this approach

CachedaCodes: The `load` event should be emitted when the results are loaded, not when they're loading

image

In this code you're signaling `QueryPreviewList` to flag the results for a query as loaded when they're not.

annacv: TRUE

annacv:
Hi @herrardo & @CachedaCodes here is another approach that makes more sense. Updated PR description 👀 . Let me know what you think, please.

packages/x-components/src/x-modules/queries-preview/wiring.ts
```diff
@@ -59,6 +67,9 @@ export const queriesPreviewWiring = createWiring({
   QueryPreviewUnmountedHook: {
     clearQueryPreviewWire
   },
+  QueriesPreviewDestroyedHook: {
```
CachedaCodes, line 70: The event `QueryPreviewUnmountedHook` is already there to do the same that `QueriesPreviewDestroyedHook`

packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
```diff
@@ -212,6 +211,16 @@
...
 
+    /**
+     * Removes all the QueriesPreview from the state when the component is destroyed.
+     * This will prevent re-rendering effects when a query is selected and then deleted from the searchbox.
+     *
+     * @internal
+     */
+    protected destroyed(): void {
+      this.$x.emit('QueriesPreviewDestroyedHook');
+    }
```
CachedaCodes, line 222: Emit `QueryPreviewUnmountedHook` when the component unmounts instead.

CachedaCodes (changes requested):

packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
annacv: Hi! I have tried that option and it ends up re-rendering the second carousel instead of the first. It only worked adequately removing all the list from the state. 
I have changed to emit the event in the QueriesPreviewList instead of the QueryPreview, as I think it makes more sense.

packages/x-components/src/x-modules/queries-preview/store/module.ts
```diff
@@ -22,6 +22,9 @@ export const queriesPreviewXStoreModule: QueriesPreviewXStoreModule = {
     clearQueryPreview(state, query) {
       Vue.delete(state.queriesPreview, query);
     },
+    removeQueriesPreview(state) {
+      state.queriesPreview = {};
```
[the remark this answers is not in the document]
annacv: I would say that we need to remove all. The re-rendering occurs because we load already loaded queries in the state. If we only delete one (the first), then de "duplicated info" happens in the second carousel (it appeared twice). 
Well, that's what I have observed.

packages/x-components/src/x-modules/queries-preview/events.types.ts
```diff
@@ -17,4 +17,8 @@ export interface QueriesPreviewXEvents {
    * Payload: The query whose preview has been removed.
    */
   QueryPreviewUnmountedHook: string;
+  /**
+   * The component that shows a Query preview has been destroyed.
+   */
+  QueriesPreviewDestroyedHook: void;
```
CachedaCodes, line 23: Remove this event

packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
```diff
@@ -49,6 +47,7 @@
   import { createOrigin } from '../../../utils/origin';
   import { debounce } from '../../../utils/debounce';
   import { DebouncedFunction } from '../../../utils';
+  import { LIST_ITEMS_KEY, XProvide } from '../../../components';
```
CachedaCodes, line 50: restore this imports as they were before

CachedaCodes (approved):

PR #1270, opened 2023-08-29
feat: query preview components allow adding extra params to the request

CachedaCodes:
[EMP-1555](https://searchbroker.atlassian.net/browse/EMP-1555)

## Motivation and context
We need to gain more control with the parameters that we use to create the query preview requests. Right now we were limited to the query. In this PR we add the feature of making the query preview request with extra params too

## Type of change

- [X] Breaking change (fix or feature that causes existing functionality to not work as expected)

- [X] `Main`
- [ ] Other. Specify:

## How has this been tested?
Select a query preview that has been initialized with custom extra params. Check that the search uses those extra params until it is restarted. Check that the search uses the initial extra params afterwards.

[EMP-1555]: https://searchbroker.atlassian.net/browse/EMP-1555?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/x-components/src/store/utils/store-emitters.utils.ts
```diff
@@ -35,17 +35,19 @@ export interface StateSelector<ReturnType, State extends Dictionary, Getters ext
...
    * This function exist because Vue will not stop reactivity propagation if the observed variable
    * is an `object`, an `Array`, or the `deep` mode has been enabled.
    *
    * @param newValue - The new value.
    * @param oldValue - The old value.
+   * @param state - The state of the {@link XModule} where this selector is used.
    * @returns True if the value has really changed.
    */
-  filter?(newValue: ReturnType, oldValue: ReturnType): boolean;
+  filter?(newValue: ReturnType, oldValue: ReturnType, state: State): boolean;
```
CachedaCodes, line 50: Adding access to the module's state here to be able to filter if the event of the emiter is fired based on the current values of the state.

packages/x-components/src/x-modules/queries-preview/components/query-preview-list.vue
```diff
@@ -48,37 +58,54 @@
...
+     * @param newQueries - The new queries.
+     * @param oldQueries - The old queries.
      * @internal
      */
     @Watch('queries', { immediate: true })
-    protected resetStatusRecord(): void {
-      this.queriesStatus = {};
-      this.loadNext();
+    protected resetStatusRecord(newQueries: string[], oldQueries: string[]): void {
+      if (newQueries?.toString() !== oldQueries?.toString()) {
```
CachedaCodes, line 105: Changing the prop and watching a computed property creates false positives when pairing this component with the semantic queries. This prevents that. (Thinking of creating a better function that compares arrays and adding it to the arrays utils in `x-utils`)

packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
```diff
@@ -197,7 +198,11 @@
     protected created(): void {
       this.$watch(
         () => this.queryPreviewRequest,
-        request => this.emitQueryPreviewRequestUpdated(request)
+        (newRequest, oldRequest) => {
+          if (JSON.stringify(newRequest) !== JSON.stringify(oldRequest)) {
```
CachedaCodes, line 202: Changing the prop and watching a computed property creates false positives when pairing this component with the semantic queries. This prevents that. (Thinking of creating a better function that compares objects and adding it to the objects utils in x-utils)

packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
```diff
@@ -2,12 +2,14 @@
   <NoElement v-if="queryPreviewResults && queryPreviewResults.totalResults">
     <!--
       @slot Query Preview default slot.
-          @binding {string} query - query
+          @binding {QueryPreviewInfo} queryPreviewInfo - The information about the request of the
+          query preview
+          @binding {Dictionary<unknown>} injectedParams - The extra params from the snippetConfig
```
annacv, line 7: params are being bound inside QueryPreviewInfo one, so no need for this doc here

packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
```diff
@@ -154,6 +137,10 @@
     @Inject({ default: undefined })
     protected location?: FeatureLocation;
 
+    protected get query(): string {
```
annacv: `/**
     * The query to make the queryPreviewRequest's with
     *
     * @returns The search query string.
     * @internal
     */`

packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
```diff
@@ -154,6 +137,10 @@
...
+      return this.queryPreviewInfo.query;
+    }
+
```
annacv: Add doc, something like this
`/**
     * The query to make the queryPreviewRequest's with
     *
     * @returns The search query string.
     * @internal
     */`

packages/x-components/src/views/home/Home.vue
```diff
@@ -615,7 +621,37 @@
       }
     };
 
-    protected queriesToPreview = ['sunglasses', 'handbag', 'earrings', 'jeans', 't-shirt'];
+    protected queriesPreviewInfo: QueryPreviewInfo[] = [
+      {
+        query: 'cortina',
+        extraParams: {
+          instance: 'lolahome',
+          store: 'hola',
```
annacv, line 629: We can use here `store: 'Gijón'` to suit e2e `extra-params.feature` 's `QueryPreviewExtraParamValue`

packages/x-components/src/x-modules/queries-preview/components/query-preview-list.vue
```diff
@@ -152,7 +179,10 @@ In this example, the results will be rendered inside a sliding panel.
 
 ```vue live
 <template>
-  <QueryPreviewList :queries="queries" #default="{ query, totalResults, results }">
+  <QueryPreviewList
+    :queriesPreviewInfo="queriesPreviewInfo"
+    #default="{ query, totalResults, results }"
+  >
     <div class="x-flex x-flex-col x-gap-8 x-mb-16">
       <h1 class="x-title2">{{ query }} ({{ totalResults }})</h1>
```
annacv, line 187:
```suggestion
    #default="{ queryPreviewInfo, totalResults, results }"
  >
    <div class="x-flex x-flex-col x-gap-8 x-mb-16">
      <h1 class="x-title2">{{ queryPreviewInfo.query }} ({{ totalResults }})</h1>
```

packages/x-components/src/x-modules/related-tags/wiring.ts
```diff
@@ -124,5 +134,9 @@ export const relatedTagsWiring = createWiring({
   },
   UserClickedOutOfMainModal: {
     clearRelatedTagsQuery
+  },
+  UserAcceptedAQueryPreview: {
+    setRelatedTagsQueryPreview,
+    clearSelectedRelatedTags
```
annacv, line 140: We should add here a wiring `setRelatedTagsParamsFromQueryPreview` to commit the params from the query preview, now we are sending the global extraParams instead of the selected queryPreview ones in the request

packages/x-components/src/x-modules/popular-searches/wiring.ts
```diff
@@ -60,5 +61,5 @@ export const popularSearchesWiring = createWiring({
   },
   ExtraParamsChanged: {
     setPopularSearchesExtraParams
-  }
+  },
```
annacv, line 64: As in the related tags module, we need to react to the `UserAcceptedAQueryPreview` event and update the extraParams in this module.

```suggestion
  },
  UserAcceptedAQueryPreview: {
    setPopularSearchesParamsFromQueryPreview
  }
```
and add the commit with extraParams as payload accordingly:
`export const setPopularSearchesParamsFromQueryPreview = wireCommit(
  'setParams',
  ({ eventPayload: { extraParams } }) => extraParams
);`

packages/x-components/src/x-modules/query-suggestions/wiring.ts
```diff
@@ -118,5 +128,8 @@ export const querySuggestionsWiring = createWiring({
   },
   UserClickedOutOfMainModal: {
     clearQuerySuggestionsQuery
+  },
+  UserAcceptedAQueryPreview: {
+    setQuerySuggestionsQueryPreview
   }
```
annacv, line 134: As in RTs and Popular Searches modules, we need to update the `extraParams` here:

```suggestion
  UserAcceptedAQueryPreview: {
    setQuerySuggestionsQueryPreview,
    setQuerySuggestionsParamsFromQueryPreview
  }
```
 and add the commit above: 

`export const setQuerySuggestionsParamsFromQueryPreview = wireCommit(
  'setParams',
  ({ eventPayload: { extraParams } }) => extraParams
);`

Test: select a query preview and reopen empathize: 
Captura de pantalla 2023-08-29 a les 15 31 12

annacv:
Hi, selecting a query preview and then deleting it and selecting another queryPreview with different params is not working properly, params that change between requests are not kept (instance, store in the examples) no matter if the changes proposed are applied or not: 

https://github.com/empathyco/x/assets/21217131/18c90fb5-79d0-480e-bae9-f406673ffa1e

packages/x-components/src/x-modules/history-queries/wiring.ts
```diff
@@ -39,13 +39,34 @@ const wireDispatchWithoutPayload = namespacedWireDispatchWithoutPayload(moduleNa
  */
 export const addQueryToHistoryQueries = wireDispatch('addQueryToHistory');
 
+/**
+ * Saves a new query from a selectedQueryPreview into the history queries.
```
diegopf, line 43:
```suggestion
 * Saves the selectedQueryPreview query into the history queries.
```

packages/x-components/src/x-modules/queries-preview/components/query-preview-button.vue
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    /**
+     * List of events to emit by the BaseEventButton.
+     *
+     * @returns An object with the event and payload.
+     *
+     * @internal
+     */
+    protected get events(): Partial<XEventsTypes> {
```
diegopf, line 76: I'm thinking if it would make sense to add a prop to allow emit other events that are configured from the outside. I think that approach came handy in the past in some setups.

packages/x-components/src/x-modules/queries-preview/components/query-preview-button.vue
```diff
@@ -0,0 +1,151 @@
...
+    }
+  };
+</script>
+```
+
+## Events
+
+A list of events that the component will emit:
+
+- `UserClickedPanelToggleButton`: the event is emitted after the user clicks the button. The event
```
diegopf, line 149:
```suggestion
- `UserAcceptedAQueryPreview`: the event is emitted after the user clicks the button. The event
```

packages/x-components/src/x-modules/queries-preview/components/query-preview-button.vue
```diff
@@ -0,0 +1,151 @@
...
+  payload is the id of the panelId that is going to be toggled.
```
diegopf, line 150:
```suggestion
  payload is the `QueryPreviewInfo` of the query.
```

packages/x-components/src/x-modules/queries-preview/store/module.ts
```diff
@@ -12,9 +12,14 @@ import { loadedQueriesPreview } from './getters/loaded-queries-preview.getter';
 export const queriesPreviewXStoreModule: QueriesPreviewXStoreModule = {
   state: () => ({
     config: {
+      debounceInMs: 250,
```
diegopf, line 15: are we actually using this? If we do, I do not see where 😅

diegopf (changes requested):

packages/x-components/src/views/home/Home.vue
```diff
@@ -348,8 +353,8 @@
                 </h1>
                 <LocationProvider :location="$x.noResults ? 'no_results' : 'low_results'">
                   <QueryPreviewList
-                    :queries="queries"
-                    #default="{ query, results }"
+                    :queries-preview-info="queries.map(q => ({ query: q }))"
```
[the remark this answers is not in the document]
CachedaCodes: but `queries` comes from the scope slot from `SemanticQueries`, I can't make it a property. I could move the mapping to a function that receives the queries list.

packages/x-components/src/x-modules/queries-preview/store/module.ts
CachedaCodes: Nope, leftover from previous implementation (like 20 commits before 😆)

diegopf (dismissed):

packages/x-components/src/x-modules/queries-preview/events.types.ts
```diff
@@ -17,4 +19,16 @@ export interface QueriesPreviewXEvents {
...
+   * User has clicked on a query preview.
+   * Payload: The {@link QueryPreviewInfo | query preview info}
+   * of the selected {@link QueryPreviewItem | item}.
+   */
+  UserAcceptedAQueryPreview: QueryPreviewInfo;
+  /**
+   * The selected query preview has changed.
+   * Payload: The new selected query preview with extra params or the initial module
+   * params if the query has been cleared.
+   */
```
annacv, line 32: Hi!
If I am not wrong, I think this doc should be updated to be more precise, at least the payload description, as now it is not a change event that will always be fired.

packages/x-components/src/x-modules/queries-preview/components/query-preview-button.vue
annacv: Hi! 
That's a good point to have in mind, but I think we should go on to be able to implement the filters part ASAP.

packages/x-components/src/x-modules/semantic-queries/wiring.ts
```diff
@@ -8,7 +8,7 @@ import { SearchXEvents } from '../search/events.types';
 import { ExtractMutationPayload } from '../../store/store.types';
 
 /**
- * `semanticQueries` {@link XModuleName name}.
+ * `semanticQueries` {@link XModuleName | name}.
```
annacv, line 11: Hi!
Shouldn't we also update this module `extraParams` when `QueryPreviewUnselected` is emitted??

packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
CachedaCodes: Doesn't work for deeply nested objects, I've [opened a task](https://searchbroker.atlassian.net/browse/EMP-2123) to have a deep object comparator and then [another](https://searchbroker.atlassian.net/browse/EMP-2124) to use it here.

diegopf (approved):

annacv (approved):
🚀

PR #1272, opened 2023-08-30
Support max-len in Vue components

### LINKED ISSUE #1063: [FEATURE]: support max-len in Vue components
### How can the project be improved?

In the Playboard team, we use the `eslint-plugin-x` and during development, we found out that `eslint` with its `max-len` rule and `prettier` didn't work together. The main problem is there are some cases that `prettier` doesn't want to transform the code even when the code's length surpasses the limit.

Some cases:
- Strings length (e.g: long imports paths, long attribute names in `vue` files, long `describe()` titles, long URL ). Further [info](https://prettier.io/docs/en/rationale.html#print-width) 
- Template Literals when there are some expressions inside `prettier` doesn't do a wrap transformation itself. It is a known [discussion](https://github.com/prettier/prettier/issues/3368) topic of the `prettier` formatter.
```typescript 
imports { foo } from 'very-long-long-long-long-long-long-long-long-long-import-path'; // ERROR: eslint max-len

const x = `${msgs?.explanation1}<br>${msgs?.explanation2}<ul><li>"${promotions[0].title}" with queries: ${promotions[0].keywords[0]}</li></ul>`// ERROR: eslint max-len

describe('very-long-long-long-long-long-long-long-long-long-long-title')// ERROR: eslint max-len
```
- CSS long lines. There are some cases in that you can not add a line break.
```vue
@include example( 'very-long-long-long-long-long-long-long-long-long-long-class');
```
- long attributes in Vue components. There are some cases that you can not add a line break in an attribute.
```vue
<div  class="example-component example-component--expanded">
```
```vue
<path
   d="M11 .5c-1.7.2-3.4.9-4.7 2-1.1.9-2 2-2.5 3.2-1.2 2.4-1.2 5.1-.1 7.7 1.1 2.6 2.8 5 5.3 7.5 1.2 1.2 2.8 2.7 3 2.7 0 0
/>"
```

### How can this be solved?

- Make the linter ignore strings
- Make the linter ignore template literals
- Make the linter ignore long CSS lines inside Vue
- Make the linter ignore Vue components' attributes

### Proposed solution

In Play team core repository we defined a [configuration](https://github.com/empathyco/play-playboard-core/blob/master/.eslintrc.js) in `.eslintrc.js` which solved these issues.

* We ignored all the [strings](https://github.com/empathyco/play-playboard-core/blob/master/.eslintrc.js#L19) with the `ignoreStrings` flag.
* We [ignored all the template literals ](https://github.com/empathyco/play-playboard-core/blob/master/.eslintrc.js#L28)with the `ignoreTemplateLiterals` flag.
* We also [ignored](https://github.com/empathyco/play-playboard-core/blob/master/.eslintrc.js#L29) all the URLs, because there were some big urls in the `play-core` [repository](https://github.com/empathyco/play-playboard-core).
* We also [ignore](https://github.com/empathyco/play-playboard-core/blob/master/.eslintrc.js#L128) attribute length in Vue components with the `ignoreHTMLAttributeValues` flag.

### Customizations supported

I would suggest using our approach or something along those lines in `eslint-plugin-x` so everyone could have a general config that solves these issues.

Suggestions:
* Do an override in `.vue` files, which disabled the original `max/len` but used the extended one `vue/max-len`
```Javascript
{
      files: ['*.vue'],
      rules: {
        'max-len': 'off',
        'vue/max-len': [
          'error',
          {
           ignoreStrings: true,
           ignoreTemplateLiterals: true,
            ignoreUrls: true,
            ignoreHTMLAttributeValues: true
          }
        ]
      }
```
* Change config for `max-len`
```javascript
{
  rules: {
      'max-len': ['error', {
                         ignoreStrings: true,
                         ignoreTemplateLiterals: true,
                         ignoreUrls: true      
                         }],
  }
}
```

### Additional information

Previously prettier would have formatted this with one class per line. like so:
```vue
 <div class="example-component
                        example-component--expanded
                         example-component--collapsed"
  >
```
But it is no longer an option since [prettier@2.5.0](https://prettier.io/blog/2021/11/25/2.5.0.html#collapse-html-class-attributes-onto-one-line-11827httpsgithubcomprettierprettierpull11827-by-jlongsterhttpsgithubcomjlongster).

Further info: 
- [eslint max-len](https://eslint.org/docs/latest/rules/max-len) 
- [vue max-len which extends `eslint` config](https://eslint.vuejs.org/rules/max-len.html)
- [play eslint config](https://github.com/empathyco/play-playboard-core/blob/master/.eslintrc.js#L19)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

annacv:
[EMP-1908](https://searchbroker.atlassian.net/browse/EMP-1908)

resolves #1063
[Ref issue](https://github.com/empathyco/x/issues/1063)

TEST
1- [Here](https://github.com/vuejs/eslint-plugin-vue/blob/master/docs/rules/max-len.md#ignorestrings-true) are some examples for testing `ignoreUrls`, `ignoreStrings`, `ignoreTemplateLiterals` rules, that can be applied in, for example, `Home.vue`

2- Run eslint in your project:
run `build eslint-plugin-x`
run `pnpm install —frozen-lockfile` in x-components
run `npm run lint`  and check:
- It should ignore long urls in vue files
- It should ignore long classes, for example line 331 in Home.vue
- It should ignore icon files as ‘user-filled.vue’

[EMP-1908]: https://searchbroker.atlassian.net/browse/EMP-1908?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/eslint-plugin-x/lib/configs/eslint.js
+++ b/packages/eslint-plugin-x/lib/configs/eslint.js
@@ -9,7 +9,13 @@ module.exports = {
       indent: 'off',
       'max-len': [
         'error',
-        { code: 100, ignoreStrings: true, ignoreTemplateLiterals: true, ignoreUrls: true }
+        {
+          code: 100,
+          ignoreStrings: true,
+          ignoreTemplateLiterals: true,
+          ignoreUrls: true,
+          ignorePattern: 'class=".*"$'
+        }
       ],
       'no-alert': process.env.NODE_ENV === 'production' ? 'error' : 'off',
       'no-console': process.env.NODE_ENV === 'production' ? 'error' : 'off',
--- a/packages/eslint-plugin-x/lib/configs/eslint.js
+++ b/packages/eslint-plugin-x/lib/configs/eslint.js
@@ -7,8 +7,10 @@ module.exports = {
       curly: ['error', 'all'],
       eqeqeq: ['error', 'always', { null: 'ignore' }],
       indent: 'off',
-      // THESE 'max-len' rules ARE ALMOST THE SAME OVERRIDE CONFIG THAT IS SET IN x-components/.eslintrc.js for vue files, except the ignorePattern, so we will override it below in vue files
-      'max-len': ['error', { code: 100, ignoreComments: false, ignorePattern: 'class=".*"$' }],
+      'max-len': [
+        'error',
+        { code: 100, ignoreStrings: true, ignoreTemplateLiterals: true, ignoreUrls: true }
+      ],
       'no-alert': process.env.NODE_ENV === 'production' ? 'error' : 'off',
       'no-console': process.env.NODE_ENV === 'production' ? 'error' : 'off',
       'no-debugger': process.env.NODE_ENV === 'production' ? 'error' : 'off',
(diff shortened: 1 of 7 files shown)


herrardo (changes requested):
I think the proposal in the issue was correct, it was to add this configuration to the `eslint-plugin-x` config, so everybody using that package would be using the new rules.

That would mean to add it here:

https://github.com/empathyco/x/blob/main/packages/eslint-plugin-x/lib/configs/eslint.js

packages/eslint-plugin-x/lib/configs/eslint.js
```diff
@@ -23,6 +23,48 @@ module.exports = {
...
+           * the ".vue" files.
+           * It helps us in cases that the "<style>" tag's section has long lines of
+           * code.
+           * (e.g: "@include foo(background-color, color, $selector: '#{$component-class}__kpi');
+           * https://github.com/vuejs/eslint-plugin-vue/blob/master/docs/rules/max-len.md
+           */
+          'vue/max-len': [
+            'error',
+            {
+              ignorePattern: '|class=".*"$ |https://github',
```
annacv, line 42: This `ignorePattern` config was previously set in [x-components' eslint config](https://github.com/empathyco/x/blob/main/packages/x-components/.eslintrc.js#L22).

packages/x-components/.eslintrc.js
```diff
@@ -12,21 +12,12 @@ module.exports = {
...
-        'max-len': [
-          'error',
-          {
-            code: 100,
-            ignoreComments: false,
-            ignorePattern: '|class=".*"$ |https://github'
-          }
-        ]
-      }
-    },
```
annacv, line 26: As we are disabling the entire `max-len` rule for vue components, in order to use` vue/max-len` one, we get rid of this config here.

packages/x-components/.eslintrc.js
```diff
@@ -12,20 +12,7 @@ module.exports = {
...
-    {
-      files: 'src/components/icons/*.vue',
+      files: ['*.vue'],
       rules: {
         'max-len': 'off'
       }
```
annacv, line 18: This is the only way I found to effectively disable `max-len`, I tried in eslint-x plugin file but it didn't have effect

annacv:
This config seems not to be working properly

PR #1277, opened 2023-09-11
feat(scroll): add top offset when automatically scrolling to main-scroll-item

CachedaCodes:
[EMP-2021](https://searchbroker.atlassian.net/browse/EMP-2021)

## Motivation and context
Depending on the setup's header it was possible that, when automatically scrolling to an element in the main scroll, the element was cut by the header. This feature adds a prop to the main-scroll-item to add an offset to the scrolling.

## Type of change

- [X] `Main`
- [ ] Other. Specify:

## How has this been tested?
Set the prop and reload a url that has `scroll` attribute.

[EMP-2021]: https://searchbroker.atlassian.net/browse/EMP-2021?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/x-components/src/x-modules/scroll/components/main-scroll-item.vue
+++ b/packages/x-components/src/x-modules/scroll/components/main-scroll-item.vue
@@ -3,7 +3,7 @@
     :is="tag"
     v-on="$listeners"
     :data-scroll="item.id"
-    :style="{ scrollMarginBottom: `-${scrollTopOffset}px` }"
+    :style="{ scrollMarginTop: `${scrollTopOffset}px` }"
   >
     <slot />
   </component>
@@ -112,8 +112,8 @@
       oldObserver?.unobserve(this.$el);
       newObserver?.observe(this.$el);
       if (this.pendingScrollTo === this.item.id) {
-        this.$el.scrollIntoView({
-          block: 'nearest'
+        Vue.nextTick(() => {
+          this.$el.scrollIntoView();
         });
         this.$x.emit('ScrollRestoreSucceeded');
       }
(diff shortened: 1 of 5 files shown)


herrardo (commented):
I was curious about this and started testing it, but I can't do what I want, it might be some kind of uselessness on my end, but I would like to check it together.

lauramargar (commented):
I have tried to test it on the archetype and it seems like the element to scroll is always at the bottom. Maybe we can check it together.

lauramargar:
It works perfectly with the regular results (`result` component), but I have tried to reload the page with a result of `CustomNextQueriesPreview` and it doesn't work. I have talked with @CachedaCodes and we found that the problem is an `overflow: hidden` in the sliding panel of the `CustomNextQueriesPreview` component.

packages/x-components/src/x-modules/scroll/components/main-scroll-item.vue
```diff
@@ -99,7 +112,9 @@
       oldObserver?.unobserve(this.$el);
       newObserver?.observe(this.$el);
       if (this.pendingScrollTo === this.item.id) {
-        this.$el.scrollIntoView();
+        Vue.nextTick(() => {
+          this.$el.scrollIntoView();
```
herrardo, line 116:
```suggestion
          this.$el.scrollIntoView({block: 'center'})
```

After doing some testing, this seems a good option. This option centers always the item in the viewport, so we would have some part of the previous row visible, but the specified item will be always in the view port.

Downside: We would have to modify some test probably.

herrardo (changes requested):

lauramargar (approved):

PR #1299, opened 2023-09-25
bump lerna to v7

diegopf:
This PR updates Lerna to version 7. The decision to update Lerna was made because Prettier v3 requires it, and this transition has been seamless for us.

The changes introduced in this PR go beyond just bumping the version numbers. They also include updates made necessary by running `lerna repair`, as recommended in the Lerna v7 [changelog](https://github.com/lerna/lerna/blob/main/CHANGELOG.md#breaking-changes).

### DIFF
--- a/nx.json
+++ b/nx.json
@@ -21,5 +21,11 @@
     "test": {
       "outputs": ["{projectRoot}/coverage"]
     }
+  },
+  "$schema": "./node_modules/nx/schemas/nx-schema.json",
+  "namedInputs": {
+    "default": ["{projectRoot}/**/*", "sharedGlobals"],
+    "sharedGlobals": [],
+    "production": ["default"]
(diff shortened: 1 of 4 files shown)


nx.json
```diff
@@ -21,5 +21,11 @@
     "test": {
       "outputs": ["{projectRoot}/coverage"]
     }
+  },
+  "$schema": "./node_modules/nx/schemas/nx-schema.json",
+  "namedInputs": {
```
diegopf, line 26: This configuration is related to how the nx cache works and when it should be invalidated, [see](https://nx.dev/concepts/more-concepts/customizing-inputs) Given that we had issues with the cache in the past, we should take a look at this in the future.

annacv (approved):
gj!

PR #1284, opened 2023-09-13
chore(deps): update prettier version

diegopf:
Update prettier and related plugins to latest version.

PR #1301, opened 2023-09-26
Revert "chore(deps): update prettier version"

diegopf:
Reverts empathyco/x#1284


herrardo (approved):
I have launched the eDocs Dynamic Sync and its creating this PR, where only modified vue components are in the PR, and the docs are not being removed, so it's ok from my point of view:

https://github.com/empathyco/docs-empathy/pull/1423/files

PR #1305, opened 2023-09-27
Feature/emp 2048 create response mapper

AnaOstio:
# Pull request template

## Motivation and context
In this PR we have implemented the response mapper, with its types and schema.

- [X] Open issue. If applicable, link: [EMP-2048](https://searchbroker.atlassian.net/browse/EMP-2048)

## Type of change

- [ ] `Main`
- [X] Other. Specify: [feature/EMP-1939-create-endpoint-adapter](https://github.com/empathyco/x/tree/feature/EMP-1939-create-endpoint-adapter)

## How has this been tested?

## Checklist:

### DIFF
--- a/packages/x-adapter-platform/src/schemas/models/experience-controls.schema.ts
+++ b/packages/x-adapter-platform/src/schemas/models/experience-controls.schema.ts
@@ -1,0 +1,11 @@
+import { createMutableSchema } from '@empathyco/x-adapter';
+import { ExperienceControlsResponse } from '@empathyco/x-types';
+import { PlatformExperienceControlsResponse } from '../../types';
+
+export const experienceControlsSchema = createMutableSchema<
+  PlatformExperienceControlsResponse,
+  ExperienceControlsResponse
+>({
+  numberOfCarousels: 'numberOfCarousels',
+  resultsPerCarousels: 'resultsPerCarousels'
+});
--- a/packages/x-adapter-platform/src/schemas/responses/experience-controls-response.schema.ts
+++ b/packages/x-adapter-platform/src/schemas/responses/experience-controls-response.schema.ts
@@ -1,0 +1,11 @@
+import { createMutableSchema } from '@empathyco/x-adapter';
+import { ExperienceControlsResponse } from '@empathyco/x-types';
+import { PlatformExperienceControlsResponse } from '../../types';
+
+export const experienceControlsResponseSchema = createMutableSchema<
+  PlatformExperienceControlsResponse,
+  ExperienceControlsResponse
+>({
+  numberOfCarousels: 'numberOfCarousels',
+  resultsPerCarousels: 'numberOfCarousels'
+});
--- a/packages/x-adapter-platform/src/types/responses/experience-controls-response.model.ts
+++ b/packages/x-adapter-platform/src/types/responses/experience-controls-response.model.ts
@@ -1,0 +1,9 @@
+/**
+ * Response for the `experience controls` endpoint.
+ *
+ * @public
+ */
+export interface PlatformExperienceControlsResponse {
+  numberOfCarousels: number;
+  resultsPerCarousels: number;
+}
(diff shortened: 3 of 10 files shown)


packages/x-adapter-platform/src/schemas/models/experience-controls.schema.ts
CachedaCodes, line 1: This file is not needed.
The models in this folder are the small schemas that bigger schemas use (the model schema `result` is used in the response schema from `search`).
You're just duplicating code here.

packages/x-adapter-platform/src/schemas/responses/experience-controls-response.schema.ts
```diff
@@ -0,0 +1,11 @@
...
+import { ExperienceControlsResponse } from '@empathyco/x-types';
+import { PlatformExperienceControlsResponse } from '../../types';
+
+export const experienceControlsResponseSchema = createMutableSchema<
+  PlatformExperienceControlsResponse,
+  ExperienceControlsResponse
+>({
+  numberOfCarousels: 'numberOfCarousels',
+  resultsPerCarousels: 'numberOfCarousels'
+});
```
CachedaCodes, line 11: This is the structure that a normal mapper for this endpoint would use, but there are 2 things to take into account:
1. The nature of the endpoint is open to change. Not only we don't know if these are the fields that the endpoint will return by default for all the setups, we also don't know if these are the only fields that it will return.
2. This structure is not compatible with the front components (the module). We need to have an object wrapping the configuration coming from the endpoint (`controls`, for example) because we'll also need an `events` object (populated in the setups, not here)


I believe something like this should work and makes sense for what we want to do (you'll need to adapt the typings):

```suggestion
>({
  controls: (response) => response,
  events: {}
});
```

packages/x-adapter-platform/src/types/responses/experience-controls-response.model.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Response for the `experience controls` endpoint.
+ *
+ * @public
+ */
+export interface PlatformExperienceControlsResponse {
+  numberOfCarousels: number;
+  resultsPerCarousels: number;
```
CachedaCodes, line 8: This interface will need to change

packages/x-types/src/response/experience-controls-response.model.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Response for the experience controls endpoint.
+ *
+ * @public
+ */
+export interface ExperienceControlsResponse {
+  numberOfCarousels: number;
+  resultsPerCarousels: number;
```
CachedaCodes, line 8: This interface will need to change

CachedaCodes (changes requested):

packages/x-adapter-platform/src/types/responses/experience-controls-response.model.ts
```diff
@@ -0,0 +1,11 @@
...
+
+/**
+ * Response for the `experience controls` endpoint.
+ *
+ * @public
+ */
+export interface PlatformExperienceControlsResponse {
+  controls: Dictionary;
+  events: Dictionary;
+}
```
CachedaCodes, line 11: The response type in this package represents the structure of the endpoint's response. The endpoint is not returning neither `controls` or `events` properties, is returning a dictionary with anything inside, so this should be a `Dictionary<unknown>`.

```suggestion
export interface PlatformExperienceControlsResponse extends Dictionary<unknown> {}
```

packages/x-types/src/response/experience-controls-response.model.ts
```diff
@@ -0,0 +1,10 @@
+import { Dictionary } from '@empathyco/x-utils';
+/**
+ * Response for the experience controls endpoint.
+ *
+ * @public
+ */
+export interface ExperienceControlsResponse {
+  controls: Dictionary;
+  events: Dictionary;
```
CachedaCodes, line 9:
```suggestion
  controls: Dictionary<unknown>;
  events: Dictionary<unknown>;
```

CachedaCodes (changes requested):

CachedaCodes (approved):

PR #1313, opened 2023-10-04
feat: utils for the mutation of states with config

CachedaCodes:
EMP-2327

## Motivation and context
We will now need to change the configs in the module programatically. This utils can be used by any module to do so.

## Type of change

- [X] `Main`
- [ ] Other. Specify:

## How has this been tested?
Import this utils to a module that has config in the state, similarly to what we are doing with the status.


lauramargar (approved):
Good job!

PR #1315, opened 2023-10-06
feat: filters in history queries

lauramargar:
# Pull request template
 
We want to save filters in history queries. At the moment we are only saving the query and total results, so when you search for a query and apply filters to it, it means that these filters are important in your search. So, why don't we save them? We think that it would be a useful feature for the user. 

The idea is to search for a query, apply filters and remove it. Then, if you click on the history query, the search request should contain the query and the filters applied previously.

## Motivation and context

- [X] Open issue. If applicable, link: [EMP-2208](https://searchbroker.atlassian.net/browse/EMP-2208)

## Type of change

- [X] `Main`
- [ ] Other. Specify:

## How has this been tested?

You can follow these steps:
- Search for a query
- Apply filters
- Remove query
- Select the same query in the history queries column
- Check if the filters are applied 

## Checklist:

[EMP-2208]: https://searchbroker.atlassian.net/browse/EMP-2208?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/x-components/src/x-modules/facets/wiring.ts
```diff
@@ -166,6 +166,16 @@ export const setSelectedFiltersFromPreview = wireCommit(
   ({ eventPayload: { filters } }) => (filters ? createRawFilters(filters) : [])
 );
 
+/**
+ * Sets the filters of the facets module from a selectedHistoryQuery's filters.
+ *
+ * @public
+ */
+export const setSelectedFiltersFromHitoryQueries = wireCommit(
```
annacv, line 174: typo here `setSelectedFiltersFromHitoryQueries`

packages/x-components/src/x-modules/history-queries/store/actions/update-history-queries-with-search-response.action.ts
```diff
@@ -29,18 +29,45 @@ export const updateHistoryQueriesWithSearchResponse: HistoryQueriesXStoreModule[
...
+            totalResults: searchResponse.totalResults,
+            filters: filters
           };
           return dispatch('setHistoryQueries', newHistoryQueries);
         }
       }
     }
   };
+
+// eslint-disable-next-line jsdoc/require-description
```
annacv, line 50: comment can be removed
```suggestion
```

packages/x-components/src/x-modules/history-queries/store/__tests__/actions.spec.ts
```diff
@@ -355,7 +355,7 @@ describe('testing history queries module actions', () => {
       expectHistoryQueriesToEqual([{ ...gato, totalResults }, perro]);
     });
 
-    it('does not update a history query if the new totalResults is lower', async () => {
+    it('updates a history query if the new totalResults is lower', async () => {
```
annacv, line 358: Hi! I'm not sure if this test definition matches the condition that is being tested. Should not it be related to this one: `* 3. If there is a new search response.`?

lauramargar: Hi @annacv!! This test definition matches with the condition that is being tested because the number of totalResults of the response is 2 and totalResults saved into history query ("gato") are set to 50. 
So `"new totalResults"(2) < "historyQuery totalResults" (50)`

Screenshot 2023-10-11 at 08 14 54

Related to your question, I think it's not necessary to make a test with this condition `3. If there is a new search response`, because when we test if the new totalResults are lower or higher than historyQuery totalResults, we are testing the totalResults of a new search response.

annacv: Hi, @lauramargar, I meant adding a test to show that History Queries are being saved with filters 👀 .
And, about the definition, I was willing if it should be something like "updates a history query with a new response", I'll ping you to check if I'm lost in some way.

packages/x-components/src/x-modules/history-queries/store/actions/update-history-queries-with-search-response.action.ts
```diff
@@ -29,18 +29,44 @@ export const updateHistoryQueriesWithSearchResponse: HistoryQueriesXStoreModule[
...
+    const facet = requestFilters[key];
+    facet.forEach(filter => {
+      if (filtersApplied.includes(filter)) {
+        const filterIndex = filtersApplied.findIndex(value => value.id === filter.id);
+        filtersApplied.splice(filterIndex, 1);
+      } else {
+        filtersApplied.push(filter);
+      }
+    });
+  });
```
CachedaCodes, line 70: Here you're iterating the keys of the `requestFilters` records to then use it to retrieve the value, you can simplify and iterate the values directly
`Object.values(requestFilters).forEach(filter => {`

Also, if in the end you don't need to do the splice, you can try and make this with a `reduce`

````typescript
filtersApplied = Object.values(requestFilters).reduce((accFilters, filters) => {
  accFilters.push(...filters);
  return accFilters;
});
````

File: packages/x-types/src/history-query.model.ts [hunk 5]
```diff
@@ -9,4 +10,6 @@ import { Previewable } from './previewable.model';
 export interface HistoryQuery extends Previewable, NamedModel<'HistoryQuery'> {
   /** Timestamp when the history query was created. */
   timestamp: number;
+  /** Filters selected for the query to search for. */
+  filters?: Filter[];
```
CachedaCodes, line 14:
```suggestion
  selectedFilters?: Filter[];
```

annacv (approved):
🚀

PR #1318, opened 2023-10-09
fix: set query in facets module from selecting a query preview

CachedaCodes:
EMP-2407

## Motivation and context
The query of the facets module wasn't being updated when a query preview was selected and that caused probles with the events from the emitters not having the correct `oldValue`.

## Type of change

- [X] `Main`
- [ ] Other. Specify:

## How has this been tested?
Click a query preview with filters (`cortina` for example) and search for something that doesn't have results (a spellchecked query, `pantlon` for example). Only the spellcheck message should appear. Check removing the new wire and two messages appear doing the same.


annacv (approved):
🎉

PR #1320, opened 2023-10-10
Replace usage of `secrets.SUPPORT_TOKEN` PAT with standard `github.token`

Roberdvs:
# Pull request template
 

## Motivation and context
Full context of this change on https://searchbroker.atlassian.net/browse/OPS-8556

I've replaced the usage of `secrets.SUPPORT_TOKEN` with the [standard github.token](https://docs.github.com/en/actions/security-guides/automatic-token-authentication) where possible, with the exception of the edocs workflows since those target a separate private repository, where the workflow wouldn't have proper permissions since permissions of the `github.token` are scoped to the current repository so, in that case, a Personal Access Token (PAT) is still necessary. We might revisit that trigger in the future.

Also removed the `github_token` input when calling the `ad-m/github-push-action` action as `${{ github.token }}` it's already the [default](https://github.com/ad-m/github-push-action#inputs)

## Type of change

- [x] `Main`
- [ ] Other. Specify:

### DIFF
--- a/.github/actions/release-alpha/action.yml
+++ b/.github/actions/release-alpha/action.yml
@@ -4,6 +4,9 @@ inputs:
   npm_token:
     description: Token to push code to npm
     required: true
+  github_token:
+    description: Token to push code to github
+    required: true
 runs:
   using: 'composite'
   steps:
@@ -30,5 +33,6 @@ runs:
     - name: Push tags
       uses: ad-m/github-push-action@master
       with:
+        github_token: ${{ inputs.github_token }}
         branch: ${{ github.ref }}
         tags: true
--- a/.github/actions/release-alpha/action.yml
+++ b/.github/actions/release-alpha/action.yml
@@ -4,9 +4,6 @@ inputs:
   npm_token:
     description: Token to push code to npm
     required: true
-  github_token:
-    description: Token to push code to github
-    required: true
 runs:
   using: 'composite'
   steps:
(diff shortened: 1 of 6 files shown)


diegopf (commented):
@Roberdvs we have the main branch write protected, so every change must be done through a PR. However, the Lerna's release process generates a new commit that needs to be push to main from the CI. Currently we use the `SUPPORT_TOKEN` to bypass the write protection for this use case. Is the `github.token` able to do that? Last time I checked I think was not possible but maybe we can change the default permissions.

Roberdvs:
> @Roberdvs we have the main branch write protected, so every change must be done through a PR. However, the Lerna's release process generates a new commit that needs to be push to main from the CI. Currently we use the `SUPPORT_TOKEN` to bypass the write protection for this use case. Is the `github.token` able to do that? Last time I checked I think was not possible but maybe we can change the default permissions.

You're right, seems that still isn't an option:
- https://github.com/orgs/community/discussions/25305
- https://github.com/orgs/community/discussions/13836

And the workaround is to either use a PAT or use custom GitHub app, which can be configured to bypass branch protections (See [here](https://github.blog/changelog/2022-05-17-consistently-allow-github-apps-as-exceptions-to-branch-protection-rules/) & [here](https://github.com/orgs/community/discussions/13836#discussioncomment-4605864)).

I'll rollback the change for now on the `publish-release.yml` workflow and we'll come back to this later.

diegopf (approved):

PR #1321, opened 2023-10-11
Feature/emp 2328 implement config utils

annacv:
[EMP-2328](https://searchbroker.atlassian.net/browse/EMP-2328)

# Pull request template
 
A new store util has [been created](https://searchbroker.atlassian.net/browse/EMP-2327). This PR uses this config store util in the modules that have a config, no matter if they are actually doing it.

## Motivation and context

The aim of this task is to enable modifying the module's config from outside them (e.g., via the experience controls module).

## Type of change

- [x] `Main`
- [ ] Other. Specify:

## How has this been tested?
- [x] Existing **unit tests pass locally** with my changes.


packages/x-components/src/x-modules/empathize/store/types.ts
```diff
@@ -40,6 +40,12 @@ export interface EmpathizeMutations {
    * @param isOpen - The new isOpen state to save.
    */
   setIsOpen(isOpen: boolean): void;
+  /**
+   * Sets the {@link EmpathizeState.config } config.
+   *
+   * @param config - The new config.
+   */
+  setConfig(config: EmpathizeConfig): void;
```
CachedaCodes, line 48: This type already exists in the `config-store.utils.ts`. You have to import the mutations type and make the mutations extend that type.

packages/x-components/src/x-modules/empathize/store/module.ts
```diff
@@ -14,7 +15,8 @@ export const empathizeXStoreModule: EmpathizeXStoreModule = {
   mutations: {
     setIsOpen(state, isOpen) {
       state.isOpen = isOpen;
-    }
+    },
+    setConfig
```
CachedaCodes, line 19: You're missing the `mergeConfig` mutation.

packages/x-components/src/x-modules/tagging/store/module.ts
```diff
@@ -26,9 +27,7 @@ export const taggingXStoreModule: TaggingXStoreModule = {
     setQueryTaggingInfo(state, queryTaggingInfo: TaggingRequest) {
       state.queryTaggingInfo = queryTaggingInfo;
     },
-    setTaggingConfig(state, taggingConfig) {
-      Object.assign(state.config, taggingConfig);
-    }
```
CachedaCodes, line 31: Removing this mutation (or renaming it) is a breaking change. You have to document it in the PR description and [in the commit message](https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-description-and-breaking-change-footer)

packages/x-components/src/x-modules/search/store/module.ts
```diff
@@ -81,9 +82,7 @@ export const searchXStoreModule: SearchXStoreModule = {
     setPage(state, page) {
       state.page = page;
     },
-    setPageSize(state, pageSize) {
-      state.config.pageSize = pageSize;
-    },
```
CachedaCodes, line 86: Removing this mutation (or renaming it) is a breaking change. You have to document it in the PR description and [in the commit message](https://www.conventionalcommits.org/en/v1.0.0/#commit-message-with-description-and-breaking-change-footer)

CachedaCodes (changes requested):

packages/x-components/src/x-modules/empathize/store/types.ts
annacv: Hi, regarding this, if we extend the mutations we shouldn't need to declare them here, so we could remove `setConfig` & `mergeConfig` types from here. If this is okay, why are we keeping `setQuery` mutation types in modules if we are extending `QueryMutations`??

PR #1322, opened 2023-10-13
feat: using setQuery util in tests that need it

albertjcuac:
# Pull request template
 

## Motivation and context

This task is to check in the different modules that all mutations of a query are taking advantage of the setQuery util functionality and import the setQuery & use it if necessary. I noticed that in some tests we weren't using the setQuery function

- [X] Open issue. If applicable, link: [EMP-2424](https://searchbroker.atlassian.net/browse/EMP-2424)

## Type of change

- [X] `Main`
- [ ] Other. Specify:

[EMP-2424]: https://searchbroker.atlassian.net/browse/EMP-2424?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/x-components/src/x-modules/semantic-queries/store/module.ts
+++ b/packages/x-components/src/x-modules/semantic-queries/store/module.ts
@@ -1,10 +1,10 @@
 import { mergeConfig, setConfig } from '../../../store/utils/config-store.utils';
+import { setQuery } from '../../../store/utils/query.utils';
 import { SemanticQueriesXStoreModule } from './types';
 import { fetchSemanticQuery } from './actions/fetch-semantic-query.action';
 import { fetchAndSaveSemanticQuery } from './actions/fetch-and-save-semantic-query.action';
 import { request } from './getters/request.getter';
 import { normalizedQuery } from './getters/normalized-query.getter';
-
 /**
  * {@link XStoreModule} For the `semantic-queries` module.
  *
@@ -32,9 +32,7 @@ export const semanticQueriesXStoreModule: SemanticQueriesXStoreModule = {
     setSemanticQueries(state, queries) {
       state.semanticQueries = queries;
     },
-    setQuery(state, query) {
-      state.query = query;
-    },
+    setQuery,
     setTotalResults(state, totalResults) {
       state.totalResults = totalResults;
     },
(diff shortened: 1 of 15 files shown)


CachedaCodes (changes requested):
gj!

Aside from replacing the mutations in the `module.ts` of the modules you also have to check the types. The mutations type of the modules should extend of `QueryMutations` instead of defining the `setQuery` type. Check the History Queries module to see an example.

packages/x-components/src/x-modules/semantic-queries/store/types.ts
```diff
@@ -4,13 +4,14 @@ import { XActionContext } from '../../../store/actions.types';
...
 import { ConfigMutations } from '../../../store/utils/config-store.utils';
+import { QueryMutations, QueryState } from '../../../store/utils/query.utils';
 
 /**
  * SemanticQueries store state.
  *
  * @public
  */
-export interface SemanticQueriesState {
+export interface SemanticQueriesState extends QueryState {
```
CachedaCodes, line 14: Aside from extending `QueryState` you have to remove the `query` type definition in the state. `QueryState` already has the `query` type definition in it, so you're basically duplicating the type definition.

There are several modules that do this wrong so remember to fix those too.

CachedaCodes (changes requested):

packages/x-components/src/x-modules/semantic-queries/store/types.ts
annacv: Hi @albertjcuac, @CachedaCodes : I think this should be also done for mutations if we are extending `QueryMutations`: is that right?
Example below: 
Captura de pantalla 2023-10-18 a les 8 47 59

CachedaCodes: Yes, that's true

packages/x-components/src/x-modules/semantic-queries/store/module.ts
```diff
@@ -1,10 +1,10 @@
 import { mergeConfig, setConfig } from '../../../store/utils/config-store.utils';
+import { setQuery } from '../../../store/utils/query.utils';
 import { SemanticQueriesXStoreModule } from './types';
 import { fetchSemanticQuery } from './actions/fetch-semantic-query.action';
 import { fetchAndSaveSemanticQuery } from './actions/fetch-and-save-semantic-query.action';
 import { request } from './getters/request.getter';
 import { normalizedQuery } from './getters/normalized-query.getter';
-
```
CachedaCodes, line 7: Here should be a linebreak

CachedaCodes (changes requested):

annacv (approved):
gj! 
I have been searching and I think it is all done!

PR #1325, opened 2023-10-16
Set correctly the filters in History Query when selecting a Query Preview with filters

annacv:
[EMP-2428](https://searchbroker.atlassian.net/browse/EMP-2428)

When we save the filters in the History Queries action update-history-queries-with-search-response.action.ts we were doing: `const filters = createHistoryQueriesFiltersList(searchResponse.request.filters);` In the case of filters that come from a `QueryPreview`, we only have` id` so we should compare the filters with the response to get the rest of the data.

In this PR we use facets with selected filters from the response to be able to provide a label for the `selectedFilters` although the stored item has only the `id` available.

**TEST**
Use VueDevTools and check the history queries state: 
* click a `queryPreview` with filters (e.g. `cortina`) and check the historyQuery's filters have a label.

Check that saving filters on history queries is still working properly:
* write a query & filter it, delete the query. If you select the query from history queries, the filters are applied.
* select a `queryPreview` with filters, and delete the query. If you select the query from history queries, the filters are applied.

[EMP-2427]: https://searchbroker.atlassian.net/browse/EMP-2427?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[EMP-2428]: https://searchbroker.atlassian.net/browse/EMP-2428?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/x-components/src/x-modules/history-queries/store/actions/update-history-queries-with-search-response.action.ts
```diff
@@ -31,7 +31,7 @@ export const updateHistoryQueriesWithSearchResponse: HistoryQueriesXStoreModule[
         const historyQuery = state.historyQueries[indexOfHistoryQuery];
         const isCurrentSessionHistoryQuery = historyQuery.timestamp > state.sessionTimeStampInMs;
         if (!isCurrentSessionHistoryQuery || historyQuery.totalResults == null || searchResponse) {
-          const filters = createHistoryQueriesFiltersList(searchResponse.request.filters);
+          const filters = createHistoryQueriesFiltersList(searchResponse.facets);
```
CachedaCodes, line 34: This is sketchy. The search response filters aren't selected, but this isn't the actual search response, this is the search response rebuilt from the fields in the search module state.
The search module state changes because of the facets service, which is the thing that changes the filters to selected when they arrive from the search response.

This approach might not always work. This action is called from the wiring due to a search module emitter event and when that happens the filters are unselected. Is working now because the filters service happens to process the filters before the wiring of the history queries module receives the event to call the action.

It's worth checking it a little bit further and check if it's possible that this action runs before the facets service processes the filters

packages/x-components/src/x-modules/history-queries/store/__tests__/actions.spec.ts
```diff
@@ -369,9 +391,9 @@ describe('testing history queries module actions', () => {
     });
 
     // eslint-disable-next-line max-len
-    it('updates a history query if search response change because a filter is selected', async () => {
+    it('updates a history query when the search response changes because a filter is selected', async () => {
       gato.totalResults = 50;
-      const selectedFilters = Object.values(requestFilters)[0];
+      const selectedFilters = createHistoryQueriesFiltersList(responseFacets);
```
CachedaCodes, line 396: You're asking the function that you are testing what it should return. `selectedFilters` should be a hardcoded value of what the history query selected filters should be given the request you're dispatching the action with.

CachedaCodes (changes requested):
This flow results in the history query not applying the filters:
Open X -> search query -> select filter -> click "clear" in searchbar -> select new HQ

https://github.com/empathyco/x/assets/7124620/ccf75519-66ef-4999-b040-0a0010287c04

packages/x-components/src/x-modules/history-queries/store/actions/update-history-queries-with-search-response.action.ts
```diff
@@ -46,20 +49,29 @@ export const updateHistoryQueriesWithSearchResponse: HistoryQueriesXStoreModule[
...
+  }
+
+  return Object.values(requestFilters).flatMap(filters =>
+    filters.map(requestFilter => ({
+      ...requestFilter,
+      ...responseFacets
+        .flatMap(facet => facet.filters)
+        .find(filter => filter.id === requestFilter.id)
+    }))
+  );
```
CachedaCodes, line 76: You're overriding the fields of the `requestFilter` (where you have the `selected` as `true`) with the response filter value (where you can't know for sure that the filter is correctly selected). You would have to deconstruct the response filter before the `requestFilter`.

In any case, this algorithm can be very slow (you're always iterating all the list of filters from the response) and you are not taking into account possibility that you might not find any matching filter in the response four the request filters.

This is the best I've come up with. I think that it's a good starting point to see if it works well

```suggestion
  return Object.entries(requestFilters).flatMap(([facetId, facetFilters]) => {
    const matchingFacet =
      facetId !== '__unknown__' ? responseFacets.find(facet => facet.id === facetId) : null;

    return facetFilters.reduce<Filter[]>((accFilters, requestFilter) => {
      const matchingFilter = matchingFacet
        ? matchingFacet.filters.find(filter => filter.id === requestFilter.id)
        : responseFacets
            .flatMap(facet => facet.filters)
            .find(filter => filter.id === requestFilter.id);

      if (matchingFilter) {
        accFilters.push({ ...matchingFilter, ...requestFilter });
      }

      return accFilters;
    }, []);
  });
```

CachedaCodes (changes requested):

packages/x-components/src/x-modules/history-queries/store/__tests__/actions.spec.ts
```diff
@@ -369,9 +401,8 @@ describe('testing history queries module actions', () => {
     });
 
     // eslint-disable-next-line max-len
-    it('updates a history query if search response change because a filter is selected', async () => {
+    it('updates a history query when the search response changes because a filter is selected', async () => {
```
CachedaCodes, line 404: It's worth testing with facets with `__unknown__` as id. Create a new test for it and where done 👍

CachedaCodes (changes requested):

packages/x-components/src/x-modules/history-queries/store/__tests__/actions.spec.ts
```diff
@@ -369,9 +401,26 @@ describe('testing history queries module actions', () => {
...
+      });
+      expectHistoryQueriesToEqual([{ ...gato, totalResults, selectedFilters }, perro]);
+    });
+
+    // eslint-disable-next-line max-len
+    it('updates a history query when the search response changes although the facet id is unknown', async () => {
       gato.totalResults = 50;
-      const selectedFilters = Object.values(requestFilters)[0];
       resetStateWith({ historyQueries: [gato, perro] });
       await store.dispatch('updateHistoryQueriesWithSearchResponse', {
```
CachedaCodes, line 425: Testing this doesn't make much sense to me. The response shouldn't return facets with id `__unknown__`. The backend should know the id of the facet and `__unknown__` is a key used only in the front to identify a filter which we don't know the facet id.

The test I was suggesting was the behaviour of the action when a Query preview was selected. Selecting a QP creates a HQ of that request, but the QP creates the filters with facet id `__unknown__`. The test should check that the filters map correctly when the `requestFilters` have an unknown facet id.

CachedaCodes (changes requested):

CachedaCodes (approved):

PR #1333, opened 2023-10-19
fix(history-queries): selected filters now working for legacy history queries saved in storage without them

CachedaCodes:
## Motivation and context
Legacy HQ (history queries stored in the local storage before the new `selectedFilters` feature was created) caused an error because clicking on them resulted in the utility `groupBy` being called with `undefined`.

## Type of change

- [X] `Main`
- [ ] Other. Specify:

## How has this been tested?
Check if you have any legacy HQ in the local storage or create one by removing the `selectedFilters` field from an existing one and the select it.

### DIFF
--- a/packages/x-components/src/x-modules/history-queries/store/actions/load-history-queries-from-browser-storage.action.ts
+++ b/packages/x-components/src/x-modules/history-queries/store/actions/load-history-queries-from-browser-storage.action.ts
@@ -13,7 +13,6 @@ import { HistoryQueriesXStoreModule } from '../types';
 // eslint-disable-next-line max-len
 export const loadHistoryQueriesFromBrowserStorage: HistoryQueriesXStoreModule['actions']['loadHistoryQueriesFromBrowserStorage'] =
   ({ commit, getters }) => {
-    const historyQueries =
-      localStorageService.getItem<HistoryQuery[] | HistoryQuery>(getters.storageKey) ?? [];
-    commit('setHistoryQueries', [historyQueries].flat());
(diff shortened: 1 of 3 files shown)


packages/x-components/src/x-modules/history-queries/store/actions/load-history-queries-from-browser-storage.action.ts
```diff
@@ -13,6 +13,7 @@ import { HistoryQueriesXStoreModule } from '../types';
 // eslint-disable-next-line max-len
 export const loadHistoryQueriesFromBrowserStorage: HistoryQueriesXStoreModule['actions']['loadHistoryQueriesFromBrowserStorage'] =
   ({ commit, getters }) => {
-    const historyQueries = localStorageService.getItem<HistoryQuery[]>(getters.storageKey) ?? [];
-    commit('setHistoryQueries', historyQueries);
+    const historyQueries =
+      localStorageService.getItem<HistoryQuery[] | HistoryQuery>(getters.storageKey) ?? [];
+    commit('setHistoryQueries', [historyQueries].flat());
```
diegopf, line 18: I don't understand the need for this change. AFAIK we always store the history queries as an [array](https://github.com/empathyco/x/blob/main/packages/x-components/src/x-modules/history-queries/store/actions/add-query-to-history.action.ts#L34+L42) So I don't think we need the flatting.

diegopf (changes requested):

diegopf (approved):

PR #1358, opened 2023-11-20
ci: remove lint and test steps

diegopf:
This PR removes the test and lint steps from the workflow for creating a new stable release. Currently, these steps run on every push and PR. Since the prepare release workflow is launched against the main branch, where these steps have already been executed, there's no need to run them again, especially considering this workflow will only create a PR and this new PR will execute them.

### DIFF
--- a/.github/workflows/prepare-release.yml
+++ b/.github/workflows/prepare-release.yml
@@ -22,12 +22,6 @@ jobs:
       - name: install dependencies
         run: pnpm install
 
-      - name: lint packages
-        run: pnpm run lint
-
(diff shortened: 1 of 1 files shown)


annacv (approved):
good catch!

herrardo (approved):
Should we specify...if we deactivate this, that to properly launch the prepare-release action, the last chore commit should have finished the pipeline?

herrardo (dismissed):

herrardo (approved):

annacv (approved):

PR #1360, opened 2023-11-21
Set cache on queries preview results

annacv:
[EMP-2974](https://searchbroker.atlassian.net/browse/EMP-2974) 

# Pull request template
 
We could face a memory leak if we use the Query Results Preview in the Empathize to show something as an Instant Search as we showed in the Sprint Review, in that case, we wouldn’t want to cache the results.

**CHANGES APPLIED**: 
* Add a cache prop and only cache the results accordingly (`saveCache` prop set to false by default, we will only catch the queries in` BrandRecommendations` to avoid requesting twice).
    * `saveCache` is set at true only with brand recommendations, because is the only feature where we want to save the queries. 

* Only trigger request if there are no cached results
    * When the `queryPreview` component is created we also check if the query is cached before triggering the request. If so, we emit `load` to render the queryPreview, otherwise, we emit `QueryPreviewRequestUpdated`. 

**PENDING TO RESOLVE:** --> [New task](https://searchbroker.atlassian.net/browse/EMP-3010): Note that in the current implementation, we are only saving queries. As a second step, we should also support filters, but then it should be the hash of query+filters+any other property that we might need or think about another workaround.

**PENDING TO RESOLVE:** --> Maybe we should deal with this in a separated PR
* Animation effect now it is linked to requests, DOM renders as soon as the response is received, by catching queries we lose that effect.
* Adding an animation component doesn’t solve the issue, it causes max call stack if used as a default in prop, or the lists are not rendered if passed as a computed prop. If this worked we could use the StaggeredFadeAndSlide when the caché is activated and ul when not. Before adding extraParams & filters, it worked only as a computed as follows: 
	```
        public get animation(): Vue | string | VueClass<Vue> {
  	  return StaggeredFadeAndSlide;
	} ```
* Adding a debounce when we emit load doesn’t solve the issue, then only one query preview is rendered and is kept in the loading state

## Motivation and context
<!--Include information on the purpose of the change and any background information that may help. Why is this change required? What problem does it solve? -->
Now, We are requesting the Query Preview search requests every time that they are rendered, so, every time that the user deletes the query.
<!-- List any dependencies that are required for this change. If the change fixes an **open** issue, please link to the issue here.-->

- [x] Open issue. If applicable, link: [EMP-2867](https://searchbroker.atlassian.net/browse/EMP-2867)

## Type of change
<!-- Indicate the type of change involved in the PR -->

<!-- Although this may seem obvious, please include the destination branch as an extra check to ensure your PR targets the right branch.-->
- [x] `Main`
- [ ] Other. Specify:

## How has this been tested?

<!-- Please describe in detail how you tested your changes. Include details of your testing environment, the test cases used, and the tests you ran to see how your change affects other areas of the code, etc.-->
**TEST:**
* Open vue dev tools and check if brand recommendations are being saved in queryPreview state.
* Search for `battery` (has semantics) and check if brand recommendations still keep saving and if semantic queries are added to `queriesPreview` list.
* Clean the query in the search box and check the `queriesPreview` list. It should contain only the brand recommendations queries, semantic-queries have `saveCache` false, which means that these queries should be removed.

## Checklist:

[EMP-2867]: https://searchbroker.atlassian.net/browse/EMP-2867?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[EMP-2974]: https://searchbroker.atlassian.net/browse/EMP-2974?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


annacv:
Tests should be fixed after the implementation is validated 👀

herrardo (changes requested):

packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
```diff
@@ -222,21 +235,26 @@
...
+      } else {
+        this.emitQueryPreviewRequestUpdated(this.queryPreviewRequest);
+      }
     }
 
     /**
      * Cancels the (remaining) requests when the component is destroyed
      * via the `debounce.cancel()` method.
-     * If the prop 'clearOnDestroy' is set to true, it also removes the QueryPreview
+     * If the prop 'saveCache' is set to true, it also removes the QueryPreview
```
annacv, line 250: Hi! 👀 
```suggestion
     * If the prop 'saveCache' is set to false, it also removes the QueryPreview
```

packages/x-components/src/x-modules/queries-preview/components/__tests__/query-preview.spec.ts
```diff
@@ -101,8 +103,23 @@ describe('query preview', () => {
     expect(getXComponentXModuleName(wrapper.vm)).toBe('queriesPreview');
   });
 
+  it('no sends the `QueryPreviewRequestUpdated` event', () => {
```
annacv, line 106: Hi! I think it would be nice to test that `load` is being emitted with the `queryPreviewInfo.query`. I would update the test definition this way, to point to the prop & the alternative behavior that provoques:
```suggestion
  it('no sends the `QueryPreviewRequestUpdated` event if saveCache is true, but emits load', () => {
```

annacv (commented):
🎉

packages/x-components/src/x-modules/queries-preview/components/__tests__/query-preview.spec.ts
```diff
@@ -101,8 +103,25 @@ describe('query preview', () => {
     expect(getXComponentXModuleName(wrapper.vm)).toBe('queriesPreview');
   });
 
+  it('no sends the `QueryPreviewRequestUpdated` event if saveCache is true, but emits load', () => {
```
CachedaCodes, line 106: The descriptions of the tests have to make sense gramatically with the 'it', so it reads like "it tests this thing"

```suggestion
  it('does not send the `QueryPreviewRequestUpdated` event if saveCache is true, but emits load', () => {
```

packages/x-components/src/x-modules/queries-preview/components/__tests__/query-preview.spec.ts
CachedaCodes, line 1: I'm missing a test checking that the query preview is still in the state after the component is destroyed with `saveCache` to true.

packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
```diff
@@ -208,6 +208,19 @@
...
+     * in the queriesPreview in the state.
+     *
+     * @internal
+     *
+     * @returns True if the query has been saved.
+     */
+    protected get isSavedQuery(): boolean {
+      const previewItemQuery = this.queryPreviewInfo.query;
+      return !!this.previewResults[previewItemQuery];
+    }
```
CachedaCodes, line 222: This property is used only once in the created hook (so is a dynamic property that doesn't need to be dynamic). You can delete it and make the check in the `created` directly.

packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
```diff
@@ -222,21 +235,26 @@
           }
         }
       );
-      this.emitQueryPreviewRequestUpdated(this.queryPreviewRequest);
+
+      // If the query has been saved it will emit load instead of the emitting the updated request.
+      if (this.isSavedQuery && this.saveCache) {
+        this.$emit('load', this.queryPreviewInfo.query);
```
CachedaCodes, line 241: Not sure if this is right. If the query preview errored in the first request, and it was saved with an error status in the cache, you are now emitting an event saying it's all good, which is not correct.

You have to check the status of the query preview in the cache too.

CachedaCodes (changes requested):

packages/x-components/src/x-modules/queries-preview/components/__tests__/query-preview.spec.ts
```diff
@@ -103,9 +103,10 @@ describe('query preview', () => {
     expect(getXComponentXModuleName(wrapper.vm)).toBe('queriesPreview');
   });
 
-  it('no sends the `QueryPreviewRequestUpdated` event if saveCache is true, but emits load', () => {
+  // eslint-disable-next-line max-len
+  it('does not send the `QueryPreviewRequestUpdated` event if saveCache is true, but emits load', () => {
```
annacv, line 107: Change `saveCache` by `persistInCache` in test definition

packages/x-components/src/x-modules/queries-preview/components/__tests__/query-preview.spec.ts
```diff
@@ -119,9 +120,24 @@ describe('query preview', () => {
     expect(wrapper.emitted('load')?.[0]).toEqual(['shoes']);
   });
 
+  it('does not remove the query before destroy the component', () => {
```
annacv, line 123: before destroying or in the beforeDestroy component's hook

packages/x-components/src/x-modules/queries-preview/components/__tests__/query-preview.spec.ts
```diff
@@ -119,9 +120,24 @@ describe('query preview', () => {
...
+      queryPreviewInfo: {
+        query: 'shoes',
+        extraParams: { directory: 'Magrathea' },
+        filters: ['fit:regular']
+      }
+    });
+
+    wrapper.destroy();
+    expect(wrapper.emitted('load')?.length).toBe(1);
+    expect(wrapper.emitted('load')?.[0]).toEqual(['shoes']);
```
annacv, line 135: here I think what should done is to check the state & see that there are still query preview there

packages/x-components/src/x-modules/queries-preview/components/__tests__/query-preview.spec.ts
annacv: sorry, and add the value of the cache prop in the test statement to make sure we are testing cache is working ok

packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
```diff
@@ -222,21 +222,28 @@
           }
         }
       );
-      this.emitQueryPreviewRequestUpdated(this.queryPreviewRequest);
+      const previewItemQuery = this.queryPreviewInfo.query;
+      const isSavedQuery = !!this.previewResults[previewItemQuery];
+
+      // If the query has been saved it will emit load instead of the emitting the updated request.
+      if (isSavedQuery && this.persistInCache) {
```
CachedaCodes, line 229: You're still not checking the status of the queryPreview request to decide if you should use the cache or not

packages/x-components/src/x-modules/queries-preview/components/__tests__/query-preview.spec.ts
```diff
@@ -101,8 +103,41 @@ describe('query preview', () => {
...
+      }
+    });
+
+    jest.advanceTimersByTime(0); // Wait for first emission.
+    expect(queryPreviewRequestUpdatedSpy).toHaveBeenCalledTimes(0);
+    expect(wrapper.emitted('load')?.length).toBe(1);
+    expect(wrapper.emitted('load')?.[0]).toEqual(['shoes']);
+  });
+
+  it('does not remove the query before destroying the component', () => {
```
CachedaCodes, line 123: We're still not testing well the functionality. We have to test both cases, `persistInCache=true` and `persistInCache=false`

CachedaCodes (changes requested):

packages/x-components/src/x-modules/queries-preview/components/__tests__/query-preview.spec.ts
```diff
@@ -101,8 +103,77 @@ describe('query preview', () => {
...
+            selected: true
+          }
+        ]
+      },
+      origin: undefined,
+      query: 'shoes',
+      rows: 24
+    });
+  });
+
```
CachedaCodes, line 173: We're testing a whole flow instead of just the component. We shouldn't care about what the app does when we emit the event to remove the query preview, we just want to tests that it does emit the event.
This way, if the store implementation fails or changes for whatever reason, these tests aren't affected.

packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
```diff
@@ -222,21 +222,33 @@
...
      * from the state when the component is destroyed.
      *
      * @internal
      */
     protected beforeDestroy(): void {
       this.emitQueryPreviewRequestUpdated.cancel();
-
-      if (this.clearOnDestroy) {
+      if (!this.persistInCache) {
         this.$x.emit('QueryPreviewUnmountedHook', this.queryPreviewInfo.query, {
```
CachedaCodes, line 252: I hate this event name xD. Can you change it to `NonCacheableQueryPreviewUnmounted`? 🙏

CachedaCodes (approved):

PR #1370, opened 2023-12-19
feat: update to rollup 4

diegopf:
This PR updates rollup to the latest version which will help us in the transition from rollup to vite in the future.

The biggest changes are:

1. [Convert configuration files to ES modules](https://rollupjs.org/migration/#using-configuration-files)
2. [Changes to rollup options](https://github.com/rollup/rollup/blob/master/CHANGELOG.md#changes-to-rollup-options)
3. [Different file name pattern](https://github.com/rollup/rollup/pull/4565)

In order to test this changes you should make a `pnpm install --frozen-lockfile` which will trigger the build for every package in the monorepo with the new rollup version. Afterwards, you can use the archetype and install locally the new versions and see if everything is still working as expected.


packages/x-components/package.json
```diff
@@ -30,7 +30,7 @@
     "./tests/**/*",
     "**/*.scss",
     "**/*.scss.js",
-    "**/*vue_styles.*.vue.js",
+    "**/*vue[0-9].js",
```
diegopf, line 33: Due to how rollup generates now the output filenames when preserving modules, we need to adjust the sideEffect field in the package.json so bundlers do not tree shake styles of vue components.

packages/x-archetype-utils/rollup.config.mjs
annacv, line 1: Hi @diegopf, 
We can apply some changes to avoid the warnings in the build process: 
-  `preserveModules:true`: rm from `input` level & set it inside each `output` object
- point to external resources: `external: ['@empathyco/x-deep-merge', '@empathyco/x-utils', 'vue-i18n']`

Captura de pantalla 2023-12-20 a les 17 01 12

packages/x-tailwindcss/rollup.config.mjs
annacv, line 1: iep: here also:
-  `preserveModules:true`: rm from `input` level & set it inside each `output` object

packages/x-types/rollup.config.mjs
annacv, line 1: iep: the same: 
-  `preserveModules:true`: rm from `input` level & set it inside each `output` object

packages/x-tailwindcss/demo/rollup.config.mjs
annacv, line 1: Here too:
-  `preserveModules:true`: rm from `input` level & set it inside each `output` object

annacv:
@diegopf great job!!

Tested in archetype & everything is working fine. The warnings mentioned don't break the build process.

Here is just a minor doubt I have:
- I think there is an error [here](https://github.com/empathyco/x/blob/main/packages/x-components/build/rollup-plugins/design-system.rollup-plugin.ts#L59): it refers to [lintcss](http://csslint.net/), but checking [rollup-plugin-styles](https://www.npmjs.com/package/rollup-plugin-styles#emitting-processed-css) documentation I think it should be `litcss` .

packages/x-tailwindcss/demo/rollup.config.mjs
annacv, line 1: iep! We forgot this one!
`output: [{ format: 'esm', dir: resolve(__dirname, '../showcase'), preserveModules: true }],`

annacv (dismissed):
🎉 

Tested x-components library
Tested x-tailwind library
Tested all inside archetype

Just found some warns in x-types, which are not new nor related to this task (so these should be handled in another task -small-)

All fine 💯

lauramargar (dismissed):
🚀

annacv (approved):

PR #1392, opened 2024-01-15
feat: allow queries preview with same query but different filters or params

lauramargar:
# Pull request template
 
In [the current implementation](https://github.com/empathyco/x/pull/1360), we are only saving queries. As a second step, we should also support filters and extraparams, but then we should save the hash of query+filters.

We have used [`js-md5`](https://www.npmjs.com/package/js-md5) hash to make an id to differentiate the same queries but with different filters applied.

There was a problem when dealing with multiple instances of the `QueryPreview` component with the same query hash. In that case, the order of mounting and unmounting the components matter and it could lead to a scenario where a component is mounted but afterwards, the other one with the same query hash is unmounted and both of them will be gone from the DOM.

We have applied the following **SOLUTION**:

Refactor the events emitted from `QueryPreview`. Remove `NonCacheableQueryPreviewUnmounted` and add `QueryPreviewMounted: string;` `QueryPreviewUnmounted: { query: string, cache: boolean };`

Enhance the `QueryPreviewItem` from the state with a field 
```
  /** The number of instances showing the query preview*/
  instances: number;
```

The `fetchAndSaveQueryPreview` action will set the instances field to 1 since it will be the first instance of a `QueryPreview` using this query hash.

Every `QueryPreview` instance when is created and picks the data from the cache will emit `QueryPreviewMounted` so the instances count is increased. Every instance of `QueryPreview` will emit `QueryPreviewUnmounted` so the instances count for that query hash will be decreased and if the cache is set to false, then the query will be removed safely from the state since there will be no component showing that data.

## Motivation and context

- [X] Open issue. If applicable, link: [EMP-3079](https://searchbroker.atlassian.net/browse/EMP-3079)

## Type of change

- [X] Breaking change (fix or feature that causes existing functionality to not work as expected)

- [X] `Main`
- [ ] Other. Specify:

## How has this been tested?

You can test 2 scenarios:

1. Same query + different filters
2. Same query in different instances

In **scenario 1** --> you can add in `Home.vue` file this new query preview in `queriesPreviewInfo` : 
```
{
  query: 'summer dress',
  filters: ['brand:marni']
}
```
You should see 2 carrousels with 'summer dress' as the title but with different results.

In **scenario 2** --> add in the same place as the previous scenario the following query preview:
```
{
  query: 'charger'
}
```
Then search for `battery` in the search box. This query has the query `charger` as a semantic query. 
As we had saved `charger` in the state from the brand recommendations, in semantic queries should be loaded from the state instead of being requested again.

Once we clear the query in the search box, we should see that `charger` keeps saved in the query preview state and is displayed as a brand recommendation.

## Checklist:

[EMP-3079]: https://searchbroker.atlassian.net/browse/EMP-3079?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


annacv (dismissed):
Also tested with:`{ query: 'wallet', filters: ['brand:valentino']},` as a queryPreview + searching for `charger` to check that Semantic Queries will show results for the query `wallet` without being filtered.

packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
```diff
@@ -151,6 +152,13 @@
     @Inject({ default: undefined })
     protected location?: FeatureLocation;
 
+    /**
+     * Query Preview key converted into a unique id.
+     *
+     * @internal
+     */
+    public queryOfQueryPreview = getHashFromQueryPreviewInfo(this.queryPreviewInfo);
```
diegopf, line 160: I would rename it from `queryOfQueryPreview` to something like `queryOfQueryPreviewHash` or similar so it is clear that we are dealing with a hash.

packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
```diff
@@ -223,12 +235,16 @@
         }
       );
 
-      const previewItemQuery = this.queryPreviewInfo.query;
+      const previewItemQuery = this.queryOfQueryPreview;
```
diegopf, line 238: I think we can skip the `previewItemQuery` variable and just use `this.queryOfQueryPreview`

packages/x-components/src/x-modules/queries-preview/components/query-preview.vue
```diff
@@ -223,12 +235,16 @@
...
       const cachedQueryPreview = this.previewResults[previewItemQuery];
 
       // If the query has been saved it will emit load instead of the emitting the updated request.
-      if (cachedQueryPreview?.status === 'success' && this.persistInCache) {
-        this.$emit('load', this.queryPreviewInfo.query);
+      if (cachedQueryPreview?.status === 'success') {
+        this.$emit('load', previewItemQuery);
+        this.$x.emit('QueryPreviewMounted', this.queryOfQueryPreview, {
```
diegopf, line 244: Following the above comment, here we are using `previewItemQuery` and in the next line `this.queryOfQueryPreview` but both are the same. We should just use `this.queryOfQueryPreview` everywhere.

packages/x-components/src/x-modules/queries-preview/events.types.ts
```diff
@@ -10,10 +10,15 @@ import { QueryPreviewInfo } from './store/index';
...
-   * Any property of the queries preview request has changed.
+   * Any property of cacheable queries preview request has changed.
    * Payload: The new {@link @empathyco/x-types#SearchRequest | request}.
    */
   QueryPreviewRequestUpdated: SearchRequest;
+  /**
+   * Any property of no cacheable queries preview request has changed.
+   * Payload: The new {@link @empathyco/x-types#SearchRequest | request}.
+   */
+  QueryPreviewRequestUpdatedForNoCache: SearchRequest;
```
diegopf, line 21: do we need this event?

packages/x-components/src/x-modules/queries-preview/utils/get-hash-from-query-preview.ts
diegopf, line 1: where are the tests for these utils? 🙃 We should add some unit tests

diegopf (changes requested):

packages/x-components/src/x-modules/queries-preview/utils/get-hash-from-query-preview.ts
```diff
@@ -0,0 +1,40 @@
...
+ * @param queryPreviewInfo - The {@link QueryPreviewInfo | QueryPreviewInfo} of a QueryPreview.
+ * @returns A unique id that will be used as a key to check the QueryPreview in the state.
+ */
+export const getHashFromQueryPreviewInfo = (queryPreviewInfo: QueryPreviewInfo): string => {
+  let queryPreviewFilters = '';
+  if (queryPreviewInfo.filters) {
+    queryPreviewInfo.filters.forEach(filter => {
+      queryPreviewFilters = queryPreviewFilters.concat('-' + filter);
+    });
+  }
```
CachedaCodes, line 37: You can refactor to this:

```suggestion
  const queryPreviewFilters = queryPreviewInfo.filters ? queryPreviewInfo.filters.join('-') : '';
```

packages/x-components/src/x-modules/queries-preview/utils/get-hash-from-query-preview.ts
```diff
@@ -0,0 +1,40 @@
...
+export const getHashFromQueryPreviewItem = (queryPreview: QueryPreviewItem): string => {
+  let queryPreviewFilters = '';
+  if (queryPreview.request.filters) {
+    const queryPreviewFiltersKeys = Object.keys(queryPreview.request.filters);
+    queryPreviewFiltersKeys.forEach(key => {
+      queryPreview.request.filters![key].forEach(filter => {
+        queryPreviewFilters = queryPreviewFilters.concat('-' + filter.id.toString());
+      });
+    });
+  }
```
CachedaCodes, line 20: You can refactor this to:

```suggestion
  const queryPreviewFilters = Object.values(queryPreview.request.filters)
  .flat()
  .map(filter => filter.id.toString())
  .join('-');
```

packages/x-components/src/x-modules/queries-preview/utils/__tests__/utils.spec.ts
CachedaCodes, line 1: Make the QP in the tests have filters. Half of the implementation is not being tested right now

packages/x-components/src/x-modules/queries-preview/components/query-preview-list.vue
```diff
@@ -87,10 +88,10 @@
      * @internal
      */
     protected get renderedQueryPreviews(): QueryPreviewInfo[] {
-      return this.queriesPreviewInfo.filter(
-        ({ query }) =>
-          this.queriesStatus[query] === 'success' || this.queriesStatus[query] === 'loading'
-      );
+      return this.queriesPreviewInfo.filter(item => {
+        const query = getHashFromQueryPreviewInfo(item);
```
CachedaCodes, line 92: Rename to `queryPreviewHash`

packages/x-components/src/x-modules/queries-preview/components/__tests__/query-preview-list.spec.ts
```diff
@@ -86,7 +86,9 @@ describe('testing QueryPreviewList', () => {
     expect(queryPreviews.at(1).text()).toEqual('jeans - Sick jeans');
   });
 
-  it('hides queries with no results', async () => {
+  // TODO Uncomment when the 'error' event is fixed
```
CachedaCodes, line 89: If there's a task already, add the task ID here

CachedaCodes (changes requested):

packages/x-components/src/x-modules/queries-preview/store/module.ts
```diff
@@ -26,34 +26,37 @@ export const queriesPreviewXStoreModule: QueriesPreviewXStoreModule = {
...
     },
     setSelectedQueryPreview(state, selectedQueryPreview) {
       state.selectedQueryPreview = selectedQueryPreview;
     },
     setConfig,
     mergeConfig,
-    addQueryPreviewInstance(state, query: string) {
-      if (state.queriesPreview[query]) {
-        state.queriesPreview[query].instances += 1;
+    addQueryPreviewInstance(state, queryPreviewHash: string) {
```
CachedaCodes, line 46: No need to specify the type in `queryPreviewHash: string`, it's already being done in the types file, remove the `string`.

packages/x-components/src/x-modules/queries-preview/store/module.ts
```diff
@@ -26,34 +26,37 @@ export const queriesPreviewXStoreModule: QueriesPreviewXStoreModule = {
...
+      if (state.queriesPreview[queryPreviewHash]) {
+        state.queriesPreview[queryPreviewHash].instances += 1;
       }
     },
-    removeQueryPreviewInstance(state, { query, cache }: { query: string; cache: boolean }) {
-      if (state.queriesPreview[query]) {
-        state.queriesPreview[query].instances -= 1;
+    removeQueryPreviewInstance(
+      state,
+      { queryPreviewHash, cache }: { queryPreviewHash: string; cache: boolean }
```
CachedaCodes, line 53: Same here, you don't need the type in ` { queryPreviewHash, cache }: { queryPreviewHash: string; cache: boolean }`, leave it as ` { queryPreviewHash, cache }`

packages/x-components/src/x-modules/queries-preview/store/types.ts
```diff
@@ -111,8 +111,14 @@ export interface QueriesPreviewMutations extends ConfigMutations<QueriesPreviewS
    * @param selectedQueryPreview - The selected query preview to save to the state.
    */
   setSelectedQueryPreview(selectedQueryPreview: QueryPreviewInfo | null): void;
-  addQueryPreviewInstance(query: string): void;
-  removeQueryPreviewInstance({ query, cache }: { query: string; cache: boolean }): void;
+  addQueryPreviewInstance(queryPreviewHash: string): void;
```
CachedaCodes, line 114: Add docs

packages/x-components/src/x-modules/queries-preview/store/types.ts
```diff
@@ -111,8 +111,14 @@ export interface QueriesPreviewMutations extends ConfigMutations<QueriesPreviewS
...
+  removeQueryPreviewInstance({
```
CachedaCodes, line 115: Add docs

packages/x-components/src/x-modules/queries-preview/components/__tests__/query-preview-list.spec.ts
CachedaCodes: Reference task EMP-3402

CachedaCodes (changes requested):
Fix the conflicts with main too

CachedaCodes (approved):

PR #1393, opened 2024-01-15
fix: refactor snippet-config-extra-params to use watchers preventing false event firing

CachedaCodes:
## Motivation and context
The component `SnippetConfigExtraParams` was causing the firing of the event `ExtraParamsProvided` when the snippet config changed, even if the extra params didn't actually change. This PR changes the logic from a computed property to two watchers, preventing the bug.

## Type of change

- [X] `Main`
- [ ] Other. Specify:

## How has this been tested?
Check that updating a parameter in the snippet config with `InterfaceX.setSnippetConfig` that is exclueded from the extra params doesn't fire `ExtraParamsProvided`.


diegopf (commented):
what about adding a test for this scenario?

PR #1402, opened 2024-02-01
feat: create useState composable

lauramargar:
[EMP-3458](https://searchbroker.atlassian.net/browse/EMP-3458)

# Pull request template
 
Currently, we have a decorator [@state](https://github.com/empathyco/x/blob/main/packages/x-components/src/components/decorators/store.decorators.ts#L17) that allows a component to read a piece of the state. We need to create a new composable that provides the same functionality. 

We have created a reactive composable that returns a dictionary of the state of a list of params in a specific module.

## Motivation and context

- [X] Open issue. If applicable, link: [EMP-3458](https://searchbroker.atlassian.net/browse/EMP-3458)

## Type of change

- [X] `Main`
- [ ] Other. Specify:

## How has this been tested?

To test it, I have created a `test` component where I use the `useState` composable to save the value in a computed variable. Then I updated (with the browser console) the store to test if the reactivity was working correctly.

**Test component:**
```
<template>
  <div>
    <span>{{ queriesPreview.params.value }}</span>
    <span>{{ queriesPreview.config.value }}</span>
  </div>
</template>

<script lang="ts">
  import { defineComponent } from 'vue';
  import { useStore } from '../composables';
  export default defineComponent({
    setup() {
      const { useState } = useStore();
      const queriesPreview = useState('queriesPreview', ['params', 'config']);
      console.log(queriesPreview);
      return {
        queriesPreview
      };
    }
  });
</script>
```
**Update store:**

`InterfaceX.setSnippetConfig({store: "Gijon"})`

**(display test component in the demo)**

## Checklist:

[EMP-3458]: https://searchbroker.atlassian.net/browse/EMP-3458?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[EMP-3458]: https://searchbroker.atlassian.net/browse/EMP-3458?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/x-components/src/composables/use-state.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+export function useState<Module extends XModuleName, Path extends keyof ExtractState<Module>>(
+  module: Module,
+  path: Path
+): ExtractState<Module>[Path] {
+  const $store = useStore();
+  const state = computed((): ExtractState<Module>[Path] => {
+    return $store.state.x[module][path];
+  });
+  return state.value;
```
CachedaCodes, line 21: Returning the `value` instead of the computed property breaks the reactivity. Better to return the computed property directly.

```suggestion
  return computed(() => $store.state.x[module][path]);
```

packages/x-components/src/composables/use-state.ts
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Generates a computed property which returns the selected state.
+ *
+ * @param module - The {@link XModuleName} of the getter.
+ * @param path - The state path.
+ * @returns The state properties of the module.
+ * @public
+ */
+export function useState<Module extends XModuleName, Path extends keyof ExtractState<Module>>(
```
CachedaCodes, line 13: What was the reason to create a new standalone composable instead of extending the `useStore` one? The task gave the two options, so I'm curious about the reasoning.

CachedaCodes (changes requested):

packages/x-components/src/composables/use-state.ts
lauramargar: No reason, I had forgotten that option.

packages/x-components/src/composables/use-state.ts
lauramargar: Thanks!!

packages/x-components/src/composables/use-store.ts
```diff
@@ -1,13 +1,33 @@
...
+export function useStore<Module extends XModuleName, Path extends keyof ExtractState<Module>>(
+  module?: Module,
+  path?: Path
+): UseStore<Module, Path> {
+  const store = (getCurrentInstance()?.proxy as { $store: Store<any> }).$store;
+  const state = computed(() => store.state.x[module][path]);
+  return {
+    store,
+    state
+  };
```
CachedaCodes, line 24: This would break if both `module` or `path` are not passed to the composable.
Instead of this approach, I'm thinking of something closer to what vueblocks is doing:
https://vueblocks.github.io/vue-use-utilities/guide/vuex/useVuex.html#example

For our case, we should return store always and the composable can receive the module name as optional.
Instead of returning the state directly, we return a `useState` function that receives the module and a list of paths to the state's properties. If the module name was provided when calling the `useStore` composable then it's binded to the `useState` function so it only needs the paths list

CachedaCodes (changes requested):

annacv (dismissed):
🎉

packages/x-components/src/composables/use-store.ts
```diff
@@ -1,13 +1,40 @@
...
+      state[path] = computed(() => store.state.x[module][path]);
+      return state;
+    }, {} as ComputedRef<Dictionary<keyof Path>>);
+  };
+  const useStateWithModule = useState.bind(module);
+  return {
+    store,
+    useState: module ? useStateWithModule : useState
+  };
 }
```
diegopf, line 32: The types for the useStore are wrong. We don't have a ComputedRef of a Dictionary but the other way around. It should be something like:

```ts
type UseStore = {
  store: Store<any>;
  useState: <Module extends XModuleName, Paths extends keyof ExtractState<Module>>(
    module: Module,
    paths: Paths[]
  ) => Dictionary<ComputedRef>;
};

export function useStore(): UseStore {
  const store = (getCurrentInstance()?.proxy as { $store: Store<any> }).$store;
  const useState = <Module extends XModuleName, Paths extends keyof ExtractState<Module>>(
    module: Module,
    paths: Paths[]
  ): Dictionary<ComputedRef> => {
    return paths.reduce<Dictionary<ComputedRef>>((state, path) => {
      state[path as string] = computed(() => store.state.x[module][path]);
      return state;
    }, {});
  };
  return {
    store,
    useState
  };
```

For this first iteration I would skip the use case of passing the module as argument.

diegopf (changes requested):

annacv (dismissed):
It looks ok to me, I tried to find a way to type Wrapper<any> as a VueInstance, but I didn't find a way.
On the other hand, when the composable is used in a template we are getting TS unsafe call, it could be nice to avoid it, still not found the way:
Captura de pantalla 2024-02-12 a les 9 01 26

annacv (approved):
gj!

diegopf (approved):

CachedaCodes (approved):

PR #1404, opened 2024-02-07
Update facets' module when we change preselected filters in the snippetConfig

annacv:
[EMP-3520](https://searchbroker.atlassian.net/browse/EMP-3520)

When we have preselected filters and want to change or update them using the `setSnippetConfig` function, there is a bug: the `PreselectedFilters` component is updated, but the facets' module is not.

To solve this I've defined a computed property to set the preselected filters and a watcher to emit the `PreselectedFiltersProvided` event when it changes.

This component is meant [to select the preselected filters when `UserClickedOpenX` is fired](https://github.com/empathyco/x/blob/main/packages/x-components/src/x-modules/facets/wiring.ts#L232), so, this change won't be instant without initializing the application. If we want to make this change instant we should [select the preselected filters (not only set them, as now) through the facets service](https://github.com/empathyco/x/blob/main/packages/x-components/src/x-modules/facets/service/facets.service.ts#L51) when the update is called. But, if filters are defined as sticky (which is [dutty](https://github.com/empathyco/x-massimodutti-next/blob/main/src/main.ts) use case) we won't see this change instantly.

**TEST** 
**Test in x-components:**
1. Open X and paste this code in the console: 
```
InterfaceX.setSnippetConfig({
  filters: [`brand:adidas`],
});
```
2. Search for `trousers` or `dress`: the preselected filter should be shown as selected & the rendered results should work as expected.
3. Delete the query, close the search engine & change the preselected filters:
```
 InterfaceX.setSnippetConfig({
  filters: [`brand:marni`],
});
```
4. Search for `trousers` or `dress`: the preselected filter has changed & the rendered results should work as expected.

**Test in dutty:**
1. Run a build process in `x-components` & install the version locally in dutty
2. Remove [the watcher](https://github.com/empathyco/x-massimodutti-next/blob/main/src/App.vue#L75) in dutty `App.vue` file
3. Follow the steps defined in [this PR](https://github.com/empathyco/x-massimodutti-next/pull/80)

[EMP-3398]: https://searchbroker.atlassian.net/browse/EMP-3398?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[EMP-3520]: https://searchbroker.atlassian.net/browse/EMP-3520?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


lauramargar:
Great job!! For me it's ok!! 
Only one thing to comment, add the injectionKey `@Inject('snippetConfig')`

annacv:
Hi! Please @lauramargar help me test again on Monday, it seems that is not saving the preselected filters on a second search. Steps: 
1. Set preselected via console
2. Open X & search ('dress')
3. Clear query & search again (I removed dress from histories & then searched again)
4. Filter is not kept in the second search

lauramargar (approved):

packages/x-components/src/x-modules/facets/components/preselected-filters.vue
```diff
@@ -33,16 +33,42 @@
...
-        this.$x.emit('PreselectedFiltersProvided', createRawFilters(preselectedFilters));
+    protected get preselectedFilters(): string[] {
+      return this.snippetConfig?.filters ?? this.filters;
+    }
+
+    /**
+     * Emits the {@link FacetsXEvents.PreselectedFiltersProvided} to save
+     * the provided filters in the state.
+     */
+    protected emitPreselectedFilters(): void {
```
CachedaCodes, line 49: You can put the watcher here and still call the function independently

```suggestion
    @Watch('preselectedFilters')
    protected emitPreselectedFilters(): void {
```

packages/x-components/src/x-modules/facets/components/preselected-filters.vue
```diff
@@ -33,16 +33,42 @@
...
+    }
+
+    /**
+     * Emits the {@link FacetsXEvents.PreselectedFiltersProvided} when the
+     * preselectedFilters computed property changes.
+     */
+    @Watch('preselectedFilters')
+    syncPreselectedFilters(): void {
+      this.emitPreselectedFilters();
+    }
```
CachedaCodes, line 70: You don't need this

packages/x-components/src/x-modules/facets/components/preselected-filters.vue
CachedaCodes, line 1: Add a test that checks that the event is emitted when the snippet config changes.

CachedaCodes (changes requested):

CachedaCodes (approved):

PR #1418, opened 2024-02-20
feature(composable): create utility composable useHybridInject

CachedaCodes:
[EMP-3558](https://searchbroker.atlassian.net/browse/EMP-3558)

## Motivation and context
The regular vue 3 `inject` doesn't work with the `XProvide`, so this utility composable functions as a stepping stone for the migration between the two. It works with any of them and it should be removed when the migration to vue 3 is completed.

## Type of change

- [X] `Main`
- [ ] Other. Specify:

## How has this been tested?
Create a injected test vue component and a provider test vue component. Wrap the injected test component in the provider. Now, the injected test component should use the composable and the provider component should do the providing.
Create the provider component so it uses the `@XProvider` decorator and has a dynamic property that provides to its children (for example, a property that changes with a button). The injected test component should pick it up correctly and change dynamically. Then, do the same but with `defineComponent` and the vue 3 `provide`.

Unit tests not created due to this being a utility that will be removed in the future.

[EMP-3558]: https://searchbroker.atlassian.net/browse/EMP-3558?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/x-components/src/composables/use-hybrid-inject.ts
+++ b/packages/x-components/src/composables/use-hybrid-inject.ts
@@ -1,0 +1,26 @@
+import { computed, ComputedRef, inject } from 'vue';
+
+/**
+ * Function to use a hybrid inject, which allows to inject a value provided by the regular provide
+ * of vue  or by the XProvide decorator.
+ *
+ * @param key - The key of the value to inject.
+ * @param defaultValue - The default value to use if the value is not provided.
+ * @returns The computed value of the injected value.
+ */
+export function useHybridInject<SomeValue>(
+  key: string,
+  defaultValue?: SomeValue
+): ComputedRef<SomeValue | undefined> {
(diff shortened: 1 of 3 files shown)


packages/x-components/src/composables/use-hybrid-inject.ts
```diff
@@ -0,0 +1,27 @@
+import { computed, ComputedRef, inject } from 'vue';
+
+/**
+ * Function to use a hybrid inject, which allows to inject a value provided by the regular provide
+ * of vue  or by the XProvide decorator.
```
annacv, line 5:
```suggestion
 * of vue or by the XProvide decorator.
```

annacv (approved):
gj!

PR #1422, opened 2024-02-21
refactor: Migrate preselected filters component

annacv:
# Pull request template
 
Migrate the `PreselectedFilters` component to Vue 2.7

## Motivation and context

- [x] Open issue. If applicable, link: [EMP-3378](https://searchbroker.atlassian.net/browse/EMP-3378)

## Type of change

- [x] `Main`
- [ ] Other. Specify:

## How has this been tested?

## Checklist:

**TEST**
Check the component is working as expected: 
- Use this code snippet before initialising the library:
```
InterfaceX.setSnippetConfig({
  filters: [`brand:marni`]
});
```
- When X is opened, search for a query and check that the filter is shown as selected.
- Close X, do the same again with another filter & check that the new search applies the new filter.
- Additionally you can use the PreselectedFilters component and pass a filter using the filters prop.

[EMP-3378]: https://searchbroker.atlassian.net/browse/EMP-3378?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/x-components/src/x-modules/facets/components/preselected-filters.vue
```diff
@@ -1,69 +1,81 @@
...
+       *
+       * @public
+       */
+      filters: {
+        type: Array as PropType<string[]>,
+        default: () => []
       }
+    },
+    setup(props, { slots }) {
+      const $x = use$x();
```
CachedaCodes, line 31: use the composable for the bus instead of the $x

packages/x-components/src/x-modules/facets/components/preselected-filters.vue
```diff
@@ -1,69 +1,81 @@
...
+        }
+      };
+
+      /**
+       * Emits the {@link FacetsXEvents.PreselectedFiltersProvided} when the
+       * computed prop changes.
+       */
+      watch(preselectedFilters, () => {
+        emitPreselectedFilters();
+      });
```
CachedaCodes, line 66:
```suggestion
      watch(preselectedFilters, emitPreselectedFilters);
```

packages/x-components/src/x-modules/facets/components/preselected-filters.vue
```diff
@@ -1,69 +1,81 @@
...
+    },
+    setup(props, { slots }) {
+      const $x = use$x();
+
+      /**
+       * Injects {@link SnippetConfig} provided by an ancestor as snippetConfig.
+       *
+       * @internal
+       */
+      const snippetConfig = inject<SnippetConfig | undefined>('snippetConfig');
```
CachedaCodes, line 38:
```suggestion
      const snippetConfig = useHybridInject<SnippetConfig>('snippetConfig');
```

packages/x-components/src/x-modules/facets/components/preselected-filters.vue
```diff
@@ -1,69 +1,81 @@
...
+
+      /**
+       * Gets the provided preselected filters prioritizing the {@link SnippetConfig} over the
+       * filters prop.
+       *
+       * @returns An array of filter's ids.
+       */
+      const preselectedFilters: ComputedRef<string[]> = computed(() => {
+        return snippetConfig?.filters ?? props.filters;
```
CachedaCodes, line 47:
```suggestion
        return snippetConfig?.value?.filters ?? props.filters;
```

CachedaCodes (changes requested):

packages/x-components/src/x-modules/facets/components/preselected-filters.vue
annacv: Hi! Using the `useXBus` composable leads to a Vue warn (injection "location" not found) in this use case, maybe I am using it the wrong way?

CachedaCodes: It's just a warning and the location can be `undefined` anyway, but it makes more sense that the default value for the location is `'none'`, so change the `useXBus` composable so it has a default value for the location.

```
const location = inject<FeatureLocation>('location', 'none');
```

CachedaCodes (changes requested):

packages/x-components/src/x-modules/facets/components/__tests__/preselected-filters.spec.ts
```diff
@@ -24,12 +25,18 @@ function renderPreselectedFilters({
...
       }
     }
   });
 
+  const eventMetadata = {
+    moduleName: null,
+    location: 'none',
+    replaceable: true
+  };
+
```
CachedaCodes, line 39: If you don't care what the metadata is in the tests, you can use `expect.any(Object)`.
Remove this and use that instead.

packages/x-components/src/x-modules/facets/components/__tests__/preselected-filters.spec.ts
```diff
@@ -58,25 +65,30 @@ describe('testing Preselected filters component', () => {
...
+    const { emit, eventMetadata } = renderPreselectedFilters({
       snippetFilters
     });
 
     expect(emit).toHaveBeenCalledTimes(1);
     expect(emit).toHaveBeenCalledWith(
       'PreselectedFiltersProvided',
-      createRawFilters(snippetFilters)
+      createRawFilters(snippetFilters),
+      eventMetadata
```
CachedaCodes, line 76:
```suggestion
      expect.any(Object)
```

packages/x-components/src/x-modules/facets/components/preselected-filters.vue
```diff
@@ -1,69 +1,83 @@
...
+      // eslint-disable-next-line @typescript-eslint/unbound-method
+      const { emit } = useXBus();
+
+      /**
+       * Injects {@link SnippetConfig} provided by an ancestor as snippetConfig
+       * and sets is as a ref to get synced when it changes.
+       *
+       * @internal
+       */
+      const snippetConfig = ref(useHybridInject<SnippetConfig>('snippetConfig', baseSnippetConfig));
```
CachedaCodes, line 42: `baseSnippetConfig` is not the default snippet config, it's a mock snippet config used for the test views, shouldn't be used for default values.

```suggestion
      const snippetConfig = ref(useHybridInject<SnippetConfig>('snippetConfig'));
```

packages/x-components/src/x-modules/facets/components/preselected-filters.vue
```diff
@@ -1,69 +1,83 @@
...
+
+      /**
+       * Gets the provided preselected filters prioritizing the {@link SnippetConfig} over the
+       * filters prop.
+       *
+       * @returns An array of filter's ids.
+       */
+      const preselectedFilters: ComputedRef<string[]> = computed(() => {
```
CachedaCodes, line 50: Better to infer the type like this than force it.

```suggestion
      const preselectedFilters = computed<string[]>(() => {
```

packages/x-components/src/x-modules/facets/components/preselected-filters.vue
```diff
@@ -1,69 +1,83 @@
...
+       */
+      watch(preselectedFilters, emitPreselectedFilters);
+
+      /**
+       * Emits the {@link FacetsXEvents.PreselectedFiltersProvided} when the
+       * component is mounted.
+       */
+      onMounted(() => {
+        emitPreselectedFilters();
+      });
```
CachedaCodes, line 76: We can get rid of the `onMounted` hook altogether by using an immediate watcher.

```suggestion
      watch(preselectedFilters, emitPreselectedFilters, { immediate: true });
```

packages/x-components/src/x-modules/facets/components/preselected-filters.vue
```diff
@@ -1,69 +1,83 @@
...
+
+      return () => useNoElementRender(slots);
     }
```
CachedaCodes, line 79: Use the `render` function instead, it's more declarative.

```suggestion
    },
    render() {
      return useNoElementRender(this.$slots)
    }
```

CachedaCodes (changes requested):

packages/x-components/src/x-modules/facets/components/preselected-filters.vue
annacv: Hi, if I'm not wrong render is only available if we use the Options API, for the Composition API we should use `h` and return it directly, which I think is almost the same we are doing now.

CachedaCodes: The `defineComponent` function receives an options object that is more or less the same type of object you create in the options API (you are using it already, passing the props and setting up a name). `render` is one of the options that you can send to `defineComponent` and will work with `setup` when `setup` doesn't use `return`

CachedaCodes (changes requested):

CachedaCodes (approved):

PR #1423, opened 2024-02-26
fix: reset scroll on ExtraParamsChanged by default

pabletecodes:
https://jira.kroger.com/jira/browse/DCPPERS-12567

## Motivation and context

In the [Playboard](https://github.com/empathyco/playboard), we use the `ExtraParams` component (wrapped [here](https://github.com/empathyco/playboard/blob/main/packages/playboard-core/src/components/x/x-extra-params.component.vue#L2)). We often modify these `extraParams` programmatically.

We recently discovered a "bug" that the scroll of the `ResultList`(which we also use) wasn't getting reseted, causing and extra request to the search endpoint (for the second page or results) and the incorrect results displayed in the list 😢

The `ScrollMixin` component was already "listening" to the `UserChangedExtraParams` event (triggered by `RenderlessExtraParam`) and resetting the scroll of the scrollable element, but it wasn't "listening" to the `ExtraParamsChanged` event (triggered by `ExtraParams`).

This PR basically adds `ExtraParamsChanged` to the "default" list of events that should cause the scroll to be reseted.

## Type of change

- [x] `Main`
- [ ] Other. Specify:

## How has this been tested?

I tested this manually by adding the following to `packages/x-components/src/views/home/Home.vue`:

```
<template>
             ...
            <ExtraParams :values="extraParams" />
            <button @click="extraParams.a = extraParams.a + 1">Change extraParams</button>
            ...
</template>

<script>
    ...,
    components: {
      ...,
      ExtraParams
    },
    data() {
      return {
        extraParams: {
          a: 0
        }
      }
    }
<script>

```

**Without this fix (aka HOW TO REPRODUCE the issue):**
  - Search for "dress"
  - Scroll down a couple of "pages"
  - Click the "Change extraParams" buttons
  - Scroll of the list of results doesn't change and new results are fetched

**With this fix:**
  - Search for "dress"
  - Scroll down a couple of "pages"
  - Click the "Change extraParams" buttons
  - List of results is scrolled to the top and new results are fetched

## Checklist:

- [ ] I have added **tests** that prove my fix is effective or that my feature works

### DIFF
--- a/packages/x-components/src/components/scroll/scroll.mixin.ts
+++ b/packages/x-components/src/components/scroll/scroll.mixin.ts
@@ -63,6 +63,7 @@ export default class ScrollMixin extends Vue {
       'SelectedFiltersChanged',
       'SelectedFiltersForRequestChanged',
       'SelectedRelatedTagsChanged',
+      'ExtraParamsChanged',
       'UserChangedExtraParams'
     ]
   })


diegopf:
Thanks for submitting the PR @alonsogarciapablo! We will review it as soon as we can.

alonsogarciapablo:
We're finally solving the issue in `playboard-core` by using the `base-scroll` component and [passing the list of events (including `ExtraParamsChanged`) that should trigger the scroll reset via the `resetOn` prop](https://github.com/empathyco/playboard/pull/451/files#diff-2c0788a0873b862f65bf8866031ddae1f12080f5c5d6430e8663cc7fc00c662dR6-R14). Still, I think this PR makes sense  and we're planning to upgrade the version of `x-components` once this is merged 👍

annacv:
Hi @alonsogarciapablo,
Is it possible to emit the `UserChangedExtraParams` event with the new params as payload when you modify the `extraParams`?

alonsogarciapablo:
Hey @annacv, we're using the `ExtraParams` component ([here](https://github.com/empathyco/playboard/blob/5fa82eb8c106294ec4cc92244a20f200b4db5d8e/packages/playboard-core/src/components/x/x-extra-params.component.vue#L2)) which is triggering the `ExtraParamsChanged`.

Could we also trigger an `UserChangedExtraParams` event? Yes, that would be an option too. But then both events would be triggered... Also, perhaps this should happen more automagically so that anyone who uses `ExtraParams` can expect this behaviour?

annacv (changes requested):
Hi @alonsogarciapablo, 
The `ExtraParams` component is not intended for changing extra-params dynamically, it is just a “register” of all the params that will be used when the application is initialized. IMO, the `UserChangedExtraParams` event should not be attached by default to the component in the library, but for your use case, reacting to the `ExtraParams` 's computed prop [values](https://github.com/empathyco/playboard/blob/5fa82eb8c106294ec4cc92244a20f200b4db5d8e/packages/playboard-core/src/components/x/x-extra-params.component.vue#L2)'s change by emitting the `UserChangedExtraParams` event would be a good approach.

Anyway, we can argue about changing the `ScrollMixin` behavior and the events it is listening to. In this case, we can remove the `UserChangedExtraParams` event from the list and only listen to `ExtraParamsChanged`. At the end, whenever the`UserChangedExtraParams` event will be emitted, the `ExtraParamsXStoreModule` will emit `ExtraParamsChanged` reacting to the state change.

PR #1447, opened 2024-04-18
feat: make `useXBus` get bus from `XPlugin`

diegopf:
This PR refactors the use-xbus composable to first try to reuse the same from the XPlugin and fallback to the default x-bus implementation if the XPlugin has not been instantiated.

### DIFF
--- a/packages/x-components/src/components/modals/__tests__/base-id-modal-close.spec.ts
+++ b/packages/x-components/src/components/modals/__tests__/base-id-modal-close.spec.ts
@@ -2,8 +2,7 @@ import { mount, Wrapper } from '@vue/test-utils';
 import Vue from 'vue';
 import { getDataTestSelector, installNewXPlugin } from '../../../__tests__/utils';
 import BaseIdModalClose from '../base-id-modal-close.vue';
-import { bus } from '../../../plugins/index';
-import { dummyCreateEmitter } from '../../../__tests__/bus.dummy';
+import { XPlugin } from '../../../plugins/index';
 
 /**
  * Renders the {@link BaseIdModalClose} with the provided options.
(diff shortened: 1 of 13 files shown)


packages/x-components/src/components/modals/__tests__/base-id-modal-close.spec.ts
```diff
@@ -18,7 +18,7 @@ function renderBaseIdModalClose({
   // Making bus not repeat subjects
   jest.spyOn(bus, 'createEmitter' as any).mockImplementation(dummyCreateEmitter.bind(bus) as any);
 
-  const [, localVue] = installNewXPlugin();
+  const [, localVue] = installNewXPlugin(undefined, undefined, bus);
```
joseacabaneros, line 21: You can avoid it in every test sheet, changing every `bus.` use by `XPlugin.bus`. In this way, all test sheets will use the same bus, the dummy one. 

There are two types of "listeners" used:
```TS
XPlugin.bus.on('UserClickedCloseModal').subscribe(listener);
instead of
bus.on('UserClickedCloseModal').subscribe(listener);
```

```TS
emitSpy: jest.spyOn(XPlugin.bus, 'emit'),
instead of 
emitSpy: jest.spyOn(bus, 'emit'),
```

Moreover, if you do that change, you can also remove everything related with async events 🥳 
```TS
// Making bus not repeat subjects
  jest.spyOn(bus, 'createEmitter' as any).mockImplementation(dummyCreateEmitter.bind(bus) as any);

jest.runAllTimers();

beforeAll(() => {
  jest.useFakeTimers();
});
```

packages/x-components/src/composables/use-x-bus.ts
```diff
@@ -23,7 +24,12 @@ export function useXBus(): UseXBusAPI {
   if (currentComponent && currentXComponent) {
     currentComponent.xComponent = currentXComponent;
   }
-
+  let bus: XBus<XEventsTypes, WireMetadata>;
+  try {
```
joseacabaneros, line 28: Why is it a `try/catch` and not an `undefined` checking? 🤔

diegopf: The `XPlugin` getter for the bus relies on the [getInstance](https://github.com/empathyco/x/blob/main/packages/x-components/src/plugins/x-plugin.ts#L75) static method that throws if the plugin was not instantiated.

joseacabaneros: Oouh! Makes sense. Could we add a small comment here with that reason? Or with a `@remarks` tag in the composable doc

joseacabaneros (approved):

joseacabaneros (approved):

PR #1456, opened 2024-04-26
feat: migrate `MainScroll` and `MainScrollItem` components

diegopf:
This PR would be a BREAKING CHANGE since we drop the `NoElement` usage for the `MainScroll` and we use an extra div.


packages/_vue3-migration-test/src/components/scroll/test-scroll.vue
diegopf, line 1: Once we have the store ready in this package, we could test the restore scroll position from url. I tested it locally hardcoding the value and seems to work just fine.

packages/x-components/src/x-modules/scroll/components/main-scroll.vue
```diff
@@ -26,234 +23,260 @@
...
-     * @returns The first visible element in this component.
-     * @internal
-     */
-    @XEmit('UserScrolledToElement')
-    public get firstVisibleElement(): string | '' {
-      if (this.intersectingElements.length === 0) {
-        return '';
+  export default defineComponent({
+    name: 'MainScroll',
+    props: {
```
joseacabaneros, line 28: Pending xModule name for the wiring; `xModule: scrollXModule.name`

packages/x-components/tests/unit/main-scroll.spec.ts
```diff
@@ -24,8 +24,9 @@ function renderMainScroll({
   useWindow,
   windowScrollingElement,
   template = `
-      <MainScroll v-bind="{ threshold, margin, useWindow }">
-        <div class="container" ${!useWindow ? `data-test="scroll"` : ''}>
+      <MainScroll v-bind="{ threshold, margin, useWindow }"
+                  class="root-scroll">
+        <div id="rootScroll" class="container" ${!useWindow ? `data-test="scroll"` : ''}>
```
joseacabaneros, line 29: I think now you can avoid this `id`

packages/x-components/src/x-modules/scroll/components/main-scroll.vue
```diff
@@ -26,234 +23,257 @@
...
-      this.intersectionObserver = new IntersectionObserver(this.updateVisibleElements, {
-        root: this.useWindow ? document : this.$el,
-        threshold: this.threshold,
-        rootMargin: this.margin
+      /**
+       * Initialise the observer after mounting the component.
+       */
+      onMounted(() => {
+        intersectionObserver.value = new IntersectionObserver(updateVisibleElements, {
+          root: props.useWindow ? document : rootRef.value,
```
joseacabaneros, line 169: 🔝

packages/x-components/src/x-modules/scroll/components/main-scroll.vue
```diff
@@ -1,19 +1,16 @@
 <template>
-  <NoElement :class="dynamicClasses">
+  <div ref="rootRef" :class="dynamicClasses">
```
joseacabaneros, line 2: Styling this `div` in the component itself could be an option? In this way, we could avoid breaking changes in this PR

packages/x-components/src/x-modules/scroll/components/main-scroll-item.vue
```diff
@@ -19,95 +31,134 @@
...
+      /**
+       * The item data. Used to set the scroll identifier.
+       *
+       * @public
+       */
+      item: {
+        type: Object as PropType<Identifiable>,
+        required: true
+      },
+
```
joseacabaneros, line 47: Format this SFC

packages/x-components/src/x-modules/scroll/components/main-scroll-item.vue
```diff
@@ -19,95 +31,134 @@
...
+      /**
+       * The tag to render.
+       *
+       * @public
+       */
+      tag: {
+        type: [String, Object] as PropType<string | typeof Vue>,
+        default: () => NoElement
```
joseacabaneros, line 55: If `NoElement` as root node, is able to retrieve the `el`? Because seems the same behaviour as `MainScroll` 🤔

packages/_vue3-migration-test/src/components/scroll/test-scroll.vue
```diff
@@ -0,0 +1,23 @@
+<script setup lang="ts">
+  import MainScroll from '../../../../x-components/src/x-modules/scroll/components/main-scroll.vue';
+  // eslint-disable-next-line max-len
+  import MainScrollItem from '../../../../x-components/src/x-modules/scroll/components/main-scroll-item.vue';
+  const items = Array.from({ length: 24 }, (_, index) => ({ id: `item-${index}` }));
+</script>
+
+<template>
+  <MainScroll>
+    <ul id="rootScroll" class="list" data-test="scroll">
```
joseacabaneros, line 10: `id="rootScroll"` is not necessary, right?

joseacabaneros (commented):
Looks good!

packages/_vue3-migration-test/src/components/scroll/test-scroll.vue
diegopf: Not anymore, not really.

packages/x-components/src/components/animations/disable-animation.mixin.ts
```diff
@@ -22,7 +21,7 @@ export default class DisableAnimationMixin extends Vue {
    *
    * @public
    */
-  @XInject(DISABLE_ANIMATIONS_KEY)
+  @Inject({ from: DISABLE_ANIMATIONS_KEY as string, default: false })
```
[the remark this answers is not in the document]
diegopf: I did not think about it since we will remove the mixin anyway in the near future. However you are right!

packages/x-components/src/x-modules/scroll/components/main-scroll-item.vue
diegopf: mmm I'm running prettier locally but nothing change. What's missing? 🤔

packages/x-components/src/x-modules/scroll/components/main-scroll-item.vue
diegopf: Good catch! I will check but probably you are right

packages/x-components/src/x-modules/scroll/components/main-scroll.vue
diegopf: I think that highly depends on the layout implementation from the outside. Not sure if we could just style it here

packages/x-components/src/components/animations/disable-animation.mixin.ts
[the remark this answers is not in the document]
diegopf: I've tried but the tests are failing. I think the inject is not working well along with the vue class component. IMO we leave the decorator for now and handle this when we get rid of this mixin.

packages/x-components/src/x-modules/scroll/components/main-scroll-item.vue
diegopf: Indeed, same problem. I followed same approach here and default to a div.

packages/x-components/src/x-modules/scroll/components/main-scroll-item.vue
joseacabaneros: The line breaks between `defineComponent` options

joseacabaneros (approved):

packages/_vue3-migration-test/src/components/index.ts
```diff
@@ -2,4 +2,5 @@ export * from './animations';
 export { default as TestBaseColumnPickerDropdown } from './column-picker/test-base-column-picker-dropdown.vue';
 export { default as TestBaseColumnPickerList } from './column-picker/test-base-column-picker-list.vue';
 export { default as TestBaseDropdown } from './test-base-dropdown.vue';
+export * from '../x-modules/scroll';
```
joseacabaneros, line 5: I think you have to export it in `src/x-modules/index.ts` instead of here

packages/_vue3-migration-test/src/x-modules/scroll/test-scroll.vue
```diff
@@ -0,0 +1,23 @@
+<script setup lang="ts">
+  import MainScroll from '../../../../x-components/src/x-modules/scroll/components/main-scroll.vue';
+  // eslint-disable-next-line max-len
```
joseacabaneros, line 3: You can remove it beacuse it is disabled

packages/_vue3-migration-test/src/x-modules/scroll/x-module.ts
```diff
@@ -0,0 +1,7 @@
+export const scrollXModule = {
```
joseacabaneros, line 1: Is it possible to type it?

joseacabaneros (approved):

PR #1495, opened 2024-05-30
Make base-grid return an empty array by default

annacv:
...to avoid missing list injection Key warn 
Captura de pantalla 2024-05-30 a les 13 16 36

### DIFF
--- a/packages/x-components/src/components/base-grid.vue
+++ b/packages/x-components/src/components/base-grid.vue
@@ -114,7 +114,7 @@
        *
        * @internal
        */
-      const injectedListItems = inject<Ref<ListItem[] | []>>(LIST_ITEMS_KEY as string, []);
+      const injectedListItems = inject<Ref<ListItem[] | []>>(LIST_ITEMS_KEY as string, ref([]));
 
       const gridEl = ref<ElementRef | HTMLElement | null>(null);
 
--- a/packages/x-components/src/components/base-grid.vue
+++ b/packages/x-components/src/components/base-grid.vue
@@ -140,8 +140,14 @@
        *
        * @public
        */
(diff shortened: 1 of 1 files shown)


joseacabaneros (dismissed):

packages/x-components/src/components/base-grid.vue
```diff
@@ -114,7 +114,7 @@
        *
        * @internal
        */
-      const injectedListItems = inject<Ref<ListItem[]>>(LIST_ITEMS_KEY as string);
+      const injectedListItems = inject<Ref<ListItem[]> | []>(LIST_ITEMS_KEY as string, []);
```
joseacabaneros, line 117: Array type inside `Ref`: `inject<Ref<ListItem[] | []>>`

annacv: Hi, wasn't sure about it as ID warned: 
Captura de pantalla 2024-06-05 a les 18 28 07

annacv: ☝️ Set ref([]) as default value made the job, but , what do we want to do with the console.warn? Keep it or shall I change tests?? 🙏 @joseacabaneros

joseacabaneros: I'm checking how we implement this same case in other injections, and it is common to use `undefined` as default value for the `inject`. 
With the aim of align the code base, could you change it by `undefined`? 
It is all about alignment, yours is also perfectly valid  ✌🏾  

```TS
const injectedListItems = inject<Ref<ListItem[]> | undefined>(
  LIST_ITEMS_KEY as string,
  undefined
);
```

joseacabaneros (approved):

PR #1552, opened 2024-07-03
fix(main-scroll): obtain observed container element reliably to work with any Vue version

joseacabaneros:
Obtain observed container element reliably to work with any Vue version

With the following code, we are able to retrieve the root element of a renderless component in both Vue 2 and Vue 3:

```TS
onMounted(() => {
   containerRef.value = getCurrentInstance()?.proxy?.$el;
});
...
return () => slots.default?.()[0];
```

In Vue 2 and Vue 3:
- `slots.default?.()` -> Gives an array of `vNodes` | `[0]` Gets the `vNode` root and renders it 

IF we would return the full array of `vNodes` `return () => slots.default?.()`:
- In Vue 2: Renders the `vNodes` as it is without problem, and we can get the root element with `getCurrentInstance()?.proxy?.$el`. Only if the default slot has a single node, otherwise, Vue will throw an error 
`[Vue warn]: Multiple root nodes returned from render function. Render function should return a single root node.`
- In Vue 3: Renders the `vNodes` as a `Fragment` (because in Vue3 supports multiple root nodes) and when we do `getCurrentInstance()?.proxy?.$el` it gives a `#text` node because the root node is the `Fragment`.

One more thing, this scope will support `attrs` inheritance in both Vue versions ✌🏾

### DIFF
--- a/packages/x-components/src/x-modules/scroll/components/main-scroll-item.vue
+++ b/packages/x-components/src/x-modules/scroll/components/main-scroll-item.vue
@@ -107,6 +107,10 @@
         {
           if (rootRef.value !== null) {
             const htmlElement = isElementRef(rootRef.value) ? rootRef.value.$el : rootRef.value;
+            // eslint-disable-next-line no-console
+            console.log(isElementRef(rootRef.value));
+            // eslint-disable-next-line no-console
+            console.log(rootRef.value);
 
             oldObserver?.unobserve(htmlElement);
             newObserver?.observe(htmlElement);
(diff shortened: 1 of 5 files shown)


packages/x-components/src/x-modules/scroll/components/main-scroll-item.vue
```diff
@@ -105,14 +73,12 @@
         oldObserver: ScrollVisibilityObserver | null
       ): void => {
         {
-          if (rootRef.value !== null) {
-            const htmlElement = isElementRef(rootRef.value) ? rootRef.value.$el : rootRef.value;
```
joseacabaneros, line 109: If someone can say/demonstrate me why is it necessary, recojo cable

diegopf: I'm pretty sure that we had an issue that led to this code but I can't remember the specifics. I've trying to replicate it and failed. Maybe it was a side effect due to other thing that we already changed, I don't know 🤷‍♂️ Anyway, seems that this is working just fine now

diegopf (approved):

PR #1553, opened 2024-07-04
feat(create-animation): migrate to composition API

victorcg88:
EMP-4442

⚠️ It doesn't work on Vue3. We would need to rework this after upgrading the Vue version.

## Type of change

- [ ] `Main`
- [ ] Other. Specify:

## How has this been tested?

## Checklist:


packages/x-components/src/components/animations/create-directional-animation-factory.ts
```diff
@@ -1,4 +1,14 @@
-import Vue, { VueConstructor } from 'vue';
+import { defineComponent, h } from 'vue';
+
+type AnimationOrigin =
```
diegopf, line 3: I think we should keep exporting the type
```suggestion
export type AnimationOrigin =
```

joseacabaneros: For any specific case? If not, I think it could be internal because when you use the returned function, the parameter will be type implicitly `animateTranslate('bottom');`. If we don't need to use outside explicitly, we could keep it internal

diegopf: I had issues building the project locally without exporting the type. However might be fine but then we need to label it as BREAKING-CHANGE

joseacabaneros: You are right! It will be a BC. Consider it exported

diegopf (approved):

PR #1555, opened 2024-07-05
feat(display-emitter): support Vue3 to the component

joseacabaneros:
Support Vue3 to the `DisplayEmitter`component becuase the same issue we already saw with the `MainScroll` https://github.com/empathyco/x/pull/1552.

A template ref over a slot (AKA `NoElement`) in vue3 renders a `Fragment` and the ref is lost, but in vue2 renders the first `vNode` of the slot and assign the ref in a right way.


packages/x-components/src/components/display-emitter.vue
```diff
@@ -1,55 +1,58 @@
...
-  /**
-   * A component that emits a display event when it first appears in the viewport.
-   */
+  /** A component that emits a display event when it first appears in the viewport. */
   export default defineComponent({
-    components: {
-      NoElement
-    },
+    name: 'DisplayEmitter',
+    xModule: taggingXModule.name,
```
joseacabaneros, line 19: Should this component be relocated to the `tagging` x-module? 🤔

diegopf: mmm but we don't really need to register the tagging module, do we?

joseacabaneros: As we already decided, meanwhile this component is generic, no sense to register any x-module inside

packages/x-components/src/components/display-emitter.vue
```diff
@@ -1,55 +1,56 @@
...
+            element,
+            taggingRequest: props.payload,
+            ...(props.eventMetadata && { eventMetadata: props.eventMetadata })
+          }).unwatchDisplay;
+        }
       });
 
-      onUnmounted(unwatchDisplay);
+      onUnmounted(() => {
+        unwatchDisplay();
```
diegopf, line 44:
```suggestion
        unwatchDisplay && unwatchDisplay();
```
I guess this will not happen but there's a chance that unwatchDisplay is undefined now, so I think we should just check for it before invoking it

victorcg88 (dismissed):
Code LGTM, tested locally and it works just fine.

packages/x-components/src/components/display-emitter.vue
joseacabaneros: Done

diegopf (approved):

PR #1563, opened 2024-07-10
feat: migrate internal view components

diegopf:


packages/x-components/src/views/accessibility/wai-base-result-link.vue
lauramargar, line 1: Maybe you have forgotten to define the component name (`name: 'AccessibilityCheck'`)

packages/x-components/src/views/home/result.vue
```diff
@@ -35,29 +35,34 @@
...
+    name: 'Result',
     components: {
       BaseResultAddToCart,
       BaseResultLink,
       BaseResultRating,
       BaseResultImage
+    },
+    props: {
+      result: {
+        type: Object as () => Result,
```
lauramargar, line 56: This prop type definition is ok, but you can also define the type with 
`Object as PropType<Result>`

packages/x-components/src/views/home/sliding-next-query-preview.vue
```diff
@@ -23,26 +23,26 @@
...
+    name: 'SlidingNextQueryPreview',
     components: {
       NextQueryPreview,
       NextQuery,
       SlidingPanel,
       Result
+    },
+    props: {
+      suggestion: {
+        type: Object as () => NextQueryModel,
```
lauramargar, line 43: Here the same, you can also use 
`Object as PropType<NextQueryModel>`

lauramargar (approved):
gj!!

PR #1564, opened 2024-07-11
refactor(no-element): towards getting rid of no-element component

joseacabaneros:
`NoElement` component is gonna be an issue in terms of attribute inheritance on Vue3 because `Fragment` creation to support multiple root nodes. To solve it, there is a workaround we can achieve. This is to manage the `vNodes` creation in the render function in the components themselves (moreover, without generating breaking changes). 
To do it, we must refactor them in two steps:
- Components which will work in both Vue version because they don't have fallback content in the slots (this PR)
- Components which won't work in both Vue versions because they require additional `props` in the `h` function, and they are breaking-change in the Vue migration (future PR once we have Vue3 officially)

Components refactored in this PR to get rid of using `NoElement` component:
- `DisplayResultProvider`
- `RenderlessExtraParam`
- `SelectedFilters`

Extra: `ResultVariantsProvider` component was also refactored to avoid `$scopedSlots` which is breaking-change in Vue3.

### DIFF
--- a/packages/x-components/src/x-modules/facets/components/lists/selected-filters.vue
+++ b/packages/x-components/src/x-modules/facets/components/lists/selected-filters.vue
@@ -1,12 +1,6 @@
-<template>
-  <NoElement v-if="isVisible" class="x-selected-filters">
-    <slot v-bind="{ selectedFilters }">{{ selectedFilters.length }}</slot>
-  </NoElement>
-</template>
-
 <script lang="ts">
   import { Facet } from '@empathyco/x-types';
-  import { defineComponent, PropType } from 'vue';
+  import { defineComponent, h, PropType } from 'vue';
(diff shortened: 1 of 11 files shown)


diegopf (approved):

packages/x-components/src/x-modules/facets/components/lists/selected-filters.vue
diegopf, line 1: We can remove the NoElement usage from the components and the import, right?

diegopf (changes requested):

packages/x-components/src/x-modules/facets/components/lists/selected-filters.vue
joseacabaneros: Yes! I forgot it. Removed

diegopf (approved):

PR #1576, opened 2024-07-25
fix(list-components): fix list components, migrate infinite-scroll directive and deprecated layouts

joseacabaneros:
3 things done here:
1. Getting rid of `NoElement` usage in `-list` components (`ResultsList`, `PromotedsList`, `BannersList` and `NextQueriesList`). It fixes warnings regarding attrs inheritance to the root node.
```HTML
<template>
  <NoElement>
    <!--
      @slot Customize ResultsList.
        @binding {Result[]} items - Results to render.
        @binding {Vue | string} animation - Animation to animate the elements.
    -->
    <slot v-bind="{ items, animation }">
      <ItemsList :animation="animation" :items="items">
        <template v-for="(_, slotName) in slots" v-slot:[slotName]="{ item }">
          <slot :name="slotName" :item="item" />
        </template>
      </ItemsList>
    </slot>
  </NoElement>
</template>
```
is the same as the following render function
```TS
return () => {
  const innerProps = { items: items.value, animation: props.animation };
  return slots.default?.(innerProps)[0] ?? h(ItemsList, innerProps);
};
```
2. Removing `UseLayouts` composable and refactor all deprecated layouts implied
- Revert `Home` to use again `MultiColumnMaxWidthLayout`
3. Migrate `InfiniteScroll` directive and make it simpler. Now, it doesn't depend on the component instance to emit `UserReachedResultsListEnd` event so `onInfiniteScrollEnd` function in `ResultsList` is no longer necessary.


diegopf (approved):

packages/x-components/src/x-modules/next-queries/components/next-queries-list.vue
```diff
@@ -104,30 +59,24 @@
...
       const nextQueries: ComputedRef<NextQuery[]> = useGetter('nextQueries', [
         'nextQueries'
       ]).nextQueries;
 
-      /**
-       * Injected query, updated when the related request(s) have succeeded.
-       */
-      const injectedQuery = inject<Ref<string | undefined>>(QUERY_KEY as string);
+      /** Injected query, updated when the related request(s) have succeeded. */
+      const injectedQuery = inject<Ref<string> | undefined>(QUERY_KEY as string, undefined);
```
victorcg88, line 68:
```suggestion
      const injectedQuery = inject<Ref<string> | undefined>(QUERY_KEY as string);
```
Is there a reason to explicitly set `undefined` as the default? Vue already does that:

> If no value with matching key was found, inject() returns undefined unless a default value is provided.

https://vuejs.org/api/composition-api-dependency-injection.html#inject

victorcg88 (approved):

packages/x-components/src/x-modules/next-queries/components/next-queries-list.vue
joseacabaneros: It was because at some point in the time, I saw a warning saying "you are injecting a key that wasn't provided". But I am not able to see it again. 
As you say, Vue does it by us, I restored the injected default value for the time being and also removed the `undefined` from the generic type because when you don't give a default value for the injection, `undefined` is already inferred `export declare function inject<T>(key: InjectionKey<T> | string): T | undefined;`

victorcg88 (approved):

PR #1582, opened 2024-08-01
test(auto-progress-bar): fix test after Vue3 update

victorcg88:
## Motivation and context

Fix `auto-progress-bar` unit test after Vue3 upgrade.

It has been necessary to bump `rollup-plugin-vue`, given that the previous version had as dependency the `vue-template-compiler` for Vue2 which caused incompatibilities running the test.

## Type of change

- [ ] `Main`
- [X] Other. Specify: https://github.com/empathyco/x/tree/vue3-update-rc

## How has this been tested?

## Checklist:

### DIFF
--- a/packages/x-components/src/__tests__/utils.ts
+++ b/packages/x-components/src/__tests__/utils.ts
@@ -1,7 +1,6 @@
 import { deepMerge } from '@empathyco/x-deep-merge';
 import { DeepPartial, Dictionary } from '@empathyco/x-utils';
-import { createLocalVue, Wrapper, WrapperArray } from '@vue/test-utils';
-import Vue from 'vue';
+import { VueWrapper } from '@vue/test-utils';
 import { Store } from 'vuex';
 import {
   XComponentsAdapter,
@@ -69,7 +68,7 @@ export function getDataTestSelector(dataTest: string): string {
  *
  * @returns The wrappers matching the searched test data id.
  */
-export function findTestDataById(wrapper: Wrapper<Vue>, testDataId: string): WrapperArray<Vue> {
+export function findTestDataById(wrapper: VueWrapper, testDataId: string) {
   return wrapper.findAll(getDataTestSelector(testDataId));
 }
 
@@ -194,16 +193,16 @@ function mergeStates<State extends Dictionary>(
  */
 export function installNewXPlugin(
   options: Partial<XPluginOptions> = {},
-  localVue: typeof Vue = createLocalVue(),
+  localVue = undefined,
   bus = new XDummyBus()
-): [XPlugin, typeof Vue] {
+): [XPlugin, undefined] {
   XPlugin.resetInstance();
   const xPlugin = new XPlugin(bus);
-  const installOptions: XPluginOptions = {
-    adapter: XComponentsAdapterDummy,
-    ...options
-  };
-  localVue.use(xPlugin, installOptions);
+  // const installOptions: XPluginOptions = {
+  //   adapter: XComponentsAdapterDummy,
+  //   ...options
+  // };
+  // localVue.use(xPlugin, installOptions);
   return [xPlugin, localVue];
 }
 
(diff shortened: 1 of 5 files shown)


packages/x-components/src/plugins/x-emitters.ts
```diff
@@ -57,14 +58,15 @@ export function registerStoreEmitters(
  */
 function normalizeStateSelector(
   stateSelector: AnySimpleStateSelector | AnyStateSelector
-): Required<AnyStateSelector> {
+): Required<Omit<AnyStateSelector, keyof WatchOptionsBase>> {
```
victorcg88, line 61: StateSelector extends WatchOptions which in Vue3 adds once prop and also extends WatchOptionsBase. To overcome this type error, I'm omitting the WatchOptionsBase props from being required.

joseacabaneros: I am not sure if even removing the returning type, we would do less type-overengineering... 
Perhaps making also a constant with the default options could help to understand.
```TS
const defaultOptions: Omit<AnyStateSelector, 'selector'> = {
  deep: false,
  immediate: false,
  once: false,
  filter: () => true,
  metadata: { replaceable: true }
};
```
I haven't got a strong opinion here. Whatever you decide, I approve ✌🏾

packages/x-components/src/components/__tests__/auto-progress-bar.spec.ts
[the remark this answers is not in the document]
victorcg88: Of course, I like it. My wondering here is, whether should we go this far refactoring tests or just go for the minimum required changes to make them pass.

packages/x-components/src/__tests__/utils.ts
```diff
@@ -194,16 +193,16 @@ function mergeStates<State extends Dictionary>(
  */
 export function installNewXPlugin(
```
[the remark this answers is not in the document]
victorcg88: I'm refactoring this in another PR which I will open soon. We could discuss it then and there but I would like to keep the function because we would need to use quite a lot in the tests, so we can save several lines of code. 

My proposal will be:

image

And the default use would be to use just a line:

`global: { plugins: [installNewXPlugin()] },`

packages/x-components/src/plugins/x-emitters.ts
victorcg88: I also have mixed feelings about this. But I'm happy with my proposal because in the end is simple. Don't know if  @diegopf has a different opinion about this.

packages/x-components/src/__tests__/utils.ts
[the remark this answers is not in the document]
victorcg88: PR adopting these changes: #1586

packages/x-components/src/components/__tests__/auto-progress-bar.spec.ts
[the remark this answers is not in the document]
joseacabaneros: My opinion is "minimum changes" but is you have to tweak a test to change a type (like in this PR), I would try to make a minimum refactor. The most of the cases is only removing types and making the render function simpler.

joseacabaneros (approved):

packages/x-components/src/plugins/x-emitters.ts
```diff
@@ -6,6 +6,7 @@ import { AnySimpleStateSelector, AnyStateSelector } from '../store/utils/store-e
 import { XEventPayload, XEventsTypes } from '../wiring/events.types';
 import { WireMetadata } from '../wiring/wiring.types';
 import { AnyXModule } from '../x-modules/x-modules.types';
+import { WatchOptionsBase } from 'vue';
```
diegopf, line 9: estlint complains about the import order, we should move it to above the relative ones

packages/x-components/src/plugins/x-emitters.ts
diegopf: Both approaches seem fine to me. I kind of prefer Jose's approach as it appears easier.

diegopf (approved):

PR #1603, opened 2024-08-22
Rollup production build in Vue3

joseacabaneros:
- Migrate `scss` to `css`, `scoped` when only necessary, and clean up. CSS nesting and `scoped` doesn't generate a right CSS code. Flatten class selectors and use CSS to avoid preprocessors
- Use `@vitejs/plugin-vue` instead of deprecated `rollup-plugin-vue`
- Use a simple custom CSS injector `build/tools/inject-css.js` in `vue-plugin-styles`
- Simplify the `build.ts` content
- Make `apiDocumentation` plugin work in the build process 🥳


joseacabaneros:
There is an issue with no-scoped styles in `layout`. They are no-scoped and have shared class names, so styles of all layouts are applied to the same. It can be replicated in `serve` with Vite. 
WIP in a workaround for layouts.

diegopf (approved):

PR #1604, opened 2024-08-28
feat: bump vue18n version

diegopf:
Updates vue related deps for `@empathyco/x-archetype-utils`. This package contains the custom i18n plugin used in x-archetype. 

Update vue18n to 9.14.0 and make the necessary adjustments.

BREAKING CHANGE: This package now is only compatible with vue 3

### DIFF
--- a/packages/x-archetype-utils/src/__tests__/i18n.plugin.spec.ts
+++ b/packages/x-archetype-utils/src/__tests__/i18n.plugin.spec.ts
@@ -1,6 +1,5 @@
 import { DeepPartial } from '@empathyco/x-utils';
-import Vue from 'vue';
-import { mount, createLocalVue, Wrapper } from '@vue/test-utils';
+import { mount, VueWrapper } from '@vue/test-utils';
 import { I18n } from '../i18n/i18n.plugin';
 import { Device, I18nOptions, Locale } from '../i18n/i18n.types';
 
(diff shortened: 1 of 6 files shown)


packages/x-archetype-utils/src/i18n/i18n.plugin.ts
```diff
@@ -49,19 +49,20 @@ export class I18n<SomeMessages> {
...
-    vue.use(VueI18n);
-    this.vueI18n = new VueI18n({
+  install(vue: App): void {
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+    this.vueI18n = createI18n({
       locale: this.locale,
-      messages: this.currentMessages ? { [this.locale]: this.currentMessages } : {},
       silentFallbackWarn: true,
-      missing: (locale, key) => {
+      missing: (locale, key: string) => {
```
victorcg88, line 57:
```suggestion
      missing: (_, key: string) => {
```

packages/x-archetype-utils/src/i18n/i18n.plugin.ts
```diff
@@ -76,7 +77,7 @@ export class I18n<SomeMessages> {
     // TO-DO Support function or array messages
     const message = path
       .split('.')
-      .reduce<LocaleMessage>(
+      .reduce<any>(
         (messages, key, index, pathParts) =>
           isObject(messages)
```
victorcg88, line 82: It's just me or are we getting an `Unsafe call of an `any` typed value.` error here?

packages/x-archetype-utils/src/__tests__/i18n.plugin.spec.ts
```diff
@@ -1,6 +1,5 @@
 import { DeepPartial } from '@empathyco/x-utils';
-import Vue from 'vue';
-import { mount, createLocalVue, Wrapper } from '@vue/test-utils';
+import { mount, VueWrapper } from '@vue/test-utils';
```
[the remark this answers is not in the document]
diegopf: Good catch! This is not correct and looks like it is already an issue the current published versions have. I just pushed a commit to fix it.

victorcg88 (approved):

joseacabaneros (approved):

PR #1619, opened 2024-09-20
test(e2e-components): rewrite all e2e sheet components to accomplish vue3 

joseacabaneros:
Rewrite all cypress component tests to accomplish vue3 requirements.
Passing CI ✅

### DIFF
--- a/packages/x-components/cypress.config.ts
+++ b/packages/x-components/cypress.config.ts
@@ -1,5 +1,3 @@
-import { addCucumberPreprocessorPlugin } from '@badeball/cypress-cucumber-preprocessor';
-import { createEsbuildPlugin } from '@badeball/cypress-cucumber-preprocessor/esbuild';
 import { defineConfig } from 'cypress';
 
 export default defineConfig({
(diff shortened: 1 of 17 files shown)


packages/x-components/cypress.config.ts
```diff
@@ -9,34 +7,41 @@ export default defineConfig({
...
-      await addCucumberPreprocessorPlugin(on, config);
-      on(
-        'file:preprocessor',
-        require('@bahmutov/cypress-esbuild-preprocessor')({
-          plugins: [createEsbuildPlugin(config)]
-        })
+      const { createEsbuildPlugin } = require('@badeball/cypress-cucumber-preprocessor/esbuild');
+      const createBundler = require('@bahmutov/cypress-esbuild-preprocessor');
+
+      // await here
```
victorcg88, line 25: Maybe we should improve this comment.

packages/x-components/tests/unit/banner.spec.ts
```diff
@@ -60,13 +33,8 @@ describe('testing Banner component', () => {
...
 
   it('banner is not rendered if image load fails', () => {
-    const { getBanner, getBannerImage } = mountBanner({
-      banner: createBannerStub('banner', {
-        title: 'Search UIs',
-        url: 'https://empathy.co',
-        position: 1
-      })
-    });
+    const { getBanner, getBannerImage } = render(createBannerStub('banner', bannerBase));
```
victorcg88, line 36:
```suggestion
    const { getBanner, getBannerImage } = render();
```

We could make `createBannerStub('banner', bannerBase)` the default value for the `banner` arg of `render`. WDYT? I know it is not a big deal here, but in any case, I think it is a good pattern to follow.``

victorcg88 (commented):
Mega job! 🔝 

image

packages/x-components/cypress.config.ts
joseacabaneros: Yep. I rewrote this config following the official doc for cucumber

packages/x-components/tests/unit/banner.spec.ts
joseacabaneros: I like it! Reformatted for default param value

victorcg88 (approved):

PR #1628, opened 2024-09-30
fix: Snippet config extra params reactivity

annacv:
# Pull request template
 
This PR checks if the extra params object has really changed to avoid unnecessary event emissions. 

## Motivation and context

In some customers, when a snippetConfig callback is executed, `ExtraParamsProvided` and `ExtraParamsChanged` events are fired even if they have not changed (and, as a consequence all the rest of the chained events that react to ExtraParamsChanged event emission). 

- [x] Open issue. If applicable, link: [RST-2269](https://searchbroker.atlassian.net/browse/RST-2269)

## Type of change

- [x] `Main`
- [ ] Other. Specify:

## How has this been tested?

An [open PR](https://github.com/empathyco/x-primor/pull/143) can be used to see the implementation when a callback is executed & check: 
* No events are emitted if the `extraParams` object has not changed
* No unnecessary endpoint calls (regarding chained events).

## Checklist:

[RST-2269]: https://searchbroker.atlassian.net/browse/RST-2269?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/x-components/src/x-modules/extra-params/components/snippet-config-extra-params.vue
+++ b/packages/x-components/src/x-modules/extra-params/components/snippet-config-extra-params.vue
@@ -35,25 +35,16 @@
     },
     setup(props) {
       const snippetConfig = inject('snippetConfig') as SnippetConfig;
-      const extraParams = ref<Record<string, unknown>>({});
-      const initialExtraParams = ref<Record<string, unknown>>({});
+      const extraParams = ref<Record<string, any>>({});
 
       watch(
         [() => snippetConfig, () => props.values],
         () => {
-          const newExtraParams: Record<string, unknown> = {};
-
           forEach({ ...props.values, ...snippetConfig }, (name, value) => {
-            if (!props.excludedExtraParams.includes(name)) {
-              newExtraParams[name] = value;
+            if (!props.excludedExtraParams.includes(name) && extraParams.value[name] !== value) {
+              extraParams.value = { ...extraParams.value, [name]: value };
             }
           });
-
-          // Check if extraParams are really changing to avoid event emission if they don't
-          if (JSON.stringify(newExtraParams) !== JSON.stringify(initialExtraParams.value)) {
-            extraParams.value = newExtraParams;
-            initialExtraParams.value = newExtraParams;
-          }
         },
         {
           deep: true,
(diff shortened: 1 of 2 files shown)


packages/x-components/src/x-modules/extra-params/components/snippet-config-extra-params.vue
```diff
@@ -35,16 +35,25 @@
...
 
       watch(
         [() => snippetConfig, () => props.values],
         () => {
+          const newExtraParams: Record<string, unknown> = {};
+
           forEach({ ...props.values, ...snippetConfig }, (name, value) => {
             if (!props.excludedExtraParams.includes(name)) {
-              extraParams.value = { ...extraParams.value, [name]: value };
+              newExtraParams[name] = value;
```
diegopf, line 48: I think we can do this directly with extraParams ref.

`extraParams.value[name] = value;`

This way we could skip the comparison that follows to check if something have changed. Assigning the same value should not trigger the reactivity in Vue.

diegopf (changes requested):

packages/x-components/src/x-modules/extra-params/components/snippet-config-extra-params.vue
joseacabaneros: In that case, `ExtraParams` component won't trigger the reactivity of the prop because it requires an update of the whole object. For example, triggering `InterfaceX.setSnippetConfig({ lang: 'es' })`, changes the `extraParams` value `extraParams.value[name] = value;` but the watcher inside `ExtraParams` component is not triggered. 

Trying to look for a clear solution instead of saving a copy and comparing with, I did that and tested in the X demo:
```TS
const extraParams = ref<Dictionary<unknown>>({});

watch(
  [() => snippetConfig, () => props.values],
  () => {
    const newExtraParams = { ...snippetConfig, ...props.values };
    const shouldChange = Object.entries(newExtraParams).some(
      ([name, value]) =>
        !props.excludedExtraParams.includes(name) && extraParams.value?.[name] !== value
    );
    if (shouldChange) {
      extraParams.value = newExtraParams;
    }
  },
  {
    deep: true,
    immediate: true
  }
);
```

Perhaps I missed something. Tell me WDYT

diegopf (approved):

PR #1629, opened 2024-10-01
fix(scroll): scroll direction inconsistency

victorcg88:
# Pull request template
 
After the Vue3 upgrade the `useScroll` composable created a inconsistency in the scroll direction prop in the `useScroll` instance and the scroll direction prop in the Vuex store.

## Explanation:

When scrolling, the data in `useScroll` is updated via the `storeScrollData` method, which updates values on `currentPosition` and `clientHeight`. For the first scroll event, the behavior is described below:

In **Vue2** the running order was:

1. Callback of `clientHeight` watch, which blocks updates on scroll direction because `isClientHeightChanging` is set to `true`.
2. Callback of `currentPosition` watch, skipped due to `isClientHeightChanging` = true.

But in **Vue3** the running order is:

1. Callback of `currentPosition` watch, because in the running order of `storeScrollData` currentPosition is mutated first. Then `isClientHeightChanging` is false, so it mutates `scrollDirection`
2. Callback of `clientHeight` watch, which blocks updates on scroll direction because `isClientHeightChanging` is set to `true`.
3. Callback of `scrollDirection` watch, skipped due to `isClientHeightChanging` = true. So the event is not emitted and the store keeps the original value not the current one.

## Consequences:

In the x-archetype, it is used the `useHasScrollPastThreshold` composable which relies on the scroll direction of the store. It is used in this case to collapse the subheader of the desktop and mobile layouts. Those subheaders were not hidden due to this bug.

## How it is fixed:

Changing the running order of the `storeScrollData` mutating first the `clientHeight`, it is ensured that the data is consistent between the composable and the store.

**Before:**

https://github.com/user-attachments/assets/6c3605df-97cf-4188-b07d-104c4f8ab1df

**After:**

https://github.com/user-attachments/assets/82cefb86-5d9a-4211-b857-0e7423b736a1

## Motivation and context

## Type of change

- [ ] `Main`
- [X] Other. Specify: https://github.com/empathyco/x/tree/vue3-update-rc

## How has this been tested?

## Checklist:

- [X] I have made corresponding changes to the **[documentation](./CONTRIBUTING.md#documentation-style-guide)**.

### DIFF
--- a/packages/x-components/src/components/scroll/use-scroll.ts
+++ b/packages/x-components/src/components/scroll/use-scroll.ts
@@ -113,9 +113,9 @@ export function useScroll(
    */
   const storeScrollData = () => {
     if (scrollEl.value) {
-      currentPosition.value = scrollEl.value.scrollTop;
       scrollHeight.value = scrollEl.value.scrollHeight;
       clientHeight.value = scrollEl.value.clientHeight;
+      currentPosition.value = scrollEl.value.scrollTop;
     }
   };
 
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@
 
 **/tests/e2e/videos/
 **/tests/e2e/screenshots/
+**/tests/e2e/downloads/
 
 # local env files
 .env.local
--- a/packages/x-components/.gitignore
+++ b/packages/x-components/.gitignore
@@ -8,6 +8,7 @@
 
 /tests/e2e/videos/
 /tests/e2e/screenshots/
+/tests/e2e/downloads/
 
 # local env files
 .env.local
--- a/packages/x-components/cypress.config.ts
+++ b/packages/x-components/cypress.config.ts
@@ -14,6 +14,7 @@ export default defineConfig({
     supportFile: 'tests/support/index.ts',
     fixturesFolder: 'tests/e2e/fixtures',
     screenshotsFolder: 'tests/e2e/screenshots',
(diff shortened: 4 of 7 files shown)


packages/x-components/src/components/scroll/use-scroll.ts
```diff
@@ -113,9 +113,9 @@ export function useScroll(
    */
   const storeScrollData = () => {
     if (scrollEl.value) {
-      currentPosition.value = scrollEl.value.scrollTop;
       scrollHeight.value = scrollEl.value.scrollHeight;
       clientHeight.value = scrollEl.value.clientHeight;
+      currentPosition.value = scrollEl.value.scrollTop;
```
joseacabaneros, line 118: This change makes sense in terms of watch callbacks. I was debugging the same example with this composable and the mixin and the side effect seems to be the order of reactivity execution regarding `scrollDirection` (as you explain very well in the PR description).

I don't like the `isClientHeightChanging` concept itself, and I think is the key part of the issue and side effects. It could be refactored to handle the reactivity and the side effects in a more controlled way, but for the time being, this solution is easy and works so perfect.

WDYT to add a unit test sheet for it and/or add a test in the scroll e2e? 😅

victorcg88: Added a specific e2e to check for the scrolling direction 👍

joseacabaneros (approved):

PR #1656, opened 2024-11-20
feat: update x-tailwindcss package to vue 3

diegopf:
This pull request focuses on refactoring Vue components in the `packages/x-tailwindcss/demo` directory to use the Composition API and `defineComponent` from Vue 3. This allow us to remove the peer dependency to vue 2 for the `x-tailwincdss` package.

BREAKING CHANGE: The showcase is no longer exported as Vue components but as web application.


AdrianArenal (approved):

packages/x-tailwindcss/demo/src/components/xds-suggestion.vue
```diff
@@ -21,68 +21,78 @@
...
+          'x-suggestion x-suggestion-success x-suggestion-md',
+          'x-suggestion x-suggestion-auxiliary x-suggestion-md',
+          'x-suggestion-tag x-suggestion-error-50 x-suggestion-lg'
+        ]
+      }
+    },
+    computed: {
+      sections(): ShowcaseSections {
+        return {
+          Default: [this.base],
```
annacv, line 78: `props` instead of `this`?

annacv (commented):
Hi, if I am not wrong, there is no need to copy `copyCssClassesToClipboard` and  `removeClassPrefix` as methods because we bind them when we use the `XdsBaseShowcase` component.
I tested removing those methods in `xds-input.vue`, `xds-highlight.vue` & `xds-icon.vue` and everything is working fine.

annacv (approved):

PR #1668, opened 2024-12-02
style(tailwind): extend tailwind config to be used by the demo

joseacabaneros:
Extends tailwind config with the x-tailwindcss theme to be able to use the preset config like `x-bg-neutral-0`


annacv:
Hi just to arise I see search-input styles broken and also that the `data-test` which is meant to use the `mainScrollDirection` is getting this value as undefined:
Captura de pantalla 2024-12-05 a les 12 55 44 p  m

joseacabaneros:
> Hi, just to arise I see search-input styles broken and also that the `data-test` which is meant to use the `mainScrollDirection` is getting this value as `undefined`: Captura de pantalla 2024-12-05 a les 12 55 44 p  m

Trying it in local, I see the `search-input` styles working well. 
Screenshot 2024-12-27 at 16 40 41

The `undefined` in the `data-test` I think is how it has worked before the XDS breaking changes. We could go deeper into in an independent PR.

diegopf (approved):

PR #1680, opened 2024-12-16
feat: add related prompts list component

lauramargar:
# Pull request template
 
We need to simplify the implementation of related prompts in new setups. Copy-pasting the same components and logic in every setup is not a good practice, so we have added a new component (`RelatedPromptsTagList`) to the base project.

## Motivation and context

## Type of change

- [X] `Main`
- [ ] Other. Specify:

## How has this been tested?

## Checklist:

### DIFF
--- a/packages/x-components/src/x-modules/related-prompts/components/related-prompt.vue
+++ b/packages/x-components/src/x-modules/related-prompts/components/related-prompt.vue
@@ -4,6 +4,9 @@
     @keydown="toggleSuggestion(index)"
     class="x-related-prompt__button"
     :class="[{ 'x-related-prompt-selected__button': isSelected }]"
+    role="button"
+    aria-pressed="true"
+    tabindex="0"
   >
     <slot name="related-prompt-button-info">
       <div class="x-related-prompt__button-info">
--- a/packages/x-components/src/x-modules/related-prompts/components/related-prompt.vue
+++ b/packages/x-components/src/x-modules/related-prompts/components/related-prompt.vue
@@ -1,6 +1,7 @@
 <template>
-  <button
+  <div
     @click="toggleSuggestion(index)"
+    @keydown="toggleSuggestion(index)"
     class="x-related-prompt__button"
     :class="[{ 'x-related-prompt-selected__button': isSelected }]"
   >
@@ -20,7 +21,7 @@
       <CrossTinyIcon v-if="isSelected" class="x-icon-lg" />
       <PlusIcon v-else class="x-icon-lg" />
     </slot>
-  </button>
+  </div>
 </template>
 <script lang="ts">
   import { defineComponent, PropType } from 'vue';
--- a/packages/x-components/src/x-modules/related-prompts/components/related-prompt.vue
+++ b/packages/x-components/src/x-modules/related-prompts/components/related-prompt.vue
@@ -2,13 +2,13 @@
   <button
     @click="toggleSuggestion(index)"
     class="x-related-prompt__button"
-    :class="[{ 'x-related-prompt-selected__button': isSelected(index) }]"
+    :class="[{ 'x-related-prompt-selected__button': isSelected }]"
   >
     <slot name="related-prompt-button-info">
       <div class="x-related-prompt__button-info">
         <span
           class="x-typewritter-initial"
-          :class="[{ 'x-typewritter-animation': isVisible }]"
+          :class="[{ 'x-typewritter-animation': arePromptsVisible }]"
           :style="{
             animationDelay: `${index * 0.4 + 0.05}s`,
             '--suggestion-text-length': relatedPrompt.suggestionText.length
@@ -17,8 +17,8 @@
           {{ relatedPrompt.suggestionText }}
         </span>
       </div>
-      <CrossTinyIcon v-if="isSelected(index)" class="x-icon-lg" />
-      <PlusIcon v-else class="x-icon-neutral-80 x-icon-lg" />
+      <CrossTinyIcon v-if="isSelected" class="x-icon-lg" />
+      <PlusIcon v-else class="x-icon-lg" />
     </slot>
   </button>
 </template>
@@ -49,7 +49,11 @@
         type: Object as PropType<RelatedPrompt>,
         required: true
       },
-      isVisible: {
+      arePromptsVisible: {
+        type: Boolean,
+        default: false
+      },
+      isSelected: {
         type: Boolean,
         default: false
       },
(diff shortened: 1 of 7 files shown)


packages/x-components/src/x-modules/related-prompts/components/related-prompt.vue
```diff
@@ -64,39 +49,31 @@
...
-        } else {
-          selectedNextQuery.value = nextQuery;
-        }
-      }
+      const toggleSuggestion = (index: number): void => {
+        x.emit('UserSelectedARelatedPrompt', index);
+      };
 
-      return { selectedNextQuery, onClick };
+      const isSelected = (index: number): boolean => selectedPrompt.value === index;
```
victorcg88, line 69:
```suggestion
      const isPromptSelected = (index: number): boolean => selectedPrompt.value === index;
```

Apart from the naming, I am wondering if it is a good idea to pass this `isSelected` as a prop because we are already calculating it in the parent 🤔

packages/x-components/src/x-modules/related-prompts/components/related-prompt.vue
```diff
@@ -1,60 +1,45 @@
...
-      </slot>
-    </div>
-  </div>
+  <button
+    @click="toggleSuggestion(index)"
+    class="x-related-prompt__button"
+    :class="[{ 'x-related-prompt-selected__button': isSelected(index) }]"
+  >
+    <slot name="related-prompt-button-info">
+      <div class="x-related-prompt__button-info">
```
annacv, line 8: Hi, can we convert this div into a `span`? 
A `div` [is not allowed inside](https://html.spec.whatwg.org/multipage/form-elements.html#the-button-element) a `button`, it can lead to accessibility issues, e.g. for screen readers

packages/x-components/src/x-modules/related-prompts/components/related-prompt.vue
```diff
@@ -1,60 +1,45 @@
...
+          :style="{
+            animationDelay: `${index * 0.4 + 0.05}s`,
+            '--suggestion-text-length': relatedPrompt.suggestionText.length
+          }"
+        >
+          {{ relatedPrompt.suggestionText }}
+        </span>
+      </div>
+      <CrossTinyIcon v-if="isSelected(index)" class="x-icon-lg" />
+      <PlusIcon v-else class="x-icon-neutral-80 x-icon-lg" />
```
annacv, line 21: Colors should be customizable for customers

lauramargar: yes, it was a mistake, that class should not be there

packages/x-components/src/x-modules/related-prompts/components/related-prompt.vue
lauramargar: Yes, it makes sense. If `isSelected` is already calculated in the parent, I think it's not necessary to calculate it again in `related-prompt.vue`

packages/x-components/src/x-modules/related-prompts/components/related-prompt.vue
lauramargar: with a `div` element instead of a `button` seems to work fine

annacv: Then I would add there a `role=button` and an `aria-pressed` attributes for assistive technologies

packages/x-components/src/x-modules/related-prompts/components/related-prompt.vue
```diff
@@ -64,39 +50,31 @@
         type: Object as PropType<RelatedPrompt>,
         required: true
       },
-      nextQueryButtonClass: {
-        type: String,
-        default: 'x-button-outlined'
+      arePromptsVisible: {
```
victorcg88, line 53: In `RelatedPromptsTagList` we talk about all the prompts being visible (plural). In this `RelatedPrompt` is about a single prompt, so I think here the prop should be called `isPromptViible`

packages/x-components/src/x-modules/related-prompts/components/related-prompt.vue
```diff
@@ -64,39 +50,31 @@
...
+      index: {
+        type: Number,
+        required: true
       }
     },
-    setup(props) {
-      const selectedNextQuery = ref(props.relatedPrompt.nextQueries[0]);
+    setup() {
+      const x = use$x();
+      const { selectedPrompt } = useState('relatedPrompts', ['selectedPrompt']);
```
victorcg88, line 68: `selectedPrompt` is not used anymore, right?

packages/x-components/src/x-modules/related-prompts/components/related-prompt.vue
lauramargar: You're right

annacv:
Hi, I miss the docs section in both components 👀

packages/x-components/src/x-modules/related-prompts/components/related-prompt.vue
```diff
@@ -64,39 +50,29 @@
         type: Object as PropType<RelatedPrompt>,
         required: true
       },
-      nextQueryButtonClass: {
-        type: String,
-        default: 'x-button-outlined'
```
annacv, line 69: I think we should keep this level of customization, so we can add a class to modify spacing or add background colors as per customer's requirements.

packages/x-components/src/x-modules/related-prompts/components/related-prompt.vue
annacv: @lauramargar to make the element focusable, you should use `tabindex="0"`, then eslint won't warn if we set that role in a div

annacv:
> > Hi, I miss the `<docs>` section in both components 👀, and the update of `Home.vue` were the `RelatedPrompt.vue` is implemented
> 
> Hi @annacv !! Related to the `<docs>` section, I think it's better to add or modify the current documentation after finishing the related-prompts multi-query implementation. It is already in progress.
> 
> And related to updating the `Home.vue` file, I don't know what is better, to remove the part of related-prompts because the endpoint is not working or update it even if the endpoint doesn't work.

Hi! If the multi-query implementation is done in this same file, ok we can leave the <doc> section to be finished then, otherwise if it is done in another file, I would add doc in this PR. 

Regarding the `Home.vue` file, I mean replacing [this part](https://github.com/empathyco/x/blob/main/packages/x-components/src/views/home/Home.vue#L90) by this component implementation (IMO, no matter if the endpoint is ready or not, same use case as ExperienceControls)

annacv (approved):

PR #1687, opened 2025-01-13
feat: create pagination component

annacv:
# Pull request template
 
The idea is to be able to display results using the current infinite-scroll custom directive or using a pagination component.

## Motivation and context

Navigate forward and backward given a response with x total number of results.

## Type of change

- [x] Change requires a documentation update

- [x] `Main`
- [ ] Other. Specify:

## How has this been tested?

To test in local: 
* Add the component in Home.vue & remove the v-infinite-scroll directive from ResultList
* currentPage can be inferred from the URL state module

## Checklist:


packages/x-components/src/x-modules/search/store/module.ts
```diff
@@ -29,7 +29,8 @@ export const searchXStoreModule: SearchXStoreModule = {
     selectedFilters: {},
     params: {},
     config: {
-      pageSize: 24
+      pageSize: 24,
+      pageMode: 'infinite_scroll'
```
annacv, line 33: Change to 'pagination' to test with the new component

packages/x-components/src/x-modules/search/store/actions/fetch-and-save-search-response.action.ts
```diff
@@ -33,17 +33,25 @@ const { fetchAndSave, cancelPrevious } = createFetchAndSaveActions<
...
     results
   } = state;
-  const start = page === 1 ? 0 : results.length;
+
+  const start =
+    pageMode === 'infinite_scroll'
+      ? page === 1
+        ? 0
+        : results.length
+      : state.config.pageSize * page;
```
diegopf, line 46: Maybe it's me but IMO it's better to have an if else here rather than nested ternaries

packages/x-components/src/components/page-selector.vue
```diff
@@ -0,0 +1,216 @@
...
+        if (page === '...') {
+          return;
+        }
+        if (typeof page === 'number' && page > 0 && page <= props.totalPages) {
+          bus.emit('UserSelectedAPage', page);
+          /**
+           * Emits scroll to top to prevent keeping the position if there is more content
+           * after results, as for example Next Queries Preview.
+           */
+          bus.emit('UserClickedScrollToTop', 'main-scroll');
```
diegopf, line 114: You can't hardcode the scroll id here because if you do then this component will only work if it is placed on the main scroll. Otherwise strange behaviours will happen. We might need to think a different way to handle the scroll scenario or make the scroll id configurable.

packages/x-components/src/components/page-selector.vue
lauramargar, line 1: Hi! I think you should export the component in the `index.ts` file.

lauramargar:
I think the implementation is not working ok. The number of results on each page is not always the same.
Could you review it? 


https://github.com/user-attachments/assets/bb30e28b-5682-45fd-8ed0-d1ab37016672

packages/x-components/src/x-modules/search/store/actions/fetch-and-save-search-response.action.ts
```diff
@@ -33,17 +33,25 @@ const { fetchAndSave, cancelPrevious } = createFetchAndSaveActions<
...
+
+  const rows = pageMode === 'infinite_scroll' ? pageSize * page - start : pageSize;
 
   return {
     ...restRequest,
     ...(origin && { origin }),
-    start,
-    rows: pageSize * page - start
+    start: start,
+    rows: rows
```
diegopf, line 54:
```suggestion
    start,
    rows
```

packages/x-components/src/components/page-selector.vue
```diff
@@ -0,0 +1,226 @@
+<template>
+  <nav class="x-flex x-items-center x-justify-center x-gap-2" aria-label="Pagination">
```
diegopf, line 2: We can't use tailwindcss classes here, specially with the x prefix. If we do we will force the consumer to have the same prefix.

packages/x-components/src/components/page-selector.vue
```diff
@@ -0,0 +1,226 @@
...
+    <button
+      @click="selectPage(currentPage - 1)"
+      class="x-button x-button-lead"
```
diegopf, line 5: I'd allow to customize these classes

packages/x-components/src/components/page-selector.vue
```diff
@@ -0,0 +1,226 @@
...
+        'x-cursor-not-allowed': page === '...'
+      }"
+      :data-test="`page-button-${page}`"
+    >
+      <slot name="page-button">{{ page }}</slot>
+    </button>
+
+    <button
+      @click="selectPage(currentPage + 1)"
+      class="x-button x-button-lead"
```
diegopf, line 29: Same as the previous button, I'd allow to configure the classes

packages/x-components/src/components/page-selector.vue
```diff
@@ -0,0 +1,226 @@
...
+    <button
+      v-for="page in visiblePages"
+      :key="page"
+      @click="selectPage(page)"
+      class="x-button"
+      :class="{
+        'x-button-lead x-cursor-default': page === currentPage,
+        'x-button-ghost': page !== currentPage,
+        'x-cursor-not-allowed': page === '...'
+      }"
```
diegopf, line 21: We need to think about how can we style this from the consumer side. Maybe it's a good idea to have selected and disabled states so we can target them from the outside

diegopf (changes requested):

diegopf (approved):

PR #1837, opened 2025-08-14
feat(ai): create initial `AI` x-module

joseacabaneros:
Initialize `AI` x-module and skeleton for `x-adapter-platform` & `x-types`


packages/x-components/src/x-modules/ai/components/index.ts
joseacabaneros, line 1: The decision under `AiOverview` vs `AIOverview` is following the current naming like `UrlHandler'. It is the same case in types like `AiConfig ` or `AIConfig`. I stuck to the current name

lauramargar (approved):

PR #1799, opened 2025-06-04
build: use entries instead of custom plugin

davidglezz:
# Pull request template

Simplify build process removing a custom plugin that generates entry points js files after the build, parsing output files and modifying it. Replaced by defining several input entries, the oficial way. supported by rollup, rolldown, vite & tsdown.

This facilitates migration outside of the rollup.

Tested with `x-chedraui`, requires `typescript` 5

This PR allows easy migration to vite/tsdown/rolldown

## Motivation and context

- [ ] Dependencies. If any, specify:
- [ ] Open issue. If applicable, link:

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that causes existing functionality to not work as expected)
- [ ] Change requires a documentation update

## What is the destination branch of this PR?

- [x] `Main`
- [ ] Other. Specify:

## How has this been tested?

Tests performed according to [testing guidelines](./contributing/tests.md): 

## Checklist:

- [ ] My code follows the **[style guidelines](./CONTRIBUTING.md#style-guides)** of this project.
- [ ] I have performed a **self-review** on my own code.
- [ ] I have **commented** my code, particularly in hard-to-understand areas.
- [ ] I have made corresponding changes to the **[documentation](./CONTRIBUTING.md#documentation-style-guide)**.
- [ ] My changes generate **no new warnings**.
- [ ] I have added **tests** that prove my fix is effective or that my feature works.
- [ ] New and existing **unit tests pass locally** with my changes.

PR #2025, opened 2026-03-02
Use vite

davidglezz:
After https://github.com/empathyco/x/pull/1799 and https://github.com/empathyco/x/pull/2017

the next step would be using vite only.

also, added tsdown as experimental, but for now, works better with vite.

This PR adds the last bits, copies some required files to be a package.

Untested because it require changes in the archetype/user


victorcg88:
Code from this PR has become obsolete.

